PR #25, opened 2018-06-13
feat: Timeouts

richardwerkman:
It seems I forgot to append the timeout feature in the initial code commit. So in this pull request this feature is added. 

The feature checks the duration of the initial testrun and calculates an appropriate timeout value for testing the mutants. When testing a mutant takes longer than the timeout value, the mutant gets the status: timeout. This status will count as detected for the mutationscore.


src/Stryker.CLI/Stryker.CLI.UnitTest/StrykerCLITests.cs
```diff
@@ -109,5 +109,21 @@ public void StrykerCLI_ShouldPassLogFileArgumentsToStryker(string argName)
 
             mock.Verify(x => x.RunMutationTest(It.Is<StrykerOptions>(o => o.LogOptions.LogToFile == true)));
         }
+
+        [Theory]
+        [InlineData("--timeoutMS")]
+        [InlineData("-t")]
+        public void StrykerCLI_ShouldPassTimeoutArgumentsToStryker(string argName)
```
nicojs: I'm missing the `act` part of this test name: `StrykerCLI_WithTimeoutArgument_ShouldPassTimeoutToStryker`

src/Stryker.CLI/Stryker.CLI/Stryker.CLI.csproj
```diff
@@ -4,12 +4,19 @@
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp2.0</TargetFramework>
     <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
-    <PackageId>dotnet-stryker</PackageId>
+    <PackageId>Stryker.NET.Runner.DotNetCore</PackageId>
```
nicojs: This is also part of the other PR?

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/AssemblyReferenceResolverTests.cs
```diff
@@ -28,7 +28,8 @@ public void ShouldAnalyzeProjectDependencies()
                 It.IsAny<string>(),
                 It.IsAny<string>(),
                 It.IsAny<string>(),
-                null))
```
nicojs: I think it's time to refactor this test class.

1. Create the subject under test in the constructor of the test class.
2. Create a helper method to arrange the result of the `processExecutorMock`. For example: `ArrangeProcessExecutor(string result)`

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialBuildProcessTests.cs
```diff
@@ -17,7 +17,8 @@ public void InitialBuildProcess_ShouldThrowExceptionOnFail()
                 It.IsAny<string>(),
                 It.IsAny<string>(),
                 It.IsAny<string>(),
-                It.IsAny<IEnumerable<KeyValuePair<string, string>>>()))
+                It.IsAny<IEnumerable<KeyValuePair<string, string>>>(),
+                It.IsAny<int>()))
```
nicojs: Same here IMHO

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialTestProcessTests.cs
```diff
@@ -0,0 +1,38 @@
...
+
+            var target = new InitialTestProcess();
+
+            var exception = Assert.Throws<InitialTestRunFailedException>(() => target.InitialTest(testRunnerMock.Object));
+        }
+
+        [Fact]
+        public void InitialTestProcess_ShouldCalculateTestTimeout()
+        {
+            var testRunnerMock = new Mock<ITestRunner>(MockBehavior.Strict);
```
nicojs: Why not create this mock and the subject under test in the constructor?

src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/DotnetTestRunnerTests.cs
```diff
@@ -17,15 +17,16 @@ public void RunAllExitCode0SuccessShouldBeTrue()
                 It.IsAny<string>(), 
                 It.IsAny<string>(), 
                 It.IsAny<string>(), 
-                It.IsAny<IEnumerable<KeyValuePair<string, string>>>()))
```
nicojs: Same here.

src/Stryker.Core/Stryker.Core/Options/ValidatedStrykerOptions.cs
```diff
@@ -0,0 +1,41 @@
+﻿using Stryker.Core.Logging;
+using Stryker.Core.Mutators;
+using System.Collections.Generic;
+
+namespace Stryker.Core.Options
+{
+    public class ValidatedStrykerOptions
```
nicojs: What's the reasoning behind this class? I don't like classes like this as they shadow the original class (`StrykerOptions`).

src/Stryker.Core/Stryker.Core/TestRunners/DotnetTestRunner.cs
```diff
@@ -18,15 +18,16 @@ public DotnetTestRunner(string path, IProcessExecutor processProxy)
 
         public DotnetTestRunner(string path) : this(path, new ProcessExecutor()) { }
 
-        public TestRunResult RunAll()
+        public TestRunResult RunAll(int timeoutMS = 0)
```
nicojs: I think a `int? timeoutMS = null` is clearer than using `0` as a magical number that means `off`

src/Stryker.Core/Stryker.Core/TestRunners/ITestRunner.cs
```diff
@@ -2,7 +2,7 @@
 {
     public interface ITestRunner
     {
-        TestRunResult RunAll();
+        TestRunResult RunAll(int timeoutMS = 0);
```
nicojs: Why does this interface need a default? I think it should be required here and optionally given a default higher up in the chain.

src/Stryker.Core/Stryker.Core/Testing/ProcessExtensions.cs
```diff
@@ -0,0 +1,136 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+
+namespace Stryker.Core.Testing
+{
+    internal static class ProcessExtensions
```
nicojs: Could we add a link to the original code here?

src/Stryker.Core/Stryker.Core/Options/ValidatedStrykerOptions.cs
richardwerkman: This is not a complete shadow class of `StrykerOptions`. StrykerOptions only has string values. In the `ValidatedStrykerOptions` this has been cast to int. Future properties could also have different types.

nicojs: I didn't say _copy_. If the name is the same, but type is different: it's still a shadow IMHO ;).

Couldn't we force the dotnet cli to interpret a value as an `int`? If not, i still think it should be done in the dotnet cli package, rather than the core Stryker package. Don't you agree?

richardwerkman: I didn't want to make the CLI responsible for casting values. So now the validator casts and validates all option values. If this is a huge issue I can change the casting to take place in the CLI. The need for the "Shadow" class would be gone that way.

richardwerkman:
@nicojs The tests have been refactored and your other review points have been updated as well. The only open question right now is the `ValidatedStrykerOptions` class. I would like to merge this pr as soon as possible because the current version on NuGet still misses this important feature.

nicojs:
Awesome! I'll take a look later today

src/Stryker.CLI/Stryker.CLI/Stryker.CLI.csproj
```diff
@@ -10,6 +10,13 @@
     <Product>Mutation Testing</Product>
     <AssemblyName>dotnet-stryker</AssemblyName>
     <RootNamespace>Stryker.CLI</RootNamespace>
+    <Version>0.0.1</Version>
```
nicojs: I thought this was for #45 ?

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/TimeoutValueCalculatorTests.cs
```diff
@@ -0,0 +1,39 @@
+﻿using Shouldly;
+using Stryker.Core.Initialisation;
+using Xunit;
+
+namespace Stryker.Core.UnitTest.Initialisation
+{
+    public class TimeoutValueCalculatorTests
+    {
+        [Fact]
+        public void Calculator_ShouldCalculateTimeoutValueNoExtra()
```
nicojs: I'm missing the `act` part in these method names.

src/Stryker.Core/Stryker.Core.UnitTest/MockExtensions.cs
```diff
@@ -0,0 +1,25 @@
+﻿using Moq;
+using Stryker.Core.Testing;
+using System.Collections.Generic;
+using System.IO;
+
+namespace Stryker.Core.UnitTest
+{
+    public static class MockExtensions
+    {
+        public static void SetupProcessMockToReturn(this Mock<IProcessExecutor> processExecutorMock, string result, int exitCode = 0)
```
nicojs: Good idea to make it sharable across all tests.

src/Stryker.Core/Stryker.Core/Options/ValidatedStrykerOptions.cs
nicojs: I think the CLI should be responsible for casting to the correct values. 

The reason why they are strings is because they are interpreted from the command line as strings. The Stryker core shouldn't be bothered with this detail, right? If you would want to reuse the code from a place where you would already have it as an int, you would need to cast it back to string. Doesn't make sense.

Validating that the int is in the expected range however _is_ a responsibility of the Stryker core.

src/Stryker.Core/Stryker.Core/TestRunners/DotnetTestRunner.cs
```diff
@@ -18,15 +18,16 @@ public DotnetTestRunner(string path, IProcessExecutor processProxy)
...
         {
             var result = _processExecutor.Start(
                 _path,
                 "dotnet",
                 "test --no-build --no-restore",
                 new Collection<KeyValuePair<string, string>>() {
                     new KeyValuePair<string, string>("ActiveMutation", _activeMutation.ToString())
-                });
+                },
+                timeoutMS ?? 0);
```
nicojs: I don't think that `0` is a logical way of communicating "no timeout".

src/Stryker.Core/Stryker.Core/Testing/ProcessExecutor.cs
```diff
@@ -29,24 +34,44 @@ public ProcessExecutor(bool redirectOutput = true)
...
-            process.WaitForExit();
+                string output = "";
+                var processDoneHandle = new ManualResetEvent(false);
 
-            return new ProcessResult()
-            {
-                ExitCode = process.ExitCode,
-                Output = output
-            };
+                Task.Run(() =>
```
nicojs: This is a pretty long method as it is. Could we move this code to a `private Task<string> RunProcessAsync(Process p)` method?

src/Stryker.Core/Stryker.Core/Testing/ProcessExtensions.cs
nicojs: Ah it is in the header. Thanks!

src/Stryker.Core/Stryker.Core/TestRunners/DotnetTestRunner.cs
```diff
@@ -18,15 +18,16 @@ public DotnetTestRunner(string path, IProcessExecutor processProxy)
 
         public DotnetTestRunner(string path) : this(path, new ProcessExecutor()) { }
 
-        public TestRunResult RunAll()
+        public TestRunResult RunAll(int? timeoutMS = null)
```
nicojs: Why does timeoutMS have a default value? It is never called without arguments.

src/Stryker.Core/Stryker.Core/Stryker.Core.csproj
```diff
@@ -8,12 +8,13 @@
     <OutputType>Library</OutputType>
     <StartupObject />
     <PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
-    <Description>The main package for stryker.net</Description>
+    <Description>The core package for Stryker.NET. Used by other Stryker packages to run. Please install a Stryker.NET.Runner.* package to use this package.</Description>
     <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
-    <PackageId>Stryker.Core</PackageId>
+    <PackageId>Stryker.NET.Core</PackageId>
```
nicojs: I thought these changes were part of #45 ?

PR #46, opened 2018-06-29
Checked mutator

Timoosterbroek:


src/Stryker.Core/Stryker.Core/Mutators/CheckedMutator.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        public override IEnumerable<Mutation> ApplyMutations(CheckedExpressionSyntax node)
+        {
+            if (node.Kind() == SyntaxKind.CheckedExpression)
+            {
+                yield return new Mutation()
+                {
+                    OriginalNode = node,
+                    ReplacementNode = node.Expression,
+                    DisplayName = "Checked expression to expression mutation",
```
richardwerkman: I think this would look better like: "Remove checked expression". What do you think?

Timoosterbroek: Good idea!

PR #102, opened 2018-09-01
Check if there is a config file, else use default

### LINKED ISSUE #101: Remove the need for a config file
Atm you must have a config file if you dont give any arguments to the CLI. I would like to be able to not give a config file and still dont have a config file. 

It would be cool if we check whether there is a config file and if it does not exist we take the default values.

yaelkeemink:
Check if there is a config file called stryker-config.json or with the given name. else it should use the default 

Fixes #101

### DIFF
--- a/src/Stryker.CLI/Stryker.CLI.UnitTest/Stryker.CLI.UnitTest.csproj
+++ b/src/Stryker.CLI/Stryker.CLI.UnitTest/Stryker.CLI.UnitTest.csproj
@@ -22,8 +22,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <None Include="stryker-config.json" CopyToOutputDirectory="Always" />
-    <None Include="filled-stryker-config.json" CopyToOutputDirectory="Always" />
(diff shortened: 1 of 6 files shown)


src/Stryker.CLI/Stryker.CLI.UnitTest/Stryker.CLI.UnitTest.csproj
```diff
@@ -22,8 +22,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <None Include="stryker-config.json" CopyToOutputDirectory="Always" />
-    <None Include="filled-stryker-config.json" CopyToOutputDirectory="Always" />
+    <None Update="filled-stryker-config.json" CopyToOutputDirectory="Always" />
```
Mobrockers: The [docs ](https://docs.microsoft.com/en-us/visualstudio/msbuild/item-element-msbuild?view=vs-2017) say this about the Update attribute:

> Enables you to modify metadata of a file that was included by using a glob.

This does not seem to do what we want to happen to this file. What is the purpose of this change?

PR #107, opened 2018-09-01
Test runs are now multithreaded

### LINKED ISSUE #20: Multi threaded testruns
Right now only one testrun occurs at a time. This is not optimal since .Net is excellent in managing threads and parallel processes. 

A better situation would be if more testrunners could run in parallel. The number of testrunners should be configurable. When not configured, the number of cores of the machine could be used as default.

This feature would speed up the mutationtest process by a large margin.

ArpadK:
Closes #20

### DIFF
--- a/src/Stryker.CLI/Stryker.CLI.UnitTest/StrykerCLITests.cs
+++ b/src/Stryker.CLI/Stryker.CLI.UnitTest/StrykerCLITests.cs
@@ -27,7 +27,7 @@ namespace Stryker.CLI.UnitTest
         public void StrykerCLI_WithNoArguments_ShouldStartStrykerWithDefaultOptions()
         {
             var mock = new Mock<IStrykerRunner>(MockBehavior.Strict);
-            mock.Setup(x => x.RunMutationTest(It.Is<StrykerOptions>(c => c.AdditionalTimeoutMS == 2000 &&
+            mock.Setup(x => x.RunMutationTest(It.Is<StrykerOptions>(c => c.AdditionalTimeoutMS == 30000 &&
                                                                         c.LogOptions.LogLevel ==LogEventLevel.Information &&
                                                                         c.LogOptions.LogToFile == false &&
                                                                         c.ProjectUnderTestNameFilter == null &&
@@ -65,7 +65,7 @@ namespace Stryker.CLI.UnitTest
         public void StrykerCLI_WithConfigFileIsFalse_ShouldStartStrykerWithDefaultOptions(string argName)
         {
             var mock = new Mock<IStrykerRunner>(MockBehavior.Strict);
-            mock.Setup(x => x.RunMutationTest(It.Is<StrykerOptions>(c => c.AdditionalTimeoutMS == 2000 &&
+            mock.Setup(x => x.RunMutationTest(It.Is<StrykerOptions>(c => c.AdditionalTimeoutMS == 30000 &&
                                                                         c.LogOptions.LogLevel == LogEventLevel.Information &&
                                                                         c.LogOptions.LogToFile == false &&
                                                                         c.ProjectUnderTestNameFilter == null &&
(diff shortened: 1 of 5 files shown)


src/Stryker.CLI/Stryker.CLI.UnitTest/StrykerCLITests.cs
```diff
@@ -27,7 +27,7 @@ public void StrykerCLI_WithHelpArgument_ShouldNotStartStryker(string argName)
         public void StrykerCLI_WithNoArguments_ShouldStartStrykerWithDefaultOptions()
         {
             var mock = new Mock<IStrykerRunner>(MockBehavior.Strict);
-            mock.Setup(x => x.RunMutationTest(It.Is<StrykerOptions>(c => c.AdditionalTimeoutMS == 2000 &&
+            mock.Setup(x => x.RunMutationTest(It.Is<StrykerOptions>(c => c.AdditionalTimeoutMS == 30000 &&
```
yaelkeemink: Can't we get the default max timeout from CLIOptions?

src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
```diff
@@ -49,7 +49,7 @@ public static class CLIOptions
             ArgumentName = "--timeoutMS",
             ArgumentShortName = "-t <ms>",
             ArgumentDescription = "When passed, a logfile will be created for this mutationtest run on trace level",
-            DefaultValue = 2000,
+            DefaultValue = 30000,
```
Mobrockers: Default value has changed, please update the readme accordingly:
https://github.com/stryker-mutator/stryker-net/blame/master/src/Stryker.CLI/README.md#L36

Mobrockers:
It would be nice if the number of used threads was user configurable. Defaulting to 4 should be fine.

nicojs:
How about defaulting to be of cores / 2?

Mobrockers:
@nicojs Currently there does not seem to be a cross platform way to determine core count, so we cannot use this. We assume most developers would have at least 4 cores available these days.

nicojs:
How about https://docs.microsoft.com/en-us/dotnet/api/system.environment.processorcount?redirectedfrom=MSDN&view=netcore-2.1#System_Environment_ProcessorCount ?

src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
ArpadK: Ah good point

richardwerkman:
@nicojs the Environment.ProcessorCount returns the number of logical processors. With hyper threading enabled this is 2x the core count. Could this be a problem? Since dotnet test also seems to spawn multiple background processes.

nicojs:
> With hyper threading enabled this is 2x the core count. Could this be a problem? 

The idea behind hyperthreading is that a processor can handle 2 threads. So isn't the _number of logical cores_ exactly what we want? We shouldn't care about the technical details of the processor, right?

> Since dotnet test also seems to spawn multiple background processes.

That's why I suggest **number of cores / 2**. We could also some tests to see if the CPU doesn't get too high. For example > 80% could be considered too high.

Just to be clear, even 2 test runners could be a bad idea depending on the use case. A unit test that tests a thread pool might do stuff like creating threads for each core and hit 100% under normal circumstances. So it should still be configurable and well documented.

Mobrockers:
@ArpadK Do we want to make the number of used threads user configurable in this PR or will you make a new issue for that change?

richardwerkman:
For me it would be ok to create a new issue for the configurable number of parallel testruns.

ArpadK:
Yeah, I was thinking about that as well. Seen as this is a massive performance improvement I think it would be best to create a new issue for the configurable threads.

PR #112, opened 2018-09-09
fix: add new targets file with nugetID as name

### LINKED ISSUE #47: DotNetCore 2.1 support
On .NET Core 2.1 projects the `build/dotnet-stryker.targets` seems to be ignored. This file is being copied by msbuild in the other .NET Core versions. 

An awesome feature is added in 2.1 called [DotnetTools](https://blogs.msdn.microsoft.com/dotnet/2018/05/30/announcing-net-core-2-1/#user-content-net-core-tools). This can be used to create global tools like with npm `npm i -g`

To support DotnetTool a new package has to be created with the type `DotnetTool`. This could also be added within the current CLI package.

For now the CLI seems to be broken on 2.1. To fix this users can copy the task by hand:

``` xml
  <Target Name="PrintReferences" DependsOnTargets="ResolveProjectReferences;ResolveAssemblyReferences">
    <Message Text="@(_ResolveAssemblyReferenceResolvedFiles)" Importance="high" />
  </Target>
```

Into their test .csproj file

richardwerkman:
closes #47.

### DIFF
--- a/src/Stryker.CLI/Stryker.CLI/build/StrykerMutator.DotNetCoreCli.targets
+++ b/src/Stryker.CLI/Stryker.CLI/build/StrykerMutator.DotNetCoreCli.targets
@@ -1,0 +1,5 @@
+﻿<Project>
+  <Target Name="PrintReferences" DependsOnTargets="ResolveProjectReferences;ResolveAssemblyReferences">
+    <Message Text="@(_ResolveAssemblyReferenceResolvedFiles)" Importance="high" />
+  </Target>
+</Project>


Mobrockers:
Do we still need https://github.com/stryker-mutator/stryker-net/blob/96652efedefcf02ad20d17c8ef7177330eb9c203/src/Stryker.CLI/Stryker.CLI/build/dotnet-stryker.targets

Or does this targets file replace the one with the wrong name? If so, can't we simply delete it?

richardwerkman:
@Mobrockers I'm not sure. I'm worried I will break the package for .NET Core 2.0 if i remove the old one. I couldn't test this locally yet. What do you think? Will 2.0 also import the new targets file when restoring? If so, the old one can be removed.

Mobrockers:
I don't see anywhere in the linked document that says it's specific to dotnet 2.1, so I'm wondering if this isn't how the feature was designed to begin with. Perhaps the old style is incorrect and should not have worked? I'm wondering if perhaps we might get both target files included on older projects if we have two build target files. That does not seem desirable either.

PR #124, opened 2018-10-01
feat(logging): add logLevel validation

### LINKED ISSUE #97: Add validation for logLevel
At the moment you can give the CLI invalid loglevels which result in warning-loglevel. I think we might want to throw an error here.

Valdas3:
This commit resolves #97

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
+++ b/src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
@@ -1,6 +1,4 @@
-﻿using System;
-using System.IO;
-using Microsoft.Extensions.Configuration;
+﻿using Serilog.Events;
 using Stryker.Core.Exceptions;
 using Stryker.Core.Logging;
 
@@ -25,7 +23,7 @@ namespace Stryker.Core.Options
             Reporter = ValidateReporter(reporter);
             ProjectUnderTestNameFilter = projectUnderTestNameFilter;
             AdditionalTimeoutMS = additionalTimeoutMS;
-            LogOptions = new LogOptions(logLevel, logToFile);
+            LogOptions = new LogOptions(ValidateLogLevel(logLevel), logToFile);
         }
 
         private string ValidateReporter(string reporter)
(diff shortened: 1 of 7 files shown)


src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
```diff
@@ -36,5 +34,25 @@ private string ValidateReporter(string reporter)
             }
             return reporter;
         }
+
+        private LogEventLevel ValidateLogLevel(string levelText)
+        {
+            switch (levelText?.ToLower() ?? "")
+            {
+                case "error":
+                case "":
```
simondel: I would expect the default to still be `Warning` like it was before.

simondel: My bad! Our documentation says error is the default, while warning was the actual default.

I think warning should be the default though. I'll ask around what others think.

src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
```diff
@@ -36,5 +34,25 @@ private string ValidateReporter(string reporter)
...
+                    return LogEventLevel.Error;
+                case "warning":
+                    return LogEventLevel.Warning;
+                case "info":
+                    return LogEventLevel.Information;
+                case "debug":
+                    return LogEventLevel.Debug;
+                case "trace":
+                    return LogEventLevel.Verbose;
```
simondel: This enum level is only used internally right? If that's the case, we should rename `Verbose` to `Trace` because this is a bit confusing.

Valdas3: Actually, this enum comes from Serilog, so it can't be renamed.

src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
simondel: @Mobrockers and @yaelkeemink agree with me that the default should be `Warning` so we have a majority 🎉 

@Valdas3 Could you make `Warning` the default again and update the README file(s) so they also represent this?

yaelkeemink: agreed :)

Valdas3: Sure

simondel:
Regarding the tests you've added (based on [your comment](https://github.com/stryker-mutator/stryker-net/issues/97#issuecomment-426012782): It would be cleaner if we test the validation of the loglevels in the Core project and test the integration in the CLI project. 

If you want, you could move the tests that validate each option to the Core project, but I'd also be fine with leaving it like this as we can always move the tests later on.

Issue #8, opened 2018-06-07
Merge interfaces and implementation


simondel:
Right now it seems like every interface in in a separate file. I personally find it easier to find an interface and the implementation if the interface is in the same file as the implementation (as long as there is one implementation).

simondel:
Do you have interfaces for testing purposes or because the interfaces are (or will be) used in other projects/applications? If they are just used for testing purposes I would place them in the file where the implementation is.

richardwerkman:
Mostly for testing perposes. Some will be used in other projects. But this architectural decision has not been made yet. For now it is best to place the interfaces in the implementations file.

SeeSharpy:
Sorted (i think) in #126

PR #126, opened 2018-10-01
type: refactor subject: merge interfaces and implementation 

SeeSharpy:
Moved majority of interfaces to the implementation however two provided some difficulty IMutator due to it having multiple implementations and IReadOnlyInputComponent due to the delegate void Display would need a new home


simondel:
Thanks! We'll have a look at this tomorrow.

Fixes #8

src/Stryker.Core/Stryker.Core/TestRunners/DotnetTestRunner.cs
```diff
@@ -4,6 +4,12 @@
 
 namespace Stryker.Core.TestRunners
 {
+    public interface ITestRunner
```
simondel: Could you move this back to a separate file? We will be implementing another test runner in the near future so then it will make sense to have it in a separate file (just like with the IMutator)

richardwerkman:
Thank you @SeeSharpy

PR #138, opened 2018-10-03
feat(logging): Show number of tests found in initial run

### LINKED ISSUE #105: Show number of tests found in initial run

Szpi:
Add showing in console total number of found tests in initial run. Fixes #105


simondel:
I would expect the info to be on the [TestRunResult](https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/TestRunners/TestRunResult.cs). Stryker(js) has test runners return a result that contains an [array of test results](https://github.com/stryker-mutator/stryker/blob/master/packages/stryker-api/src/test_runner/RunResult.ts#L12) (a test result is a single test execution, not the execution of a test suite. That is a RunResult).

The logging itself could be handled by the main process. As long as the runners return the expected data, all should work as expected. ([Related line in Stryker(js)](https://github.com/stryker-mutator/stryker/blob/master/packages/stryker/src/process/InitialTestExecutor.ts#L179)).

If the runner returns the number of executed tests in the TestRunResult object, we could also remove the TotalNumberOfTestsLogger since the generic code will just be a logging statement.

richardwerkman:
@simondel If we choose to make the testrunner return this info, the dotnettestrunner would have to analyse the output for every single run to get the number of tests. This information will only be used during the first testrun so this would be a waste. 

How about we store the outputmessage in the testrunner and create a method that extracts the number of tests (of the last run) only when asked for. Later testrunners can implement this in other ways, for example by counting the array of test results.

Szpi:
We could add another layer of test runners reporters for example as decorators. It could be also useful in  #9. What do you think about that idea? Parsing using span should not affect performance. I can change implementation either way.

simondel:
We could keep it simple right now by just generating the amounts of tests ran every single time. We can always optimise it later (there are bigger things to optimise on). Another option is to have different methods on a test runner, one for the initial run and one for the mutation run.

Szpi:
@richardwerkman I thought about decorating DotnetTestRunner with for example InitialTestRunnerReporter. So when we would invoke RunAll it (InitialTestRunnerReporter) would internally call DotnetTestRunner's RunAll and then report something to console and then return result of DotnetTestRunner's method. This way we would have full control about when and what to report on console. For now I will add it to result. I can refactor it in #9.

Szpi:
Have I broken something "stryker-net — Unable to find './azure-pipelines.yml'" or is it global issue?

Mobrockers:
We have transitioned to using pipelines as code using azure pipelines yaml syntax. If you merge master into your branch you should be getting the pipeline code and the build should be working again.

src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs
```diff
@@ -33,12 +35,18 @@ public class InitialisationProcess : IInitialisationProcess
...
             _inputFileResolver = inputFileResolver ?? new InputFileResolver();
             _initialBuildProcess = initialBuildProcess ?? new InitialBuildProcess();
-            _initialTestProcess = initialTestProcess ?? new InitialTestProcess();
+            _initialTestProcess = initialTestProcess ?? CreateInitionTestProcess();
             _testRunner = testRunner;
             _logger = ApplicationLogging.LoggerFactory.CreateLogger<InitialisationProcess>();
             _assemblyReferenceResolver = assemblyReferenceResolver ?? new AssemblyReferenceResolver();
         }
 
+        private static InitialTestProcess CreateInitionTestProcess()
```
Mobrockers: Should this not be CreateInitialTestProcess?

src/Stryker.Core/Stryker.Core/Parsers/ITotalNumberOfTestsParser.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Stryker.Core.Parsers
```
Mobrockers: This interface seems to only be for dependency injection. Consider placing it in the implementation cs file unless it's actually intended to have multiple implementations.

src/Stryker.Core/Stryker.Core/TestRunners/DotnetTestRunner.cs
```diff
@@ -1,20 +1,23 @@
...
-        public DotnetTestRunner(string path, IProcessExecutor processProxy)
+        public DotnetTestRunner(string path, IProcessExecutor processProxy, ITotalNumberOfTestsParser totalNumberOfTestsParser)
         {
+            _totalNumberOfTestsParser = totalNumberOfTestsParser;
             _path = path;
             _processExecutor = processProxy;
         }
 
-        public DotnetTestRunner(string path) : this(path, new ProcessExecutor()) { }
+        public DotnetTestRunner(string path) : this(path, new ProcessExecutor(), new TotalNumberOfTestsParser()) { }
```
Mobrockers: What is this empty constructor used for?

richardwerkman:
I also have more general feedback to the issue. Shouldn't we do something with this new information? Informing the user about their number of tests is nice, but we could also stop the program when the number of tests is 0. What use has mutating their code when there are no tests to run them on?

Mobrockers:
@richardwerkman This PR aims to solve #105 which doesn't mention using this new information. I think it's a good idea though so maybe we should create a new issue for any use cases we can come up with using the initial testrun result.

src/Stryker.Core/Stryker.Core/TestRunners/DotnetTestRunner.cs
Szpi: It was removed.

src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs
Szpi: Renamed

src/Stryker.Core/Stryker.Core/Parsers/ITotalNumberOfTestsParser.cs
Szpi: Done

src/Stryker.Core/Stryker.Core/Initialisation/InitialTestProcess.cs
```diff
@@ -35,6 +38,7 @@ public int InitialTest(ITestRunner testRunner)
             stopwatch.Start();
 
             var testResult = testRunner.RunAll(0, null);
+            _chalk.Default($"Total number of tests found in initial test run: {testResult.TotalNumberOfTests} {Environment.NewLine}");
```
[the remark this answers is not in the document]
Szpi: Have I do something to see output in console ( with logInformation)?. I can only see error logs, therefore I thought we use chalk for console output.

src/Stryker.Core/Stryker.Core/Initialisation/InitialTestProcess.cs
```diff
@@ -4,6 +4,7 @@
 using Stryker.Core.TestRunners;
 using System;
 using System.Diagnostics;
+using Stryker.Core.Testing;
```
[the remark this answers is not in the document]
Szpi: It is used, cannot be removed.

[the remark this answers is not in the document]
Mobrockers: using Stryker.Core.Testing was added to be able to use Chalk. Since Chalk is no longer used, this using is not neccesary. But I won't block this PR just on some unused usings, it doesn't matter that much.

Issue #121, opened 2018-09-28
Return an exit code from Stryker CLI 


shadow-cs:
It would be helpful if we could set a threshold for the Mutation score that if not reached would result in the process returning non-zero exit code.

Also failures (like the one described in #120) causes the process to exit with `0`.

For CI purposes it is fundamental to return a proper exit code.

shadow-cs:
Partial duplicate of #11

simondel:
This can be done using the thresholds that can be set by the user.

Stukongeluk:
I'll try to take a shot at this one too!

PR #140, opened 2018-10-03
feat(threshold): Add threshold arguments for the CLI

### LINKED ISSUE #11: Thresholds
Right now the mutation score color thresholds are fixed at 60 and 80.

It would be nice to let users pass their own preferred thresholds. Also an extra threshold could be added for returning a non zero exit code.

Param suggestion: `--thresholds 60 80 90` for:

- `< 60` return exit code 1
- `> 60` mutation score is red
- `> 80` mutation score is yellow
- `> 90` mutation score is green

Stukongeluk:
fixes #11 

I saw that there was another issue related to this one (#121), but that won't be fixed in this PR. However, there is a little setup for it to make it easier to fix.

A note with this PR:
- I've changed the test in `ConsoleReporterReporterTest.cs`, but I'm not sure if this is the right way to do this.  My knowledge about testing in C# is very <b> VERY </b> basic.


src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
```diff
@@ -54,5 +65,28 @@ private LogEventLevel ValidateLogLevel(string levelText)
...
+                   throw new ValidationException("The threshold can never be > 100");
+               } else if (threshold < 0)
+               {
+                   throw new ValidationException("The threshold can not be < 0");
+               }
+            }
+
+            if (thresholdList[0] >= thresholdList[1] || 
+                thresholdList[1] >= thresholdList[2] || 
+                thresholdList[0] >= thresholdList[2])
```
richardwerkman: When `0 >= 1` and `1 >= 2` , 0 is always >= 2. So the last conditional is not necessary.

src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
```diff
@@ -17,13 +17,24 @@ public class StrykerOptions
 
         public int AdditionalTimeoutMS { get; }
 
-        public StrykerOptions(string basePath, string reporter, string projectUnderTestNameFilter, int additionalTimeoutMS, string logLevel, bool logToFile)
+        public int ThresholdBreak { get; }
```
simondel: I'd prefer to store the Threshold as an object on the `StrykerOptions` class with separate properties. I would also expect this to be defined as a JSON object if you use a config file.

src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
```diff
@@ -17,13 +17,24 @@ public class StrykerOptions
...
+
+        public StrykerOptions(string basePath, string reporter, string projectUnderTestNameFilter, int additionalTimeoutMS, string logLevel, bool logToFile, 
+        int thresholdBreak, int thresholdLow, int thresholdHigh)
         {
             BasePath = basePath;
             Reporter = ValidateReporter(reporter);
             ProjectUnderTestNameFilter = projectUnderTestNameFilter;
             AdditionalTimeoutMS = additionalTimeoutMS;
             LogOptions = new LogOptions(ValidateLogLevel(logLevel), logToFile);
+            int[] ThresholdList = ValidateThresholds(new int[3]{thresholdBreak, thresholdLow, thresholdHigh});
```
simondel: I don't like the idea that you're moving the properties into an array and then retreiving them from the array again. You don't know if the `ValidateThresholds` method hasn't changed the order of the Thresholds. 

It doesn't do anything with the list so we don't need to return it. If we do have to return it, we could consider using a tuple.

Stukongeluk:
Thanks for the feedback @simondel & @richardwerkman ! The next commit will fix all these and also fix the conflicts ;)

Stukongeluk:
Merge conflicts fixed

src/Stryker.CLI/README.md
```diff
@@ -67,9 +67,21 @@ Example:
         "timeout-ms":2000,
         "log-file":true,
         "project-file":"ExampleProject.csproj",
-		"max-concurrent-test-runners": 4
+		"max-concurrent-test-runners": 4,
```
richardwerkman: Tab

Stukongeluk:
I didn't add the `threshold-break` case to the readme.md, because it still isn't implemented to let stryker exit. I think the one who picks up #121 should add this to the readme, because after that PR, it should actually work like documented.

Issue #142, opened 2018-10-04
Change --project argument to --project-file


yaelkeemink:
I would rather see '--project-file' instead of --project or --project-name as an argument since it is more descriptive.

see #129

PR #132, opened 2018-10-02
feat: replace logFile with log-level-file #99

cmulder:

PR #144, opened 2018-10-04
fix(logging):Ensure all references of --project are now --project-file

Bigtalljosh:
As per issue #142

### DIFF
--- a/src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
+++ b/src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
@@ -46,12 +46,12 @@
             JsonKey = "timeout-ms"
         };
 
-        public static readonly CLIOption<string> ProjectName = new CLIOption<string>
+        public static readonly CLIOption<string> ProjectFileName = new CLIOption<string>
         {
             ArgumentName = "--project-file",
-            ArgumentShortName = "-p <projectName>",
+            ArgumentShortName = "-p <projectFileName>",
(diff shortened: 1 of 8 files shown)


src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
```diff
@@ -48,7 +48,7 @@ public static class CLIOptions
 
         public static readonly CLIOption<string> ProjectName = new CLIOption<string>
         {
-            ArgumentName = "--project",
+            ArgumentName = "--project-file",
```
richardwerkman: Maybe it is confusing that the JsonKey is `project-name` and the argumentname is `--project-file`. I think they should be in line.

richardwerkman:
Thank you for your PR! I'm missing the update of the [readme](https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.CLI/README.md). 

It still says `--project`

src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
simondel: An example of a similar change can be found in #132. You could use that as a reference.

Bigtalljosh:
> Thank you for your PR! I'm missing the update of the [readme](https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.CLI/README.md).
> 
> It still says `--project`
😓 My mistake! Resolved now!

PR #147, opened 2018-10-04
feat(build): Add multiple framework support

### LINKED ISSUE #130: <TargetFrameworks> causes Object null reference exception
When running against a test project using `<TargetFrameworks>` instead of `<TargetFramework>` causes a `System.NullReferenceException`.

```
[14:39:28 DBG] Determining project under test with name filter null
[14:39:28 ERR] An error occurred during the mutationtest run
System.NullReferenceException: Object reference not set to an instance of an object.
   at Stryker.Core.Initialisation.ProjectFileReader.FindTargetFrameworkReference(XDocument document)
   at Stryker.Core.Initialisation.ProjectFileReader.ReadProjectFile(XDocument projectFileContents, String projectUnderTestNameFilter)
   at Stryker.Core.Initialisation.InputFileResolver.ReadProjectFile(String projectFilePath, String projectName)
   at Stryker.Core.Initialisation.InputFileResolver.ResolveInput(String currentDirectory, String projectName)
   at Stryker.Core.Initialisation.InitialisationProcess.Initialize(StrykerOptions options)
   at Stryker.Core.StrykerRunner.RunMutationTest(StrykerOptions options)
[14:39:28 INF] Time Elapsed 00:00:00.2799214
```

The only indication that the issue is the `TargetFramework` is that the stack trace fails on `FindTargetFrameworkReference`.  This could do with a better messaging if you only support targeting one framework, or better support when `TargetFrameworks` is used.

Jacrys:
-Add multiple framework support to ProjectFileReader.

-Add tests to ensure expected functionality and to ensure no regressions

Fixes #130, if a .csproj contains multiple framework support, currently Stryker breaks with a NullReference Exception.  This change inserts a stop-gap that instructs Stryker to use the first found framework.

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/ProjectFileReaderTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/ProjectFileReaderTests.cs
@@ -250,14 +250,11 @@ namespace Stryker.Core.UnitTest.Initialisation
         [InlineData("netcoreapp2.1")]
         public void ProjectFileReader_ShouldParseWhenMultipleFrameworks(string target)
         {
-            var xDocument = new XDocument();
-
-            if (target.Contains(";"))
-            {
-                xDocument = XDocument.Parse($@"
+            var xDocument = XDocument.Parse($@"
 <Project Sdk=""Microsoft.NET.Sdk"">
-    <PropertyGroup>
-        <TargetFrameworks>{target}</TargetFrameworks>
+    <PropertyGroup>{(
+        target.Contains(";") ? $"<TargetFrameworks>{target}</TargetFrameworks>" : $"<TargetFramework>{target}</TargetFramework>"
+    )}        
         <IsPackable>false</IsPackable>
     </PropertyGroup>
 
@@ -273,29 +270,6 @@ namespace Stryker.Core.UnitTest.Initialisation
     </ItemGroup>
                 
 </Project>");
-            }
-            else
-            {
-                xDocument = XDocument.Parse($@"
-<Project Sdk=""Microsoft.NET.Sdk"">
-    <PropertyGroup>
-        <TargetFramework>{target}</TargetFramework>
-        <IsPackable>false</IsPackable>
-    </PropertyGroup>
-
-    <ItemGroup>
-        <PackageReference Include=""Microsoft.NET.Test.Sdk"" Version = ""15.5.0"" />
-        <PackageReference Include=""xunit"" Version=""2.3.1"" />
-        <PackageReference Include=""xunit.runner.visualstudio"" Version=""2.3.1"" />
-        <DotNetCliToolReference Include=""dotnet-xunit"" Version=""2.3.1"" />
-    </ItemGroup>
-               
-    <ItemGroup>
-        <ProjectReference Include=""..\ExampleProject\ExampleProject.csproj"" />
-    </ItemGroup>
-                
-</Project>");
-            }
             
             var result = new ProjectFileReader().ReadProjectFile(xDocument, null);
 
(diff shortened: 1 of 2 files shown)


src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/ProjectFileReaderTests.cs
```diff
@@ -243,5 +243,63 @@ public void ProjectFileReader_FindAssemblyNameShouldBeNullWhenNotAvailable()
...
+        public void ProjectFileReader_ShouldParseWhenMultipleFrameworks(string target)
+        {
+            var xDocument = new XDocument();
+
+            if (target.Contains(";"))
+            {
+                xDocument = XDocument.Parse($@"
+<Project Sdk=""Microsoft.NET.Sdk"">
+    <PropertyGroup>
+        <TargetFrameworks>{target}</TargetFrameworks>
```
richardwerkman: Maybe it is a bit hacky but you could just ad the 's' to `<TargetFramework> conditionally instead of having to place the whole csproj file here twice. Something like 
``` xml
<{targetFrameworkName}>{target}</{targetFrameworkName}>
```

It would clean up this test a bit

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
```diff
@@ -6,75 +6,67 @@
 
 namespace Stryker.Core.Initialisation
 {
-    public class ProjectFileReader
-    {
+    public class ProjectFileReader {
```
richardwerkman: This project follows the [Microsoft Coding conventions](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions#layout-conventions). They place brackets under the class and method names.

Jacrys: Yeah, I noticed that on the second commit diff.. I am unsure what happened.  I can go through and fix it.  My VS does this even though C# is set up to place on new lines.  I have to fight it all the time, and it annoys me to no end.

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/ProjectFileReaderTests.cs
Jacrys: Sorry, I didn't know how to do conditional interpolation.  But I do now.

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
```diff
@@ -10,71 +10,75 @@ public class ProjectFileReader
...
+            var rererence = FindProjectReference (projectFileContents, projectUnderTestNameFilter);
+            var targetFramework = FindTargetFrameworkReference (projectFileContents);
+            var assemblyName = FindAssemblyName (projectFileContents);
 
-            return new ProjectFile()
+            return new ProjectFile ()
             {
                 ProjectReference = rererence,
-                TargetFramework = targetFramework
+                    TargetFramework = targetFramework
```
yaelkeemink: There is one tab too many

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
```diff
@@ -32,49 +32,56 @@ public ProjectFile ReadProjectFile(XDocument projectFileContents, string project
...
                     {
-                        throw new ArgumentException($"More than one project reference matched your --project-file={projectUnderTestNameFilter} argument to specify the project to mutate, please specify the name more detailed", innerException: new Exception($"Found the following references: {referencesString}"));
+                        throw new ArgumentException ($"More than one project reference matched your --project-file={projectUnderTestNameFilter} argument to specify the project to mutate, please specify the name more detailed", innerException : new Exception ($"Found the following references: {referencesString}"));
                     }
-                    return searchResult.Single();
+                    return searchResult.Single ();
                 }
             }
-            else if (!projectReferences.Any())
+            else if (!projectReferences.Any ())
```
Mobrockers: Something in your linter/formatter seems to have added a space behind almost every method call in this file. I would suggest you turn off this behaviour in your linter as it's contaminating the pull request.

Jacrys: I apologize for the large number of issues. I have copied the code over, made my modifications again, then diffed the files. They seem to be identical save for the lines I added.

Mobrockers: No problem, we appreciate the work

PR #148, opened 2018-10-04
fix(mutating): Make sure a broken mutation does not exist twice, so that we can roll back the mutation (#145)

### LINKED ISSUE #145: Some statements break Stryker.NET
### The issue
The following statement breaks Stryker:

``` csharp
string someQuery = "test";
new Uri(new Uri(""), "/API?" + someQuery);
```

The reason is that changing the `+` to an `-` will not cause one but two compile errors.

1. `cannot convert from 'System.Uri' to 'string'`
2. `Operator '-' cannot be applied to operands of type 'string' and 'string'`

During the rollback stage Stryker will try to rollback both errors. It will find the mutation that causes this twice and will try to remove that node twice. The first time is succesful, the second time the node is already removed and will throw a null reference exception. This crashes Stryker.

### Fix
https://github.com/stryker-mutator/stryker-net/blob/f5395ae6bdeb38b4aadc84dbaac9caee23a4e699/src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs#L91
The mutation made in the causing statement exists twice in the `brokenMutations` Collection. Somehow we have to make sure the list cannot contain a mutation twice. 

This will fix the bug.

B1tF8er:
Fix #145

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
+++ b/src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
@@ -98,7 +98,7 @@ namespace Stryker.Core.Compiling
                     _logger.LogError("Unable to rollback mutation for node {0} with diagnosticmessage {1}", brokenMutation, diagnostic.GetMessage());
                 }
 
-                if (!brokenMutation.Contains(mutationIf))
+                if (!brokenMutations.Contains(mutationIf))
                 {
                     brokenMutations.Add(mutationIf);
                 }
--- a/src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
+++ b/src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
@@ -97,7 +97,9 @@ namespace Stryker.Core.Compiling
                 {
                     _logger.LogError("Unable to rollback mutation for node {0} with diagnosticmessage {1}", brokenMutation, diagnostic.GetMessage());
                 }
-                brokenMutations.Add(mutationIf);
+
+                if (!brokenMutation.Contains(mutationIf))
+                    brokenMutations.Add(mutationIf);
             }
             // mark the if statements to track
             var trackedTree = rollbackRoot.TrackNodes(brokenMutations);
--- a/src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
+++ b/src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
@@ -99,7 +99,9 @@ namespace Stryker.Core.Compiling
                 }
 
                 if (!brokenMutation.Contains(mutationIf))
+                {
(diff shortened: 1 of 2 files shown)


simondel:
Thanks for the PR! Could you add braces around the if-statement to prevent possible issues in the future?

How do you think we could write a test for this?

B1tF8er:
> Thanks for the PR! Could you add braces around the if-statement to prevent possible issues in the future?
> 
> How do you think we could write a test for this?

Sure let me add braces around the if-statement 

Let me think of a possible test to cover the code

Mobrockers:
@richardwerkman I noticed on [this](https://github.com/stryker-mutator/stryker-net/blob/f5395ae6bdeb38b4aadc84dbaac9caee23a4e699/src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs#L73) line we add the id of the found mutation node to a list called  rollbackedIds. Since the findMutationIf call is done before adding the brokenmutation collection we will still get duplicate mutation ids in the rollbackedids list. Is this a problem?

richardwerkman:
Yeah I don't like the idea that the id will be in the list twice. The count of rollbacked id's is logged later on and that number would be incorrect this way. Maybe we could use a `.Distinct()` to fix this?

Mobrockers:
I think the logic should rather be moved to inside the if statement for RemoveMutantIfStatements -> brokenMutations.Add(mutationIf); because I don't think it makes sense to set the mutation as rolled back before the node is actually removed.

B1tF8er:
Yeah let me work on that

simondel:
PS: If you run into any issues with the build, please merge the current master branch into your branch.

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
```diff
@@ -97,7 +97,11 @@ private SyntaxTree RemoveMutantIfStatements(SyntaxTree originalTree, ICollection
                 {
                     _logger.LogError("Unable to rollback mutation for node {0} with diagnosticmessage {1}", brokenMutation, diagnostic.GetMessage());
                 }
-                brokenMutations.Add(mutationIf);
+
+                if (!brokenMutations.Contains(mutationIf))
```
lkts: i see that `brokenMutations` is a collection and here `Contains` is called in loop, maybe it should be a HashSet?

richardwerkman: I'm not sure if that would make much of a difference here. The brokenMutations collection normally contains from 1 to maybe 10 items. If there would be more items this would be a good improvement!

Mobrockers:
@B1tF8er How is it going, need any help with this? :)

B1tF8er:
> @B1tF8er How is it going, need any help with this? :)

Hello, sorry I wont be able to continue with this work I will close the PR

PR #149, opened 2018-10-04
fix(dependency-resolver): Support backslash in ProjectReferences on Linux

### LINKED ISSUE #120: Support backslash in ProjectReference on Linux
When backslash (`\`) is used as folder separator in `csproj` (which is inserted by Visual Studio) when referencing another project in a solution, `FileNotFoundException` is thrown.

`csproj` snippet:

```xml
  <ItemGroup>
    <ProjectReference Include="..\..\src\Lib\Lib.csproj" />
  </ItemGroup>
```

Output:

```
Analyzing project
[12:21:49 INF] Using /builds/mutation-testing/tests/Lib.Tests/Lib.Tests.csproj as project file
[12:21:49 ERR] An error occurred during the mutationtest run 
System.IO.FileNotFoundException: Could not find file '/builds/mutation-testing/tests/Lib.Tests/..\..\src\Lib\Lib.csproj'.
File name: '/builds/mutation-testing/tests/Lib.Tests/..\..\src\Lib\Lib.csproj'
   at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirectory, Func`2 errorRewriter)
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String path, OpenFlags flags, Int32 mode)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
   at System.IO.StreamReader..ctor(String path, Encoding encoding, Boolean detectEncodingFromByteOrderMarks, Int32 bufferSize)
   at System.IO.File.OpenText(String path)
   at Stryker.Core.Initialisation.InputFileResolver.FindProjectUnderTestAssemblyName(String projectFilePath)
   at Stryker.Core.Initialisation.InputFileResolver.ResolveInput(String currentDirectory, String projectName)
   at Stryker.Core.Initialisation.InitialisationProcess.Initialize(StrykerOptions options)
   at Stryker.Core.StrykerRunner.RunMutationTest(StrykerOptions options)
[12:21:49 INF] Time Elapsed 00:00:00.1874362
```

## Workaround

Converting to slashes by hand fixes the issue.

kevinlyles:
I decided to try running Stryker.NET on Linux, and it turned out to be really close to ready.  This pull request fixes what looks like the last remaining issue, which is that project file paths were always returned with the Windows directory separators. Fixes #120

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
+++ b/src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
@@ -1,6 +1,7 @@
 ﻿using Microsoft.Extensions.Logging;
 using Stryker.Core.Logging;
 using System;
+using System.IO;
 using System.Linq;
 using System.Xml.Linq;
 
@@ -19,13 +20,13 @@ namespace Stryker.Core.Initialisation
         {
             _logger.LogDebug("Reading the project file {0}", projectFileContents.ToString());
 
-            var rererence = FindProjectReference(projectFileContents, projectUnderTestNameFilter);
+            var reference = FindProjectReference(projectFileContents, projectUnderTestNameFilter);
             var targetFramework = FindTargetFrameworkReference(projectFileContents);
             var assemblyName = FindAssemblyName(projectFileContents);
 
             return new ProjectFile()
             {
-                ProjectReference = rererence,
+                ProjectReference = reference,
                 TargetFramework = targetFramework
             };
         }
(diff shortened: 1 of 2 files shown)


Mobrockers:
lgtm 🐝

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
```diff
@@ -49,22 +50,30 @@ private string FindProjectReference(XDocument document, string projectUnderTestN
...
+            return ConvertPathSeparators(projectReferences.Single());
+        }
 
+        private static string ConvertPathSeparators(string filePath)
+        {
+            if (Path.DirectorySeparatorChar == '\\')
+            {
+                return filePath;
+            }
+            return filePath.Replace('\\', Path.DirectorySeparatorChar);
```
simondel: Could you either add an `else`-statement here or use an elvis-operator to make it more clear that the last return is the else-statement?

Also, it may be nicer to store `'\\'` in a variable at the start of the method to ensure that we use the same value throughout the method.

yaelkeemink:
Looks good to me, @richardwerkman do you agree with merging?

Issue #156, opened 2018-10-05
Happy flow tests


richardwerkman:
We should add a integration test project that checks the happy flow of Stryker. This will prevent broken releases like 0.2.0. 

For instance it could create an instance of `StrykerCLI` and run it on a small example project. The exit code of stryker should be correct. Maybe we could also test more, like expected output or mutationscore. 

Anyone with experience writing these kind of tests that can give some advise on how to set this up?

simondel:
@richardwerkman @Mobrockers The integration test takes 30 minutes. Running the project itself in a regular scenario is a lot faster:
https://dev.azure.com/stryker-mutator/Stryker/_build/results?buildId=415&_a=summary&view=logs

PR #163, opened 2018-10-06
feat(integrationtest): Add happyflow integration test

richardwerkman:
I added a integrationtest project that runs the stryker source code on a small project. Right now it creates 15 mutants of different kinds and checks if the status code is correct. Therefore this PR depends on #154. The testrun takes around 50 seconds on my machine. If this is too long we can change the exampleproject to be smaller. We should add this test to run on the build server.


nicojs:
Great work! Love to see some integration tests. These should indeed run in the CI build (and when we publish a new version).

Is there a technical reason to keep the solution files separated? Seems strange to me. Having them all in 1 solution seems easier to manage.

Can we validate that reverting #157 results in breaking these tests?

richardwerkman:
@nicojs 
> Is there a technical reason to keep the solution files separated? Seems strange to me. Having them all in 1 solution seems easier to manage.

I created the new solution because the integrationtest takes quite long. This makes running the unit tests using "run all tests" inside visual studio very slow. This would also affect live unit testing. For me personally the integration test doesn't have to run while developing. I just want it to run when creating a PR to see if the happy flow still works.

> Can we validate that reverting #157 results in breaking these tests?

Not yet, I assert that the exit code is correct after the happy flow. Right now the PR (#154) that sets this exit code is not merged yet. So we should wait for that first.

richardwerkman:
closes #156

richardwerkman:
@nicojs the integration test now fails when an exception occurs while running the program. But I noticed the azure pipeline job now almost takes 40 minutes. Is this okay? The appveyor job only took a few minutes.

integrationtests/IntegrationTests/Stryker.Core.IntegrationTest/ExampleProject/ExampleProject.XUnit/DymmyBooleanTest.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Xunit;
+
+namespace ExampleProject.XUnit
+{
+    public class DymmyBooleanTest
```
Mobrockers: Class and file name are incorrect

integrationtests/IntegrationTests/Stryker.Core.IntegrationTest/ExampleProject/ExampleProject.XUnit/DummyCalcTest.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Xunit;
+
+namespace ExampleProject.XUnit
+{
+    public class DummyCalcTest
```
Mobrockers: Are these tests commented out on purpose?

integrationtests/IntegrationTests/Stryker.Core.IntegrationTest/ExampleProject/ExampleProject.XUnit/Endlesslooptest.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Xunit;
+
+namespace ExampleProject.XUnit
+{
+    public class Endlesslooptest
```
Mobrockers: Endlesslooptest should be EndlessLoopTest

integrationtests/IntegrationTests/Stryker.Core.IntegrationTest/ExampleProject/ExampleProject/Endlessloop.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace ExampleProject
+{
+    public class Endlessloop
```
Mobrockers: EndlessLoop

simondel:
It shouldn't take that long. The build in my previous comment runs the exact same project, only then from the actual NuGet package

Issue #159, opened 2018-10-06
Default loglevel info


richardwerkman:
I was just thinking about how we could clean the repo a bit and noticed the status reporter might not be needed anymore.

## Default behaviour

```
Analyzing project
Building project
Starting initial testrun
Project OK
Generating mutants

1 mutants have been created. Each mutant will now be tested, this could take a while.
```
## --log-console info behaviour
```
Analyzing project
[11:47:42 INF] Using C:\Dev\ExampleProjects\Example2.1\ConsoleApp1\XUnitTestProject1\XUnitTestProject1.csproj as project file
Building project
[11:47:42 INF] Starting initial build
[11:47:43 INF] Initial build successful
[11:47:44 INF] Using testrunner DotnetTestRunner
Starting initial testrun
[11:47:44 INF] Initial testrun started
[11:47:47 INF] Initial testrun successful in 2196 ms
[11:47:47 INF] Using 33294 ms as testrun timeout
Project OK
Generating mutants
[11:47:47 INF] 1 mutants created
[11:47:49 INF] 1 mutants ready for test

1 mutants have been created. Each mutant will now be tested, this could take a while.
```
In the second example we see a lot of double information. By making the default loglevel "information" we could clean the repo quite a bit. This would mean:
* The [status reporter](https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/Reporters/ConsoleStatusReporter.cs) could be removed.
* Some methods on the [reporter interface](https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/Reporters/IReporter.cs) can be removed, which makes the other reporters a lot cleaner.
* The "ReportOnly" reporter setting will not be needed anymore, since users can just set the loglevel to warning or error to accomplish that.

@Mobrockers @simondel @yaelkeemink what do you think about this? Would this be a good issue for hacktoberfest?

Mobrockers:
I agree. We had decided the reportonly reporter would be removed in the future anyway.

lkts:
Let me take a look at this.

PR #167, opened 2018-10-07
change(reporting): Set default log level to Info and cleanup reporters

lkts:
Fixes https://github.com/stryker-mutator/stryker-net/issues/159

### DIFF
--- a/src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
+++ b/src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
@@ -23,7 +23,7 @@
         {
             ArgumentName = "--log-console",
             ArgumentShortName = "-l <logLevel>",
-            ArgumentDescription = "Sets the logging level | Options [error, warning (default), info, debug, trace]",
+            ArgumentDescription = "Sets the logging level | Options [error, warning, info (default), debug, trace]",
             DefaultValue = "Info",
             JsonKey = "log-level"
         };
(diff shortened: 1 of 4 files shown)


src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
```diff
@@ -24,7 +24,7 @@ public static class CLIOptions
             ArgumentName = "--log-console",
             ArgumentShortName = "-l <logLevel>",
             ArgumentDescription = "Sets the logging level | Options [error, warning (default), info, debug, trace]",
```
richardwerkman: info should say (default) now

src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs
```diff
@@ -41,12 +41,10 @@ public class InitialisationProcess : IInitialisationProcess
 
         public MutationTestInput Initialize(StrykerOptions options)
         {
-            _reporter.OnInitialisationStarted();
```
richardwerkman: I don't think the _reporter property is needed at all anymore here

richardwerkman:
Thank you for your pull request! I have few remarks.

src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
lkts: indeed!

src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs
lkts: right

PR #175, opened 2018-10-09
feat(logging) : Progress reporter

Szpi:
Implementation of issue #9 
View od progress report:
new


richardwerkman:
First of all, awesome job on the progress bar! It looks really good, even better than I first imagined. Reviewing this could take a while. 

Some first remarks: 
* Using the logger to display the progress has some advantages and disadvantages. 
  * It gives nice colours which I like
  * It makes the timestamps appear on the left side which I don't like.

Maybe we could take advantage of the logger without the timestamp. I suggets creating a specific serilog logger instance for the progresbar reporter that uses a different [template](https://github.com/serilog/serilog/wiki/Formatting-Output).

* The time estimation is not acurate. When running Stryker on Stryker it predicted 4 minutes at first, while the run actually took 1m57 total (with initital build and testrun).

This probably has to do with the fact that I have 6 cores. I guess the estimation calculation doesn't take parallel testing in account? This should be improved.

* The progress bar is only visable using --log-console=info

This has to do with the first remark as well. I'm not sure if the progress bar should belong to any loglevel at all. It should just show when the user sets the reporter to be "progressbar". I'm also not sure if we can accomplish this using serilog.

* The progress bar and the ConsoleDotReporter are both on now.

It would be nice if configuring the reporters could be fixed within this PR, but I can understand if you would like to create a new issue for that, so someone else can take that of your hands.

Szpi:
> The progress bar and the ConsoleDotReporter are both on now.
I think it is better to create another story for that.

> The time estimation is not acurate. When running Stryker on Stryker it predicted 4 minutes at first, while the run actually took 1m57 total (with initital build and testrun).

Honestly I would remove time estimation, because it will never be good enough. We could divide time by threads to get more accurate result.

I can work on improvement about logging.

Mobrockers:
I agree that we should perhaps remove the time estimation, it's not really that useful and it's impossible to get right.

richardwerkman:
Removing time estimation would be fine for me. It would be a cool feature, but this can also be done as a seperate feature. I've not found any time to look at the code, maybe @Mobrockers has some time for this?

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
```diff
@@ -38,20 +41,34 @@ public class MutationTestProcess : IMutationTestProcess
         public MutationTestProcess(MutationTestInput mutationTestInput,
             IReporter reporter,
             IEnumerable<IMutator> mutators,
-            IMutationTestExecutor mutationTestExecutor,
-            IMutantOrchestrator orchestrator = null,
+            IMutationTestExecutor mutationTestExecutor, IMutantOrchestrator orchestrator = null,
```
Mobrockers: This should be back on it's own line

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
```diff
@@ -38,20 +41,34 @@ public class MutationTestProcess : IMutationTestProcess
...
 
+        private static ProgressReporter CreateProgressReporter()
+        {
+            var consoleOneLineLoggerFactory = new ConsoleOneLineLoggerFactory();
+            var progressBarReporter = new ProgressBarReporter(consoleOneLineLoggerFactory.Create());
+            var mutantsResultReporter = new MutantsResultReporter(
+                consoleOneLineLoggerFactory.Create(),
+                consoleOneLineLoggerFactory.Create(),
+                consoleOneLineLoggerFactory.Create(),
+                consoleOneLineLoggerFactory.Create());
```
Mobrockers: Could you use named params here so it is more clear why we're creating five different loggers and passing them to the same reporter?

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
```diff
@@ -106,20 +124,29 @@ public void Mutate()
...
 
             if (options.MaxConcurrentTestrunners <= logicalProcessorCount)
             {
                 usableProcessorCount = options.MaxConcurrentTestrunners;
             }
 
-            Parallel.ForEach(_input.ProjectInfo.ProjectContents.Mutants.Where(x => x.ResultStatus == MutantStatus.NotRun),
+            var mutantsNotRun = _input.ProjectInfo.ProjectContents.Mutants.Where(x => x.ResultStatus == MutantStatus.NotRun).ToList();
+
+            _progressReporter.ReportInitialState(mutantsNotRun.Count());
```
richardwerkman: I think the reporter should keep track of this count. It knows the total amout of mutations and gets notified on every mutant tested. 

We could add a ` _reporter.OnStartMutantTestRun(mutant);` to keep track of the number of not runned mutants, running mutants and runned mutants.

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
```diff
@@ -38,20 +41,34 @@ public class MutationTestProcess : IMutationTestProcess
...
             _reporter = reporter;
             _mutationTestExecutor = mutationTestExecutor;
+            _progressReporter = progressReporter ?? CreateProgressReporter();
             _orchestrator = orchestrator ?? new MutantOrchestrator(mutators);
             _compilingProcess = compilingProcess ?? new CompilingProcess(mutationTestInput, new RollbackProcess());
             _fileSystem = fileSystem ?? new FileSystem();
             _logger = ApplicationLogging.LoggerFactory.CreateLogger<MutationTestProcess>();
         }
 
+        private static ProgressReporter CreateProgressReporter()
```
richardwerkman: In my opinion this should be placed inside the reporter factory. This makes (runtime) switching between reporters more easy.

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
```diff
@@ -106,20 +124,29 @@ public void Mutate()
...
-            Parallel.ForEach(_input.ProjectInfo.ProjectContents.Mutants.Where(x => x.ResultStatus == MutantStatus.NotRun),
+            var mutantsNotRun = _input.ProjectInfo.ProjectContents.Mutants.Where(x => x.ResultStatus == MutantStatus.NotRun).ToList();
+
+            _progressReporter.ReportInitialState(mutantsNotRun.Count());
+
+            Parallel.ForEach(mutantsNotRun,
                 new ParallelOptions { MaxDegreeOfParallelism = usableProcessorCount },
                 mutant =>
                 {
+                    var timer = new Stopwatch();
```
richardwerkman: If we won't estimate time left, the custom progress reporter calls can be removed here. Same for the timer.

src/Stryker.Core/Stryker.Core/Reporters/Progress/ConsoleOneLineLogger.cs
```diff
@@ -0,0 +1,48 @@
...
+            _logger.LogInformation(text, args);
+        }
+
+        public void ReplaceLog(string text, params object[] args)
+        {
+            var currentCursorTop = Console.CursorTop;
+            var currentCursorLeft = Console.CursorLeft;
+
+            Console.SetCursorPosition(0, _cursorTop);
+            ClearLine();
```
richardwerkman: I think we could clear only the part till currentCursorLeft. This way not the whole line has to be cleared.

src/Stryker.Core/Stryker.Core/Reporters/Progress/ProgressReporter.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Stryker.Core.Reporters.Progress
+{
+    public interface IProgressReporter
+    {
+        void ReportRunTest(TimeSpan testDuration, IReadOnlyMutant mutantTestResult);
+        void ReportInitialState(int totalNumberOfTests);
+    }
+
+    public class ProgressReporter : IProgressReporter
```
richardwerkman: Why did you create a new interface for this reporter? The current IReporter interface would also work here, making switching reporters at runtime possible.

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
Szpi: Fixed

src/Stryker.Core/Stryker.Core/Reporters/Progress/ConsoleOneLineLogger.cs
Szpi: I think in that case we would have to keep track of the original cursorLeft position, the line that we want to clear not current one. I would keep it as it is, because I don't think it has so much impact on performance changing it would bring in additional complexity / bugs.

Issue #174, opened 2018-10-09
Line number in raport is wrong


richardwerkman:
The line numbers of mutations are displayed incorrect. They seem to be always +1. This can be confusing for users as they cannot find the right line when looking for the mutation in their code.

```
[Survived] Binary expression mutation on line 39: 'request.Limit.Value <= 0' ==> 'request.Limit.Value < 0'
```
Can be found on line 40 in the source code. I'm not sure where the wrong number comes from.

nicojs:
There are 3 hard problems in software development: cache invalidation and off-by-one errors. :trollface:

Szpi:
I can take this issue.

Szpi:
PR #176

simondel:
Closed with #176

PR #176, opened 2018-10-09
fix(bug): Wrong line number

Szpi:
// user-visible line and column counts are 1-based, but internally are 0-based.
Comment from [Location.cs](https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/Portable/Diagnostic/Location.cs) from roslyn github see GetDebuggerDisplay method. We need to increment Line by one 🥇.


richardwerkman:
Thanks for taking care of this as well! This looks fine.

simondel:
Closes #174

PR #184, opened 2018-10-12
fix(config): Simplify CLI option creation

KonH:
I noticed recurrent pattern in src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs to create app options.
With proposed changes we have more readable logic of intention, less code code redundancy and get ability to change it in one place, if it required.


KonH:
Also, if anyone knows how to avoid including commits which doesn't have actual changes to PR (it flows from my fork' master branch), I would be grateful. Or I need to re-create the fork and make changes only in separated branches to get it?

simondel:
This won't be an issue because when we merge all your commits will be squashed into a single commit :)

src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
```diff
@@ -6,6 +6,19 @@
 
 namespace Stryker.CLI
 {
+    internal static class Helpers
```
richardwerkman: I don't think this has to be a static helper class. This could be a private method of the class `StrykerCLI`

richardwerkman:
I like your improvement! This will be merged when the requested change has been made.

PR #185, opened 2018-10-14
feat(linq mutators): Add Mutations for LINQ Expressions

eTomate:
It looks like the mutations for different kinds of linq queries are up and running.
I've encountered some issues in the rollback process, where I'm not entirely sure what happens there.

It looks like - somehow - in `RemoveMutantIfStatements(SyntaxTree originalTree, ICollection<Diagnostic> diagnosticInfo)` when iterating over the broken mutations, the nodeToRemove can be null. This crashed the whole process, so I've added an if clause and logging there.


src/Stryker.Core/Stryker.Core/Enumerations/LinqExpression.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Stryker.Core.Enumerations
+{
+    /// <summary>
+    ///     Enumeration for the different kinds of linq expressions
+    /// </summary>
+    public enum LinqExpression
```
richardwerkman: Do we need a global enumeration for this? I think it's only used inside the LinqMutator. So it could be placed inside the LinqMutator file or in the same folder to keep the folder structure clean.

src/Stryker.Core/Stryker.Core/Mutators/LinqMutator.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        /// <summary>
+        ///     Apply mutations to an <see cref="IdentifierNameSyntax"/>
+        /// </summary>
+        /// <param name="node"></param>
+        /// <returns></returns>
+        public override IEnumerable<Mutation> ApplyMutations(IdentifierNameSyntax node)
+        {
+            if (Enum.TryParse(node.Identifier.ValueText, out LinqExpression expression) &&
+                _kindsToMutate.TryGetValue(expression, out LinqExpression replacementExpression))
```
richardwerkman: Can this ever be false? I think this can be placed inside the ifstatement and maybe doesn't have to be a Try method?

src/Stryker.Core/Stryker.Core/StrykerRunner.cs
```diff
@@ -62,7 +62,8 @@ public void RunMutationTest(StrykerOptions options)
                             new AssignmentStatementMutator(),
                             new PrefixUnaryMutator(),
                             new PostfixUnaryMutator(),
-                            new CheckedMutator()
+                            new CheckedMutator(),
+                            new LinqMutator()
```
richardwerkman: When adding a mutator I expect the documentation to be updated. It should be included in the list of mutators on both the main readme and for the Core project readme.

richardwerkman:
Thank you for your quick contribution! It looks good but I have some remarks.

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
```diff
@@ -105,8 +106,16 @@ private SyntaxTree RemoveMutantIfStatements(SyntaxTree originalTree, ICollection
             {
                 // find the if statement in the new tree
                 var nodeToRemove = trackedTree.GetCurrentNode(brokenMutation);
-                // remove the if statement and update the tree
-                trackedTree = trackedTree.ReplaceNode(nodeToRemove, nodeToRemove.Else.Statement);
+
+                if (nodeToRemove != null)
```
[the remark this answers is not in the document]
eTomate: Possible - but I can't completely test my solution, if I don't add it. What do you say? :)

src/Stryker.Core/Stryker.Core/Enumerations/LinqExpression.cs
eTomate: I'll move it to the LinqMutator.

src/Stryker.Core/Stryker.Core/StrykerRunner.cs
eTomate: Yep - I've overlooked that part.

src/Stryker.Core/Stryker.Core/Mutators/LinqMutator.cs
eTomate: It can indeed throw an exception, whenever some NameSyntax equals to None which is no key in the enum.
I've just added the None, since I overlooked the Removal of some Linq Queries before.

eTomate:
I‘ve updated the PR - but I’m still unsure about the nullcheck - should I keep it or remove it?

src/Stryker.Core/Stryker.Core/Mutators/LinqMutator.cs
```diff
@@ -0,0 +1,118 @@
...
+            {
+                SyntaxNode replacement = SyntaxFactory.IdentifierName(replacementExpression.ToString());
+                string displayName = $"{node.Identifier.ValueText} to {replacement} mutation";
```
richardwerkman: The mutation name is a bit unclear. For example:

`[Survived] First to Last mutation on line 12: 'First' ==> 'Last'`

Maybe this could be:

`[Survived] Linq method mutation (First() to Last()) on line 12: 'First' ==> 'Last'`

richardwerkman:
@eTomate can you confirm you can now (after merging master) run stryker.net on your project without failing? It would be nice to know for sure the bug has been fixed correctly and you can run on your project as well.

eTomate:
Hey @richardwerkman - works perfectly now!

nicojs:
Awesome! Can these be documented in the handbook?

PR #190, opened 2018-10-17
fix(mutating): Make sure a broken mutation does not exist twice, so that we can roll back the mutation (#145)

### LINKED ISSUE #145: Some statements break Stryker.NET

Mobrockers:
Reopen PR so someone else can pick up the changes that were left since @B1tF8er no longer has time to fix this. No sense throwing away the changes already made.

Fixes #145 Reopened from #148

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Compiling/RollbackProcessTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Compiling/RollbackProcessTests.cs
@@ -248,7 +248,7 @@ namespace ExampleProject
         }
 
         [Fact]
-        public void RollbackProcess_ShouldRollbackError_RollbackedCompilationShouldCompileWhen()
+        public void RollbackProcess_ShouldRollbackError_RollbackedCompilationShouldCompileWhenUriIsEmpty()
         {
             var syntaxTree = CSharpSyntaxTree.ParseText(@"
 using System;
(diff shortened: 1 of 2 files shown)


Mobrockers:
@B1tF8er Thank you for the help so far :) We'll pick it up from here and incorporate the work you already did for us

Mobrockers:
This PR should not be merged until the requested changes in #148 have been implemented and right now serves as a WIP so the branch where the current changes were made on is visible.

richardwerkman:
I'll look at this tonight

PR #194, opened 2018-10-19
feat(string mutators): add string and interpolated string mutators wi…

### LINKED ISSUE #118: String mutator
A mutator should be added that mutates strings into other strings.

Original | Mutated
| - | - |
`"foo"` (non-empty string) | `""` (empty string)
`""` (empty string) | `"Stryker was here!"`
`$"foo {bar}"` (string interpolation) | `$""` 

The mutator should look for StringLiteralExpression nodes

Valdas3:
…th tests, update README.md

Resolves #118

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ Right now, Stryker.NET supports the following mutators:
 - Update Operators
 - Checked Mutator
 - LINQ Mutator
+- String Mutator
+- Interpolated String Mutator
 
 For the full list of all available mutators, see the [Stryker.Core readme](/src/Stryker.Core/README.md).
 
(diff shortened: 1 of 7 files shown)


Mobrockers:
In my opinion whether we list them as separate mutators or not depends on if it makes sense to enable/disable one and not the other. I would think it does not, and should thus be the same mutator in the readme and configuration (when we have this) with both examples provided.

Valdas3:
I imagine that if more interpolated string mutations would be added in the future (like removing individual interpolation expressions inside the string) then it might make sense to have more granular control. But I agree, at this point the distinction between the mutators is not really important or visible.

PR #197, opened 2018-10-22
fix(test): Run unit tests on all platforms

kevinlyles:
Re-enable all unit tests on all platforms
Fix resulting directory separator issues
Switch to Path.Combine when that's easier to use

I noticed that some unit tests didn't run on Linux (or any non-Windows platform).  I re-enabled them and fixed things up until they passed.


kevinlyles:
I'd like feedback on Utils.cs -- I couldn't think of a better place to stick the helper function (it's used in multiple places now), but I don't mind moving it somewhere else if anyone has a better idea.

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/AssemblyReferenceResolverTests.cs
```diff
@@ -39,16 +39,16 @@ public void ShouldAnalyzeProjectDependencies()
...
                 null,
-                It.IsAny<int>()), 
+                It.IsAny<int>()),
                 Times.Once);
             foundReferences.ShouldBeSubsetOf(new List<string>()
             {
-                Path.Combine($@"C:{Path.DirectorySeparatorChar}Repos", "ExampleProject", "ExampleProject", "bin", "Debug", "netcoreapp2.0", "ExampleProject.Contracts.dll"),
-                Path.Combine($@"C:{Path.DirectorySeparatorChar}Program Files", "dotnet", "sdk", "NuGetFallbackFolder", "microsoft.netcore.app", "2.0.0", "ref", "netcoreapp2.0", "Microsoft.CSharp.dll"),
-                Path.Combine($@"C:{Path.DirectorySeparatorChar}Program Files", "dotnet", "sdk", "NuGetFallbackFolder", "microsoft.aspnetcore.mvc.formatters.json", "2.0.2", "lib", "netstandard2.0", "Microsoft.AspNetCore.Mvc.Formatters.Json.dll")
+                Path.Combine("C:", "Repos", "ExampleProject", "ExampleProject", "bin", "Debug", "netcoreapp2.0", "ExampleProject.Contracts.dll"),
```
Mobrockers: This unit test is failing on windows due to this Path.Combine. The expected values are C:Repos but the actual values are C:\Repos

Mobrockers: This unit test is failing on windows due to this Path.Combine. The expected values are C:Repos but the actual values are C:\Repos

Mobrockers: This unit test is failing on windows because C:Repos\ is expected but the actual file path value on windows is C:\Repos\

kevinlyles: Interesting.  I'll admit I didn't try it on Windows, and the GitHub outage meant I couldn't see the failure right until today.

Is C:Repos correct?  If so, what does that represent?  Also, what happens to {Path.DirectorySeparatorChar} in the old code that causes it to disappear?

kevinlyles: I've tried this now in a Windows VM, and all tests pass for me in Visual Studio.  I see that this test fails in the automated test runs on GitHub, but I'm not sure what would be causing the difference.

Mobrockers: I just tried it on my local system and the test fails there as well. For some reason on windows the DirectorySeparatorChar is not automatically added to the drive letter even though according to the docs it should be. https://docs.microsoft.com/en-us/dotnet/api/system.io.path.combine?view=netframework-4.7.2#System_IO_Path_Combine_System_String_System_String_

kevinlyles: Windows 7 seems to behave like Linux.  I don't have another version available to try things on, so I'll just revert that part.

PR #202, opened 2018-10-23
build: Add mac os build

### LINKED ISSUE #198: Add hosted mac agent
We currently run builds and tests on windows and ubuntu. There are probably more developers using mac's than ubuntu, so it might be useful to also run the build on mac hosted agents so we catch bugs earlier.

Mobrockers:
Fixes #198


richardwerkman:
Looks good! I now see three jobs on pipelines and all tests are green 👍

PR #205, opened 2018-10-24
WIP: Investigate auto-changelog generation

Mobrockers:


simondel:
Nice! How is this generated?

Mobrockers:
I am using the conventional-changelog npm package because it seems to work for strykerJS.

I'm using the following command

`conventional-changelog -p angular -i ./src/Stryker.CLI/CHANGELOG.md -s --commit-path ./src/Stryker.CLI`

However it does not yet use the correct format. It does not seem to know what to do with the # Change Log header.

simondel:
Awesome :) could you add it to a script in this PR? We could always extend it in the future to automate preparing a release even further.

Mobrockers:
That is the plan yea

richardwerkman:
@Mobrockers I've continued on your work to improve the automatic changelogs. 

```
conventional-changelog -p angular -i ./src/Stryker.CLI/CHANGELOG.md -s --commit-path ./src/Stryker.CLI -t StrykerMutator.DotNetCoreCli@
conventional-changelog -p angular -i ./src/Stryker.Core/CHANGELOG.md -s --commit-path ./src/Stryker.Core -t StrykerMutator.Core@
```

Using the -t command the right tag prefixes can be configured. This way only the changes since the last git tag will be added to the changelog. Where will we put these commands in our project? In a `package.json` file? Or in the `azure-pipelines.yml`?

Mobrockers:
It should be in the pipeline imo

simondel:
Since preparing a release consists of more steps than just updating the changelog, I'd suggest putting the commands in a script for now (powershell for example). 

The advantage of the script over using a pipeline directly is that you can also run it locally if you want and execute in a pipeline as well.

We can run node modules using `npx conventional-changelog`

richardwerkman:
The script works pretty good now. But it seems to skip some commits. I'm not sure why. For example the string mutator commit. The commit name seems to be in the right format but it doesn't apear in the changelog. I've updated the changelog manually for now, but it would be nice if it gets all commits next time.

richardwerkman:
Can this PR be closed now? The scripts has been implemented.

Mobrockers:
Yep

richardwerkman:
Closed by #206

PR #206, opened 2018-10-27
Publish 0.4.0

richardwerkman:
Publish

- StrykerMutator.Core@0.4.0
- StrykerMutator.DotNetCoreCli@0.4.0

Add changelog command in package.json file.


simondel:
@richardwerkman How did you generate the version number of the new package? (And set it in the changelog)

richardwerkman:
@simondel I had to change the changelog manually a bit before it looked like this. I put in the version numbers myself. Also fixed some naming and a missing feature.

simondel:
Good to know! I'll take a look at how Lerna does it and else I'll just script it myself

simondel:
I've managed to fix it. I'll update this PR tomorrow.

simondel:
I've executed the script I've written on a branch. You can view the result here:
https://github.com/stryker-mutator/stryker-net/commit/8a1e80b547f3f984e52d5d107882771931880a0f
And here:
https://github.com/stryker-mutator/stryker-net/releases

simondel:
@richardwerkman Some features/bugfixes are skipped because the feature-name already exists. For example. The LINQ mutator and the string mutator commits both specify the feature `mutators`, so only the first commit for the feature is shown on the changelog. I think we'll have to manually release one more time (can be done with the script though).

nicojs:
@simondel 
> The LINQ mutator and the string mutator commits both specify the feature `mutators`, so only the first commit for the feature is shown on the changelog.

Is this also how Lerna does it? Would explain how sometimes features don't end up in the changelog.

nicojs:
That's so frustrating. Just to be clear. If you have 2 commit messages: `fix(core): fix issue 1` and `fix(core): fix issue 2` only "fix issue 1" is prepended in the changelog?

simondel:
Yes, that's the behaviour that I'm noticing in this changelog. Perhaps the Lerna mode that it offers fixes this, but I'm not sure.

simondel:
@richardwerkman my scripting has to be reviewed and we have to manually fix the changelog this time.

simondel:
I have updated the changelog. If we merge anything to master in before this is merged, we have to update it again.

PR #207, opened 2018-10-28
Conditional mutant placer

### LINKED ISSUE #17: Ternary statement mutations
At the moment, mutations are being placed inside if statements like this

``` csharp
if(Environment.GetEnvironmentVariable("ActiveMutation") == "<id>") {
    // mutation
} else {
    // original
}
```

Not all mutations can be inserted like this. Some mutations can only be inserted using ternary statements like this:
``` csharp
var localVariable = if(Environment.GetEnvironmentVariable("ActiveMutation") == "<id>") ? // mutation : //original
```

These kind of mutations should also be rollbacked different. My suggestion is an interface for both type of mutant placements. Something like:
``` csharp
public interface IMutantPlacement {
    SyntaxNode InsertMutant(SyntaxNode, Mutant);
    SyntaxNode RemoveMutant(SyntaxNode, Mutant);
}
```

It would have two implementations. One for if statements and one for ternary mutations.

richardwerkman:
With this new mutantplacer finally all code will be mutated. Until now `LocalDeclarationStatement` nodes were skipped from mutating. 

Closes #17


src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
```diff
@@ -103,14 +104,20 @@ private SyntaxTree RemoveMutantIfStatements(SyntaxTree originalTree, ICollection
...
+            // mark the broken mutation nodes to track
             var trackedTree = rollbackRoot.TrackNodes(brokenMutations);
             foreach (var brokenMutation in brokenMutations)
             {
-                // find the if statement in the new tree
+                // find the mutated node in the new tree
                 var nodeToRemove = trackedTree.GetCurrentNode(brokenMutation);
-                // remove the if statement and update the tree
-                trackedTree = trackedTree.ReplaceNode(nodeToRemove, nodeToRemove.Else.Statement);
+                // remove the mutated node using its MutantPlacer remove mehtod and update the tree
```
kevinlyles: Typo: mehtod
```suggestion
                // remove the mutated node using its MutantPlacer remove method and update the tree
```

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -60,21 +60,29 @@ public IEnumerable<Mutant> GetLatestMutantBatch()
...
+                SyntaxNode ast = currentNode;
+                foreach (var mutant in FindMutants(expression))
+                {
+                    _mutants.Add(mutant);
+                    var mutatedNode = ApplyMutant(expression, mutant);
+                    ast = ast.ReplaceNode(expression, MutantPlacer.PlaceWithConditionalExpression(expression, mutatedNode, mutant.Id));
+                }
+                return ast;
+            }
+            else if (currentNode is StatementSyntax && currentNode.Kind() != SyntaxKind.Block)
```
kevinlyles: (not caused by this PR)

Won't the BlockSyntax check prevent mutators like the [Block mutator](https://github.com/stryker-mutator/stryker-handbook/blob/master/mutator-types.md#block-statement) from working correctly?

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -60,21 +60,29 @@ public IEnumerable<Mutant> GetLatestMutantBatch()
...
             {
-                // This is a statement, meaning that we can insert mutants here
+                // The mutations should be placed using a IfStatement
```
kevinlyles: a/an
```suggestion
                // The mutations should be placed using an IfStatement
```

src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        public static ConditionalExpressionSyntax PlaceWithConditionalExpression(ExpressionSyntax original, ExpressionSyntax mutated, int mutantId)
+        {
+            // place the mutated statement inside the if statement
+            return SyntaxFactory.ConditionalExpression(
+                condition: GetBinaryExpression(mutantId),
+                whenTrue: mutated,
+                whenFalse: original)
+                // Mark this node as a MutationTernary node. Store the MutantId in the annotation to retrace the mutant later
```
kevinlyles:
```suggestion
                // Mark this node as a MutationConditional node. Store the MutantId in the annotation to retrace the mutant later
```

src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs
```diff
@@ -0,0 +1,71 @@
...
+                return ifStatement.Else.Statement;
+            else return null;
+        }
+
+        public static ConditionalExpressionSyntax PlaceWithConditionalExpression(ExpressionSyntax original, ExpressionSyntax mutated, int mutantId)
+        {
+            // place the mutated statement inside the if statement
+            return SyntaxFactory.ConditionalExpression(
+                condition: GetBinaryExpression(mutantId),
+                whenTrue: mutated,
```
kevinlyles: This and/or the conditional expression as a whole will run into [operator precedence](https://docs.microsoft.com/en-us/cpp/c-language/precedence-and-order-of-evaluation?view=vs-2017#precedence-and-associativity-of-c-operators) issues.  We should add parentheses to avoid this.

src/Stryker.Core/Stryker.Core.UnitTest/Compiling/RollbackProcessTests.cs
```diff
@@ -43,7 +43,7 @@ public string Subtract(string first, string second)
             var annotatedSyntaxTree = syntaxTree.GetRoot()
```
kevinlyles: RollbackProcessTests should have new tests for rolling back conditional statements

src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_SyntaxShouldBe_ConditionalStatement_IN.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace StrykerNet.UnitTest.Mutants.TestResources
+{
+    public class TestClass
+    {
+        void TestMethod()
+        {
+            int test = 10;
+            int testPlusTest = test + test;
+
+            bool Add(int x, int y) => x + y;
```
kevinlyles: We should add a similar test to the ShouldBe_IfStatement input (with a bracketed function body)

kevinlyles: Nitpick: this should return an int
```suggestion
            int Add(int x, int y) => x + y;
```

src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_SyntaxShouldBe_ConditionalStatement_IN.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
```
kevinlyles: We should test the unary ExpressionStatement cases like:
```x++;```
(in the IfStatement version of this file)

File: src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs (hunk 2 above)
Comment (richardwerkman) [reply src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs]: For now skipping block statements makes this work, else the mutation process crashes. 

Also for the block mutator: I guess the Block Mutator will be a special kind. It will produce a compile error outside of the mutationIf when the block removes a return statement. The framework will not be able to rollback such an error. This is why I chose to not create an issue for that mutator yet. I'll create an issue for it now. Maybe someone knows how to fix that problem!

File: src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs (hunk 5 above)
Comment (richardwerkman) [reply src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs]: What do you mean by that? For me no issues occured without the parentheses. In what case would this be better?

File: src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs (hunk 2 above)
Comment (kevinlyles) [reply src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs]: Ah, I hadn't considered that.  An issue makes sense.

File: src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs (hunk 5 above)
Comment (kevinlyles) [reply src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs]: After looking into this further, I'm not sure I can actually come up with an example that breaks.  I think I had the table reversed in my head.

I think that, because we're manipulating the syntax tree, we get some protection from this for free.

**Comment by richardwerkman** [comment]
@kevinlyles I've tested your case with the:
```
int b = a += 1 + 2;
```
And it successfully mutates `1 + 2` into `1 - 2`. So I don't think we have a problem here :)

File: src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_SyntaxShouldBe_ConditionalStatement_IN.cs (hunk 8 above)
Comment (richardwerkman) [reply src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_SyntaxShouldBe_ConditionalStatement_IN.cs]: I don't think this test should call additional mutators. It should only test the placing of the mutations, not other functionality.

**Comment by kevinlyles** [comment]
> @kevinlyles I've tested your case with the:
> 
> ```
> int b = a += 1 + 2;
> ```
> 
> And it successfully mutates `1 + 2` into `1 - 2`. So I don't think we have a problem here :)

I'll need to look at it again, but I'm somehow locked out of my VS account.  I'll see if I can get that fixed tomorrow and try it again.

File: src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs [hunk 9]
```diff
@@ -0,0 +1,71 @@
...
+                whenFalse: original)
+                // Mark this node as a MutationConditional node. Store the MutantId in the annotation to retrace the mutant later
+                .WithAdditionalAnnotations(new SyntaxAnnotation("MutationConditional", mutantId.ToString()));
+        }
+
+        public static SyntaxNode RemoveByConditionalExpression(SyntaxNode node)
+        {
+            if (node is ConditionalExpressionSyntax conditional)
```
Mobrockers: This if else is missing parens

src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs
```diff
@@ -0,0 +1,71 @@
...
+                condition: GetBinaryExpression(mutantId),
+                statement: SyntaxFactory.Block(mutated),
+                @else: SyntaxFactory.ElseClause(SyntaxFactory.Block(original)))
+                // Mark this node as a MutationIf node. Store the MutantId in the annotation to retrace the mutant later
+                .WithAdditionalAnnotations(new SyntaxAnnotation("MutationIf", mutantId.ToString()));
+        }
+
+        public static SyntaxNode RemoveByIfStatement(SyntaxNode node)
+        {
+            if (node is IfStatementSyntax ifStatement)
```
Mobrockers: this if else needs parens

src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        public static SyntaxNode RemoveByConditionalExpression(SyntaxNode node)
+        {
+            if (node is ConditionalExpressionSyntax conditional)
+                return conditional.WhenFalse;
+            else return null;
+        }
+
+        private static ExpressionSyntax GetBinaryExpression(int mutantId)
```
Mobrockers: This function could do with some documentation. Maybe use named parameters at the very least. It would also help if the objects were constructed separately before combining and returning. Right now this is 20 lines of inline object construction.

kevinlyles:
I'm still seeing a missing mutation.  I put together a failing test that may illustrate the issue better at https://github.com/kevinlyles/stryker-net/tree/ternary, which is up-to-date with your branch as of a few minutes ago.

Note that the test is not ready for merging into the main branch -- it messes with the mutators used in the test and makes some other changes to the test outputs as a result.  I'm just trying to show the problem better.

richardwerkman:
@kevinlyles I've checked your test case and indeed it failed. But curiously enough this exact case works for me when I run stryker on a test project containing your test case. 

``` csharp
int a = 1;
int b = a += 1 + 2;
b = a *= 2 - 3;
```
The output:
```
[Survived] AddAssignmentExpression to SubtractAssignmentExpression mutation on line 19: 'a += 1 + 2' ==> 'a -=1 + 2'
[Survived] Binary expression mutation on line 19: '1 + 2' ==> '1 - 2'
[Survived] MultiplyAssignmentExpression to DivideAssignmentExpression mutation on line 20: 'a *= 2 - 3' ==> 'a /=2 - 3'
[Survived] Binary expression mutation on line 20: '2 - 3' ==> '2 + 3'
```
I guess there must be some other factor in the unit test that makes the test fail. During an actual mutation testrun I've seen all mutations made.

kevinlyles:
The mutant surviving would also happen if the mutant is recorded but not actually emitted, right?  I think that's what is actually happening.

This code:
```
namespace TestMultipleMutations
{
	public class Program
	{
		static void Main(string[] args)
		{
			System.Console.WriteLine(TestMethod());
			System.Console.ReadKey();
		}

		public static int TestMethod()
		{
			int a = 1;
			int b = a += 1 + 2;
			b = a *= 2 - 4;
			return b;
		}
	}
}
```
With this test:
```
using Microsoft.VisualStudio.TestTools.UnitTesting;

namespace TestMultipleMutations.UnitTests
{
	[TestClass]
	public class UnitTests
	{
		[TestMethod]
		public void TestMethod()
		{
			Assert.AreEqual(-8, Program.TestMethod());
		}
	}
}
```

Has a "surviving" mutation:
```
- \TestMultipleMutations [3/4 (75.00 %)]
--- Program.cs [3/4 (75.00 %)]
[Killed] AddAssignmentExpression to SubtractAssignmentExpression mutation on lin
e 16: 'a += 1 + 2' ==> 'a -=1 + 2'
[Survived] Binary expression mutation on line 16: '1 + 2' ==> '1 - 2'
[Killed] MultiplyAssignmentExpression to DivideAssignmentExpression mutation on
line 17: 'a *= 2 - 4' ==> 'a /=2 - 4'
[Killed] Binary expression mutation on line 17: '2 - 4' ==> '2 + 4'
```

However, that mutation is caught by the test if I make it by hand.  I think this is a real bug.

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
```diff
@@ -103,14 +104,20 @@ private SyntaxTree RemoveMutantIfStatements(SyntaxTree originalTree, ICollection
...
             var trackedTree = rollbackRoot.TrackNodes(brokenMutations);
             foreach (var brokenMutation in brokenMutations)
             {
-                // find the if statement in the new tree
+                // find the mutated node in the new tree
                 var nodeToRemove = trackedTree.GetCurrentNode(brokenMutation);
-                // remove the if statement and update the tree
-                trackedTree = trackedTree.ReplaceNode(nodeToRemove, nodeToRemove.Else.Statement);
+                // remove the mutated node using its MutantPlacer remove method and update the tree
+                if (nodeToRemove is IfStatementSyntax ifStatement)
```
Mobrockers: the ifStatement assignment is never used and seems unnecessary to me

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -60,21 +60,30 @@ public IEnumerable<Mutant> GetLatestMutantBatch()
         /// <returns>Mutated node</returns>
         public SyntaxNode Mutate(SyntaxNode currentNode)
         {
-            if (currentNode is StatementSyntax && currentNode.Kind() != SyntaxKind.Block)
+            if (GetExpressionSyntax(currentNode) is var expressionSyntax && expressionSyntax != null) {
```
Mobrockers: the is var pattern matching is not currently being used here. I think this whole if/elseif can be tidied up as the following:

```csharp
if (GetExpressionSyntax(currentNode) is var expressionSyntax && expressionSyntax != null)
{
    // The mutations should be placed using a ConditionalExpression
    ExpressionSyntax expressionAst = expressionSyntax;
    foreach (var mutant in FindMutants(expressionSyntax))
    {
        _mutants.Add(mutant);
        ExpressionSyntax mutatedNode = ApplyMutant(expressionSyntax, mutant);
        expressionAst = MutantPlacer.PlaceWithConditionalExpression(expressionAst, mutatedNode, mutant.Id);
    }
    return currentNode.ReplaceNode(expressionSyntax, expressionAst);
}
else if (currentNode is StatementSyntax ast && currentNode.Kind() != SyntaxKind.Block)
{
    // The mutations should be placed using an IfStatement
    foreach (var mutant in currentNode.ChildNodes().SelectMany(FindMutants))
    {
        _mutants.Add(mutant);
        StatementSyntax mutatedNode = ApplyMutant(currentNode as StatementSyntax, mutant);
        ast = MutantPlacer.PlaceWithIfStatement(ast, mutatedNode, mutant.Id);
    }
    return ast;
}
```

kevinlyles:
The issue with missing mutations does appear to be fixed.

However, I'm no longer able to run Stryker on itself.  I get the following error when running it on Stryker.Core.UnitTests.csproj:
```
[11:10:28 ERR] An error occurred during the mutation test run
System.NullReferenceException: Object reference not set to an instance of an obj
ect.
   at Stryker.Core.Mutants.MutantOrchestrator.GetExpressionSyntax(SyntaxNode nod
e) in C:\Users\Me\Desktop\Stryker\src\Stryker.Core\Stryker.Core\Mutants\MutantOr
chestrator.cs:line 155
   at Stryker.Core.Mutants.MutantOrchestrator.Mutate(SyntaxNode currentNode) in
C:\Users\Me\Desktop\Stryker\src\Stryker.Core\Stryker.Core\Mutants\MutantOrchestr
ator.cs:line 63
   at System.Linq.Enumerable.SelectEnumerableIterator`2.ToList()
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Stryker.Core.Mutants.MutantOrchestrator.Mutate(SyntaxNode currentNode) in
C:\Users\Me\Desktop\Stryker\src\Stryker.Core\Stryker.Core\Mutants\MutantOrchestr
ator.cs:line 94
   at System.Linq.Enumerable.SelectEnumerableIterator`2.ToList()
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Stryker.Core.Mutants.MutantOrchestrator.Mutate(SyntaxNode currentNode) in
C:\Users\Me\Desktop\Stryker\src\Stryker.Core\Stryker.Core\Mutants\MutantOrchestr
ator.cs:line 94
   at System.Linq.Enumerable.SelectEnumerableIterator`2.ToList()
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Stryker.Core.Mutants.MutantOrchestrator.Mutate(SyntaxNode currentNode) in
C:\Users\Me\Desktop\Stryker\src\Stryker.Core\Stryker.Core\Mutants\MutantOrchestr
ator.cs:line 94
   at System.Linq.Enumerable.SelectEnumerableIterator`2.ToList()
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Stryker.Core.Mutants.MutantOrchestrator.Mutate(SyntaxNode currentNode) in
C:\Users\Me\Desktop\Stryker\src\Stryker.Core\Stryker.Core\Mutants\MutantOrchestr
ator.cs:line 94
   at System.Linq.Enumerable.SelectEnumerableIterator`2.ToList()
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Stryker.Core.Mutants.MutantOrchestrator.Mutate(SyntaxNode currentNode) in
C:\Users\Me\Desktop\Stryker\src\Stryker.Core\Stryker.Core\Mutants\MutantOrchestr
ator.cs:line 94
   at Stryker.Core.MutationTest.MutationTestProcess.Mutate() in C:\Users\Me\Desk
top\Stryker\src\Stryker.Core\Stryker.Core\MutationTest\MutationTestProcess.cs:li
ne 63
   at Stryker.Core.StrykerRunner.RunMutationTest(StrykerOptions options) in C:\U
sers\Me\Desktop\Stryker\src\Stryker.Core\Stryker.Core\StrykerRunner.cs:line 74
[11:10:28 INF] Time Elapsed 00:00:11.8893802
```

richardwerkman:
I've located that issue as well. It appeared to be caused by:

```csharp
string strout;
```
Stryker wants to mutate the value of the initializer of this statement, but there is no initializer. This caused the null reference exception.

kevinlyles:
That fix looks good, but running it on Stryker.CLI.UnitTests.csproj still gives me a (different) error:
```
[11:10:28 ERR] An error occurred during the mutation test run
System.NullReferenceException: Object reference not set to an instance of an obj
ect.
   at Stryker.Core.Mutants.MutantOrchestrator.GetExpressionSyntax(SyntaxNode nod
e) in C:\Users\Me\Desktop\Stryker\src\Stryker.Core\Stryker.Core\Mutants\MutantOr
chestrator.cs:line 155
   at Stryker.Core.Mutants.MutantOrchestrator.Mutate(SyntaxNode currentNode) in
C:\Users\Me\Desktop\Stryker\src\Stryker.Core\Stryker.Core\Mutants\MutantOrchestr
ator.cs:line 63
   at System.Linq.Enumerable.SelectEnumerableIterator`2.ToList()
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Stryker.Core.Mutants.MutantOrchestrator.Mutate(SyntaxNode currentNode) in
C:\Users\Me\Desktop\Stryker\src\Stryker.Core\Stryker.Core\Mutants\MutantOrchestr
ator.cs:line 94
   at System.Linq.Enumerable.SelectEnumerableIterator`2.ToList()
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Stryker.Core.Mutants.MutantOrchestrator.Mutate(SyntaxNode currentNode) in
C:\Users\Me\Desktop\Stryker\src\Stryker.Core\Stryker.Core\Mutants\MutantOrchestr
ator.cs:line 94
   at System.Linq.Enumerable.SelectEnumerableIterator`2.ToList()
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Stryker.Core.Mutants.MutantOrchestrator.Mutate(SyntaxNode currentNode) in
C:\Users\Me\Desktop\Stryker\src\Stryker.Core\Stryker.Core\Mutants\MutantOrchestr
ator.cs:line 94
   at System.Linq.Enumerable.SelectEnumerableIterator`2.ToList()
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Stryker.Core.Mutants.MutantOrchestrator.Mutate(SyntaxNode currentNode) in
C:\Users\Me\Desktop\Stryker\src\Stryker.Core\Stryker.Core\Mutants\MutantOrchestr
ator.cs:line 94
   at System.Linq.Enumerable.SelectEnumerableIterator`2.ToList()
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Stryker.Core.Mutants.MutantOrchestrator.Mutate(SyntaxNode currentNode) in
C:\Users\Me\Desktop\Stryker\src\Stryker.Core\Stryker.Core\Mutants\MutantOrchestr
ator.cs:line 94
   at Stryker.Core.MutationTest.MutationTestProcess.Mutate() in C:\Users\Me\Desk
top\Stryker\src\Stryker.Core\Stryker.Core\MutationTest\MutationTestProcess.cs:li
ne 63
   at Stryker.Core.StrykerRunner.RunMutationTest(StrykerOptions options) in C:\U
sers\Me\Desktop\Stryker\src\Stryker.Core\Stryker.Core\StrykerRunner.cs:line 74
[11:10:28 INF] Time Elapsed 00:00:11.8893802
```

It looks like it's breaking when mutating CLIOptions.cs, specifically when trying to replace:
```
ConfigFilePath = new CLIOption<string> {
    ArgumentName = "--config-file-path",
    ArgumentShortName = "-cp <path>",
    ArgumentDescription = "Sets the config-file-path relative to current workingDirectory | stryker-config.json (default)",
    DefaultValue = "stryker-config.json"
};
```
with:
```
ConfigFilePath = new CLIOption<string> {
    ArgumentName = System.Environment.GetEnvironmentVariable("ActiveMutation")=="0"?"":"--config-file-path",
    ArgumentShortName = System.Environment.GetEnvironmentVariable("ActiveMutation")=="1"?"":"-cp <path>",
    ArgumentDescription = System.Environment.GetEnvironmentVariable("ActiveMutation")=="2"?"":"Sets the config-file-path relative to current workingDirectory | stryker-config.json (default)",
    DefaultValue = System.Environment.GetEnvironmentVariable("ActiveMutation")=="3"?"":"stryker-config.json"
};
```
while processing the VariableDeclarationSyntax node containing that.  I tried adding those changes manually, and the syntax is valid, so I'm not sure what's wrong.

richardwerkman:
I'll look at that later today 👍

richardwerkman:
@kevinlyles I'm not sure what makes that case crash... The syntax seems fine indeed but the SyntaxEditor still breaks. Maybe this is a bug inside the Roslyn class? 

I've updated the code to not use the SyntaxEditor anymore. This fixed the issue on my machine.

I was quite shocked by the way when I runned Stryker on Stryker.CLI.UnitTests.... Almost only false positives. We should definitely work on that 😞

kevinlyles:
I'll take a look at this probably tomorrow. Sorry for the delay!

kevinlyles:
That seems to have fixed it.  I think all of my issues have been addressed.

The long list of false positives is interesting -- I wonder if something might be wrong with the string mutator.

richardwerkman:
@kevinlyles I've found the reason for the false positives. There was a fault in a unit test that depended on a file.Move(). Once a test failed, the target file was not removed and all testruns after that would fail because of this file.

PR #247, opened 2018-11-18
Improve mutants performance

### LINKED ISSUE #238: Reduce number of times "ActiveMutation" envionment variable is called
The current logic used to activate the mutant to be tested is slow, as it relies on accessing environment variables.
I prototyped a version using a cached value, thanks to a constant, and it provides significant performance improvements:
Stryker generates around 1600 mutants for NFluent projects, and a test session is around 10 seconds.

Total run time using environment variable: ~4 hours (4:07)
Total run time using a constant: 3 hours (3:09)

Your mileage may vary, but 25% time decrease is interesting.

dupdob:
Replace the activemutant selection logic by injecting an helper function that caches the current identifier in a constant. Ideally, the JIT simply throws away all mutants except the active one.

Performant tests on a medium sized project (NFluent) exhibits a gain of 25%-30% run time, 2:50 down from 4:14

Closes #238


src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs
```diff
@@ -7,6 +7,22 @@ namespace Stryker.Core.Mutants
 {
     public static class MutantPlacer
     {
+        private const string helper = @"
+namespace Stryker
```
richardwerkman: Maybe it would be nice to actually include this as a C# file. That way we can test this code as well and have syntax checking. We can just read the file and parse it here.

dupdob: Done. Note that I had to change the naming of the helper class to prevent ambiguity with System.Environment.

dupdob:
Basically, this change should remove any performance impact due to Stryker. So mileage may vary, from no visible to change to dramatic improvement depending on where mutants where injected.

Mobrockers:
@dupdob Sorry for the wait. I have fixed the integration test + build and manually triggered a build of this pr, the integration test seems to not create any mutants with this change.

dupdob:
Thanks for the feedback. I guess there must some syntax error that is not caught by the unit tests.
I'll look into this right now

dupdob:
Yep, got it: discrepancy between helper function name as defined (ActiveMutant) and expected by selectors (ActiveMutation).
Just pushed the fix.

src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs
```diff
@@ -1,12 +1,28 @@
...
 
 namespace Stryker.Core.Mutants
 {
     public static class MutantPlacer
     {
+        private static readonly string helper;
+
+        static MutantPlacer()
+        {
+            using (var stream =
```
richardwerkman: Why not use:

``` csharp
string helper = File.ReadAllLines("Stryker.Core.Mutants.ActiveMutantHelper.cs");
```

?

src/Stryker.Core/Stryker.Core/Mutants/ActiveMutantHelper.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Stryker
+{
+    public static class Helper
+    {
+        private static readonly int activeMutant;
```
richardwerkman: Shouldn't this be `int?`, since the int default value is 0. When no EnvironmentVariable is set, Mutant 0 would always be active. Or we could start the mutant ids at 1?

src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs
dupdob: The file is embedded as an embedded resource. I don't how to use ReadAllLines for those. 
I am open to alternative strategies, if any.

src/Stryker.Core/Stryker.Core/Mutants/ActiveMutantHelper.cs
dupdob: Good point. I assumed that we always wanted to have an active mutant. I just added a commit that default to -1 instead of 0.

src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs
richardwerkman: We can create a static helper for this in the testinghelpers dir? That way this part of our code looks cleaner and makes reuse for embedded files possible. But the current way is also ok for me.

src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs
```diff
@@ -1,12 +1,28 @@
...
+                typeof(MutantPlacer).Assembly.GetManifestResourceStream("Stryker.Core.Mutants.ActiveMutantHelper.cs"))
```
Mobrockers: Would something similar to this:

```
Type activeMutantHelper = typeof(ActiveMutantHelper);
using (var stream = activeMutantHelper.Assembly.GetManifestResourceStream(activeMutantHelper.FullName))
```

perhaps be possible so we do not need hardcoded resource file names? I'm pretty sure this example is not complete but perhaps there is a way to use this to format the complete name.

src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs
dupdob: Can you share a link to the corresponding github folder? I am afraid I am not able to find this dir.

src/Stryker.Core/Stryker.Core/Mutants/ActiveMutantHelper.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Stryker
+{
+    public static class Helper
```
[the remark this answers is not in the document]
dupdob: Done. Final (class and file) name is ActiveMutationHelper.

src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs
dupdob: Resource file name are hardcoded identifier as those are external resources (from a source code point of view). 
I am not able to implement what you propose as the source code is not built as part of the project, only embedded. Hence there is no type to refer to.

src/Stryker.Core/Stryker.Core/Mutants/ActiveMutantHelper.cs
```diff
@@ -0,0 +1,12 @@
...
+    {
+        private static readonly int activeMutant;
+        static Helper()
+        {
+            activeMutant = int.Parse(System.Environment.GetEnvironmentVariable("ActiveMutation") ?? "-1");
+        }
+        public static int ActiveMutation => activeMutant;
```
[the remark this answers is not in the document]
dupdob: Indeed. It is a matter of code style preference, but it's your project==> Done.

src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs
Mobrockers: Fair enough

src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs
richardwerkman: This is also ok for now, when this functionallity is needed again for a new feature we'll move it to the helpers dir.

PR #249, opened 2018-11-22
Secure against sleep induced timeouts

### LINKED ISSUE #240: Many undue timeout results
Running Stryker against NFluent gives a variable number of timeouts between 12 and 15. The first three timeouts could not be explained by the mutant: running them in isolation took the usual test time (~10 seconds).
Plus, successive runs of Stryker give timeouts for different mutants.
Everything hints there is something fishy with those timeouts. I am currently doing further testing to try and understand the actual cause for those.

dupdob:
Prevents timeouts due to computer going to sleep.
Add estimated remaining time to ProgressBar Reporter
Fix #240


src/Stryker.Core/Stryker.Core/Reporters/Progress/ProgressBarReporter.cs
```diff
@@ -31,13 +32,15 @@ public ProgressBarReporter(IConsoleOneLineLogger testsProgressLogger)
 
         public void ReportInitialState(int totalNumberOfTests)
         {
+            _startTime = DateTime.UtcNow;;
```
richardwerkman: two ;;
```suggestion
            _startTime = DateTime.UtcNow;
```

src/Stryker.Core/Stryker.Core/Reporters/Progress/ProgressBarReporter.cs
```diff
@@ -17,12 +17,13 @@ public class ProgressBarReporter : IProgressBarReporter
         private const char ProgressBarDoneToken = '\u2588';
         private const char ProgressBarLeftToken = '-';
         private const int ProgressBarInitialState = -1;
-        private const string LoggingFormat = "Tests progress | {0} | {1} / {2} | {3} % |";
+        private const string LoggingFormat = "Tests progress | {0} | {1} / {2} | {3} % | rem time: {4} |";
```
richardwerkman: I prefer the original layout like in #9. It indicates that it is only a rough estimation.

It looked like: ` ~2h 17m 40s` but `~02:17:40` is also good for me.

src/Stryker.Core/Stryker.Core/Reporters/Progress/ProgressBarReporter.cs
```diff
@@ -74,7 +79,36 @@ public void ReportRunTest()
...
+
+        private string RemainingTime()
+        {
+            if (_totalNumberOfTests == 0 || _numberOfTestsRun == 0)
+            {
+                return "NA";
+            }
+
+            var elapsed = (DateTime.UtcNow - _startTime).TotalSeconds;
+            var remaining = (_totalNumberOfTests - _numberOfTestsRun) * elapsed / _numberOfTestsRun;
```
richardwerkman: The calculation is pretty accurate but could be better. We could add a little extra istimated time for future timeouts? At the start of my testrun it estimated too low (13 minutes instead of 18 actual), and started to get better later when some timeouts occured. Maybe you could experiment with this?

src/Stryker.Core/Stryker.Core/Reporters/Progress/ProgressBarReporter.cs
dupdob: Yep

src/Stryker.Core/Stryker.Core/Reporters/Progress/ProgressBarReporter.cs
dupdob: I like the ~, definitely. Also migrated to proposed time format, with a small adjustment: only display the two most significatives unit (i.e: minutes & seconds, hours & minutes or days & hours)

src/Stryker.Core/Stryker.Core/Reporters/Progress/ProgressBarReporter.cs
dupdob: Current calculation is naive and basic. I can devise a more sophisticated model, but tuning will be tricky as timeout percentage may significantly vary. If you have only a few mutants, timeouts will significantly impact the estimated remaining time.
Alternatively, if you have a lot of mutants, estimate should be pretty good when you reach 10% of testing. And as this feature is mostly useful for larger projects, I propose to keep this algorithm as is and open an issue for a more elaborated one, where the model can be discussed

src/Stryker.Core/Stryker.Core/Reporters/Progress/ProgressBarReporter.cs
dupdob: Also, thanks for the pointer to #9. I was surprised to see there was no estimated time. Why was the feature not implemented/dropped?

src/Stryker.Core/Stryker.Core/Reporters/Progress/ProgressBarReporter.cs
richardwerkman: True, it doesn't have to be perfect so this is a very good start. I was thinking just now: by using a timestamp as the starttime the estimation would be way off if the computer would go to sleep for several hours and continued later. Is there a way we could exclude the sleep time?

src/Stryker.Core/Stryker.Core/Reporters/Progress/ProgressBarReporter.cs
richardwerkman: It was dropped from that pullrequest as it was very inaccurate. And we thougt about creating an issue for this later. It seems that we forgot to add it.

src/Stryker.Core/Stryker.Core/Reporters/Progress/ProgressBarReporter.cs
dupdob: Good point. I could use a strategy similar to the one used to monitor timeouts, but this would probably badly impair precision. But I have an alternate strategy I need to try first.

dupdob: I switched to Stopwatch (instead DateTime). Less accurate, but will compensate for sleep phase. But it depends on the quality of implementation and IIRC it may be incorrect for some AMD CPUs when they enter low power modes

dupdob:
I have no work pending on this PR, unless a new review raises concern

src/Stryker.Core/Stryker.Core/Reporters/Progress/ProgressBarReporter.cs
```diff
@@ -17,12 +18,13 @@ public class ProgressBarReporter : IProgressBarReporter
...
         private const char ProgressBarLeftToken = '-';
         private const int ProgressBarInitialState = -1;
-        private const string LoggingFormat = "Tests progress | {0} | {1} / {2} | {3} % |";
+        private const string LoggingFormat = "Tests progress | {0} | {1} / {2} | {3} % | {4} |";
 
         private readonly IConsoleOneLineLogger _testsProgressLogger;
 
         private int _totalNumberOfTests;
         private int _numberOfTestsRun;
+        private Stopwatch _startTime;
```
richardwerkman: I don't think the stopwatch fixes the estimate after a long sleep. It can be stopped however when the PC goes to sleep and resumed again on wake.  This could be achieved by using SystemEvents (windows only). In my opinion we don't want to use that for such a specific case. I guess it's ok for now that the estimate is a bit off after a sleep period. At least your fix makes sure the output (report) is correct.

dupdob: Well, empiric tests showed it works on my laptop. I suggest we see what happens in the large as we don't have alternative implementations.

PR #251, opened 2018-11-24
Test myget feed as pre-release feed

Mobrockers:


richardwerkman:
@Mobrockers Maybe it would be a nice test to create this PR from a fork, to see if the fix works :)

Mobrockers:
Good point, I'll do that soon. Shouldn't be any issues though as the
secrets are plain text in the pipeline file :d

Mobrockers:
Closing as PR will come from fork to test

PR #253, opened 2018-11-25
feat(Excluded mutations): Add option to exlude one or more mutators

### LINKED ISSUE #204: Disable mutators
Since we now have a wide range of mutators available there are cases that some mutators are not wanted on certain projects. For example: a project with a lot of trace logging like:

``` csharp
Log.Trace("Usercontroller called");
``` 

I could want to disable the string mutator. 

``` json
{
    "stryker-config":
    {
        "disable-mutators": [
            "StringMutator",
            "CheckedMutator",
         ],
    }
}
```

Users should be able to pass a list of mutators to disable. This can be a tricky thing since the names of the mutators are not mapped 1:1 in the documentation. For example the string mutator is documented as one mutator, but in code there are 2 instances. We could introduce mutator categories/groups for this?

richardwerkman:
Closes #204 

Also updates the names of a lot of mutators. Users can pass the name of a mutator as in our documentation. The names are based on the discussion [here](https://github.com/stryker-mutator/stryker-handbook/issues/11).

During testing I found the name of `--log-file` to be `--log-level-file` in code. I fixed this to match the documentation.


src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
```diff
@@ -46,6 +46,14 @@ public static class CLIOptions
             JsonKey = "timeout-ms"
         };
 
+        public static readonly CLIOption<string[]> ExcludedMutations = new CLIOption<string[]>
+        {
+            ArgumentName = "--excluded-mutations",
+            ArgumentShortName = "-em <mutator-name>",
+            ArgumentDescription = "The given mutators will be excluded for this mutation testrun.",
```
Mobrockers: Imo the description for this feature should give some hint as to where you can find the list of available mutators. Preferably we should add a commandline argument that doesn't run stryker but simply lists the mutators available in this version of stryker and a link to github. For now I think it would be enough to add the link to the readme in the description.

richardwerkman: We have the MutatorType enum, we could just print it here. I don't know if an extra command would be what we want.

Mobrockers: I suppose that could work as well. I was thinking the list-mutators command
would include a detailed description and then the help text for excluded
mutations could mention using the list-mutators command to find out which
mutations to exclude.

richardwerkman: I've thought about this and for now I don't think this is needed for now. We can always add this as a new feature request later.

Issue #104, opened 2018-09-01
Use stryker on the stryker build


yaelkeemink:
It would be cool if we use stryker-net on stryker-net itself in the build. That way we can check our own score.

richardwerkman:
This would be extra useful now we have the dashboard reporter. We can show our mutation score badge in the readme and update our report every day: https://dashboard.stryker-mutator.io/reports/github.com/stryker-mutator/stryker-net/master

I suggest running on master every night is sufficient for now.

PR #274, opened 2018-12-05
chore(Easier testing setup): Add default values to StrykerOptions constructor

joostas:


src/Stryker.Core/Stryker.Core.UnitTest/Reporters/ConsoleReportReporterTests.cs
```diff
@@ -26,7 +26,7 @@ public void ConsoleReportReporter_ShouldPrintOnReportDone()
             chalkMock.Setup(x => x.DarkGray(It.IsAny<string>())).Callback((string text) => { output += text; });
             chalkMock.Setup(x => x.Default(It.IsAny<string>())).Callback((string text) => { output += text; });
 
-            var target = new ConsoleReportReporter(new StrykerOptions("", "ReportOnly", "", 1000, null, "debug", false, 1, 80, 60, 0), chalkMock.Object);
+            var target = new ConsoleReportReporter(new StrykerOptions(), chalkMock.Object);
```
ptoonen: This would now get a different loglevel, not sure if it matters though?

src/Stryker.Core/Stryker.Core.UnitTest/Reporters/ConsoleReportReporterTests.cs
```diff
@@ -61,7 +61,7 @@ public void ConsoleReportReporter_ShouldPrintKilledMutation()
                 Type = MutatorType.Arithmetic
             };
 
-            var target = new ConsoleReportReporter(new StrykerOptions("", "ReportOnly", "", 1000, null, "debug", false, 1, 80, 60, 0), chalkMock.Object);
+            var target = new ConsoleReportReporter(new StrykerOptions(), chalkMock.Object);
```
ptoonen: Different loglevel due to default values.

src/Stryker.Core/Stryker.Core.UnitTest/Reporters/ConsoleReportReporterTests.cs
```diff
@@ -103,7 +103,7 @@ public void ConsoleReportReporter_ShouldPrintSurvivedMutation()
                 Type = MutatorType.Arithmetic
             };
             
-            var target = new ConsoleReportReporter(new StrykerOptions("", "ReportOnly", "", 1000, null, "debug", false, 1, 80, 60, 0), chalkMock.Object);
+            var target = new ConsoleReportReporter(new StrykerOptions(), chalkMock.Object);
```
ptoonen: Different loglevel due to default values.

src/Stryker.Core/Stryker.Core.UnitTest/Reporters/ConsoleReportReporterTests.cs
```diff
@@ -147,7 +147,7 @@ public void ConsoleReportReporter_ShouldPrintRedOn60PercentAndLower()
                 Type = MutatorType.Arithmetic
             };
 
-            var target = new ConsoleReportReporter(new StrykerOptions("", "ReportOnly", "", 1000, null, "info", false, 1, 80, 60, 0), chalkMock.Object);
+            var target = new ConsoleReportReporter(new StrykerOptions(thresholdLow:60, thresholdBreak:0), chalkMock.Object);
```
ptoonen: Different loglevel due to default values.

src/Stryker.Core/Stryker.Core.UnitTest/Reporters/ConsoleReportReporterTests.cs
```diff
@@ -189,7 +189,7 @@ public void ConsoleReportReporter_ShouldPrintYellowOn80PercentAndLower()
                 Type = MutatorType.Arithmetic
             };
 
-            var target = new ConsoleReportReporter(new StrykerOptions("", "ReportOnly", "", 1000, null, "debug", false, 1, 80, 60, 0), chalkMock.Object);
+            var target = new ConsoleReportReporter(new StrykerOptions(thresholdLow:75), chalkMock.Object);
```
ptoonen: Different loglevel due to default values. Different reporter too.

src/Stryker.Core/Stryker.Core.UnitTest/Reporters/ConsoleReportReporterTests.cs
```diff
@@ -230,7 +230,7 @@ public void ConsoleReportReporter_ShouldGreenOn80PercentAndHigher()
                 Type = MutatorType.Arithmetic
             };
 
-            var target = new ConsoleReportReporter(new StrykerOptions("", "ReportOnly", "", 1000, null, "debug", false, 1, 80, 60, 0), chalkMock.Object);
+            var target = new ConsoleReportReporter(new StrykerOptions(), chalkMock.Object);
```
ptoonen: Different loglevel due to default values. Different reporter too.

src/Stryker.Core/Stryker.Core.UnitTest/Reporters/ReporterFactoryTests.cs
```diff
@@ -12,7 +12,7 @@ public class ReporterFactoryTests
     {
         public static IEnumerable<object[]> getParameters()
         {
-            yield return new object[] {new StrykerOptions("", "Console", "", 1000, null, "debug", false, 1, 80, 60, 0), typeof(BroadcastReporter)};
+            yield return new object[] {new StrykerOptions(), typeof(BroadcastReporter)};
```
ptoonen: Different loglevel due to default values.

src/Stryker.Core/Stryker.Core.UnitTest/StrykerOptionsTests.cs
```diff
@@ -19,7 +19,7 @@ public class StrykerOptionsTests
         [InlineData("trace", LogEventLevel.Verbose)]
         public void Constructor_WithCorrectLoglevelArgument_ShouldAssignCorrectLogLevel(string argValue, LogEventLevel expectedLogLevel)
         {
-            var options = new StrykerOptions("c:/test", "Console", "", 0, null, argValue, false, 1, 80, 60, 0);
+            var options = new StrykerOptions(logLevel:argValue);
```
ptoonen: Basepath change, as above, not sure if it matters but if it wasn't an intentional change..

src/Stryker.Core/Stryker.Core.UnitTest/StrykerOptionsTests.cs
```diff
@@ -32,7 +32,7 @@ public void Constructor_WithIncorrectLoglevelArgument_ShouldThrowStrykerInputExc
 
             var ex = Assert.Throws<StrykerInputException>(() =>
             {
-                new StrykerOptions("c:/test", "Console", "", 0, null, logLevel, false, 1, 80, 60, 0);
+                new StrykerOptions(logLevel: logLevel);
```
ptoonen: Different basepath due to default values.

src/Stryker.Core/Stryker.Core.UnitTest/StrykerOptionsTests.cs
```diff
@@ -45,7 +45,7 @@ public void Constructor_WithIncorrectSettings_ShoudThrowWithDetails()
...
```
ptoonen: Different basepath due to default values.

src/Stryker.Core/Stryker.Core.UnitTest/StrykerTests.cs
```diff
@@ -43,7 +43,7 @@ public void Stryker_ShouldInvokeAllProcesses()
 
             var target = new StrykerRunner(initialisationMock.Object, mutationTestProcessMock.Object);
 
-            var options = new StrykerOptions("c:/test", "Console", "", 2000, null, null, false, 1, 80, 60, 0);
+            var options = new StrykerOptions();
```
ptoonen: Different basepath due to default values. Different loglevel too.

src/Stryker.Core/Stryker.Core.UnitTest/Reporters/ReporterFactoryTests.cs
joostas: Yes. But does it meter here?

src/Stryker.Core/Stryker.Core.UnitTest/StrykerOptionsTests.cs
joostas: According Test name seems that only log level meters here.

src/Stryker.Core/Stryker.Core.UnitTest/StrykerOptionsTests.cs
joostas: Does it really meter here?

src/Stryker.Core/Stryker.Core.UnitTest/Reporters/ConsoleReportReporterTests.cs
joostas: Test passes. If this does mater than I think original test needs to be changed, because it failed mutation ;)

joostas:
Does it really require any changes? All modified tests passed.

src/Stryker.Core/Stryker.Core.UnitTest/Reporters/ConsoleReportReporterTests.cs
ptoonen: Hehe, check #104 for that. I'm not sure if it matters, maybe @richardwerkman cares to chime in?

ptoonen:
It's more of a 'make sure you have seen the possibly unintended changes' comment. Maybe @richardwerkman or @Mobrockers have an opinion on this?

src/Stryker.Core/Stryker.Core.UnitTest/Reporters/ConsoleReportReporterTests.cs
richardwerkman: This discussion is one of the reasons I created this issue. All these values in the test make it unclear what is being tested. 

In this case the different loglevel has nothing to do with this UnitTest. All this test is about is the output of the reporter, which doesn't use any values from the options. So the default values are ok.

src/Stryker.Core/Stryker.Core.UnitTest/Reporters/ConsoleReportReporterTests.cs
```diff
@@ -189,8 +189,7 @@ public void ConsoleReportReporter_ShouldPrintYellowOn80PercentAndLower()
                 Type = MutatorType.Arithmetic
             };
 
-            var target = new ConsoleReportReporter(new StrykerOptions("", "ReportOnly", "", 1000, null, "debug", false, false, 1, 80, 60, 0), chalkMock.Object);
-
+            var target = new ConsoleReportReporter(new StrykerOptions(thresholdLow:75), chalkMock.Object);
```
Mobrockers: thresholdLow should not be set to 75 here as the test name explicitly says the value should be 80. Since 80 is also the default value for threshold low, I suggest you do not pass any values to strykeroptions in this test and change the name to `ConsoleReportReporter_ShouldPrintYellowBetweenThresholdLowAndThresholdBreak`

src/Stryker.Core/Stryker.Core.UnitTest/Reporters/ConsoleReportReporterTests.cs
```diff
@@ -147,7 +147,7 @@ public void ConsoleReportReporter_ShouldPrintRedOn60PercentAndLower()
                 Type = MutatorType.Arithmetic
             };
 
-            var target = new ConsoleReportReporter(new StrykerOptions("", "ReportOnly", "", 1000, null, "info", false, false, 1, 80, 60, 0), chalkMock.Object);
+            var target = new ConsoleReportReporter(new StrykerOptions(thresholdLow:60, thresholdBreak:0), chalkMock.Object);
```
Mobrockers: I suggest not passing any values to strykeroptions here as 60 and 0 are the default values for thresholdLow and thresholdBreak. Instead change the test name to `ConsoleReportReporter_ShouldPrintRedUnderThresholdBreak`

src/Stryker.Core/Stryker.Core.UnitTest/Reporters/ConsoleReportReporterTests.cs
```diff
@@ -230,8 +229,7 @@ public void ConsoleReportReporter_ShouldGreenOn80PercentAndHigher()
                 Type = MutatorType.Arithmetic
             };
 
-            var target = new ConsoleReportReporter(new StrykerOptions("", "ReportOnly", "", 1000, null, "debug", false, false, 1, 80, 60, 0), chalkMock.Object);
-
+            var target = new ConsoleReportReporter(new StrykerOptions(), chalkMock.Object);
```
Mobrockers: Since no values are passed to StrykerOptions for this test, I suggest changing the test name to `ConsoleReportReporter_ShouldPrintGreenAboveThresholdHigh`

src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
```diff
@@ -24,7 +28,7 @@ public static class CLIOptions
             ArgumentName = "--log-console",
             ArgumentShortName = "-l <logLevel>",
             ArgumentDescription = "Sets the logging level | Options [error, warning, info (default), debug, trace]",
-            DefaultValue = "Info",
+            DefaultValue = "",
```
Mobrockers: The default value for log-console is not empty string

joostas: Yes. It's "info". But in code empty string value is handled as default  --log-console value
https://github.com/stryker-mutator/stryker-net/blob/f74f782fee9aca1ce3beb1e3bce05dadc4daa711/src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs#L97-L106

Having empty value here will let easier to change default logging optoin (if this will ever be needed).

On other hand using value from default `StrykerOptions` requires additional implementation, because string representation of `LogEventLevel.Information` is different from "info".
This can be solved either:
 * renaming `LogEventLevel.Information` to `LogEventLevel.Info` either
 * adding additional string parsing in `StrykerOptions.ValidateLogLevel`: `"information"` or
 * adding some attribute with string representation (seems to complicated for this case to me)

What is your preference? Or maybe there is some more elegant way?

Mobrockers: LogEventLevel is a serilog concept, we can't change this. I would prefer the option to just be hard coded string info in this case instead of empty string because at least that says something about the option.

Issue #275, opened 2018-12-05
Contribution Guide is outdated


ptoonen:
Currently the contribution guide doesn't say anything about the naming your commits. This potentially leads to commits which do not adhere to the [preferred commit message style](https://docs.google.com/document/d/1rk04jEuGfk9kYzfqCuOlPTSJw3hEDZJTBN5E5f1SALo).

Mobrockers:
Closed by #276

PR #276, opened 2018-12-05
docs(contributor): Update Contributors Guide

ptoonen:
Include the commit guidelines as proposed in #275

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,10 +9,15 @@ If you found a bug, please run ```dotnet stryker --log-console debug``` and add 
 New features are welcome! Either as requests or proposals.
 
 1.	Please create an issue first, so we know what to expect from you.
-2.	Create a fork on your github account.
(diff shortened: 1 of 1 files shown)


richardwerkman:
Well actually I removed it intentionally. Since contributors should not have to worry about the naming of commits. When merging to master we (maintainers) squash the commits into a new commit which should be conform the angular style guide. 

We could also change it to be a Maintainer specific part of the contribution guide?

Mobrockers:
If contributors happen to read the contributing guidelines they can use the recommended commit style and we won't have to change their message for them, I don't see any harm keeping it in there.

richardwerkman:
But when we select the squash option on a PR the name of the PR is selected as the name for the squash commit. So then we should put it down as a PR style guide?

ptoonen:
I was thinking about that too (calling it a PR guideline) - although I see this more as a guideline than as a strictly enforced rule. For now it's easy for maintainers to do this properly, however as the project grows larger, it would be nice if contributors help you there.

Issue #283, opened 2018-12-07
Stryker.Net do not report failure to build the mutated project post rollback step


dupdob:
If the assembly cannot be compiled after rollbacks have been processed, Stryker silently ignores the situation and process to test the mutants, resulting in 100% surviving rate.

This issue was made visible due to current limitation in safe mode.

PR #285, opened 2018-12-08
Fix(rollback): improve SafeMode

### LINKED ISSUE #279: Safe mode causes 0 mutationscore when active
When the new safe mode (#257) is activated it seems to cause a 0 mutationscore.

@dupdob Maybe you could take a look at this yourself?

dupdob:
Fix #279 and #283

### DIFF
--- a/src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
+++ b/src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
@@ -120,7 +120,13 @@ namespace Stryker.Core.MutationTest
                         _logger.LogInformation("{0} mutants got status {1}", mutantsToSkip.Count(), MutantStatus.Skipped.ToString());
                     }
                 }
+                else
+                {
+                    _logger.LogError("Failed to restore the project to a buildable state. Please report the issue. Stryker can not proceed further");
+                    throw new ApplicationException("Failed to restore build able state.");
+                }
             }
+
 
             _logger.LogInformation("{0} mutants ready for test", _input.ProjectInfo.ProjectContents.TotalMutants.Count());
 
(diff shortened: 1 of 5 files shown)


dupdob:
Working on the test issue

dupdob:
Well it turns out removing brackets is harder than I thought. Closing this PR and will keep on working

dupdob:
Fixed the issue by ensuring RemoveIfStatement returns a single statement when possible and a syntaxblock otherwise. The latter situation should not happen with current version of Stryker, but was present as a test.

Mobrockers:
Could you create an issue explaining the syntax kind you found which we do not yet properly support and attach the 'UnsupportedSyntax' test? That way you won't have to fix the missing syntax support in this PR and we don't lose the clear example you created either.

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
```diff
@@ -120,8 +120,14 @@ public void Mutate(StrykerOptions options)
                         _logger.LogInformation("{0} mutants got status {1}", mutantsToSkip.Count(), MutantStatus.Skipped.ToString());
                     }
                 }
+                else
+                {
+                    _logger.LogError("Failed to restore the project to a buildable state. Please report the issue. Stryker can not proceed further");
```
richardwerkman: Can this be moved to the CompilingProcess class?

dupdob: I guess I can, but this sounds strange to me: it means that compilingprocess.Compile either succeed or throws.
What benefit do you see there?

dupdob: No

PR #297, opened 2018-12-21
Add report json suggestions from #294

### LINKED ISSUE #294: JSON Report improvements
https://github.com/stryker-mutator/mutation-testing-elements/issues/1#issuecomment-446638027

@nicojs I'm not sure I understand the problem here. We provide the source code and we provide the mutation + span. Would it not make the most sense to include carriage returns? Otherwise, how will you match/replace the mutation on the original source code, which is going to include the carriage returns as well?

Mobrockers:
Closes #294


src/Stryker.Core/Stryker.Core/Reporters/JsonReporter.cs
```diff
@@ -1,82 +1,263 @@
...
+            var jsonReport = JsonReportComponent.FromProjectComponent(reportComponent, _options);
+            jsonReport.ThresholdHigh = _options.ThresholdOptions.ThresholdHigh;
+            jsonReport.ThresholdLow = _options.ThresholdOptions.ThresholdLow;
+            jsonReport.ThresholdBreak = _options.ThresholdOptions.ThresholdBreak;
 
-            using (var file = File.CreateText(_options.BasePath + "/StrykerLogs/mutation-report.json"))
-            {
-                file.WriteLine(SerializeJsonReport(jsonReport));
-            }
+            WriteReportToJsonFile(jsonReport, Path.Combine(_options.BasePath, "/StrykerLogs/mutation-report.json"));
```
richardwerkman: Is StrykerLogs the correct location? For me `StrykerResults` sounds better. Also, if we add a timestamp to the filename more could be stored.

src/Stryker.Core/Stryker.Core/Reporters/JsonReporter.cs
```diff
@@ -106,22 +284,29 @@ public static JsonReportComponent FromProjectComponent(IReadOnlyInputComponent c
...
                     {
-                        report.ChildResults.Add(FromProjectComponent(child, thresholdOptions));
+                        report.ChildResults.Add(FromProjectComponent(child, options));
                     }
                 }
                 else if (component is FileLeaf fileComponent)
                 {
+                    report.Name = fileComponent.Name;
                     report.Source = fileComponent.SourceCode;
+                    report.Language = fileComponent.Mutants.First().Mutation.OriginalNode.Language;
```
richardwerkman: This could possibly break when no mutations are made during a stryker run. Also I'm not comfortable with the trust in the Language propery on the node. We cannot be sure this will always be "sharp". For me a string literal would be fine here.

src/Stryker.Core/Stryker.Core/Reporters/JsonReporter.cs
```diff
@@ -1,82 +1,263 @@
...
             });
+
+            _fileSystem.Directory.CreateDirectory(Path.GetDirectoryName(filePath));
+            using (var file = _fileSystem.File.CreateText(filePath))
+            {
+                file.WriteLine(json);
+            }
         }
 
         public sealed class JsonReportComponent
```
richardwerkman: Can this class be moved to another file?

src/Stryker.Core/Stryker.Core/Reporters/JsonReporter.cs
```diff
@@ -1,82 +1,263 @@
...
 using System.Linq;
 
 namespace Stryker.Core.Reporters
 {
     public class JsonReporter : IReporter
     {
         private readonly StrykerOptions _options;
-        public JsonReporter(StrykerOptions options)
+        private readonly IFileSystem _fileSystem;
+        public JsonReporter(StrykerOptions options, IFileSystem fileSystem)
```
richardwerkman: In other classes we do `IFileSystem fileSystem = null` and
` _fileSystem = fileSystem ?? new FileSystem();`

src/Stryker.Core/Stryker.Core/Reporters/JsonReporter.cs
Mobrockers: perhaps, but I would prefer to then also change the log location to StrykerResults so we don't clutter too much

src/Stryker.Core/Stryker.Core/Reporters/JsonReporter.cs
Mobrockers: For now the only purpose of this class is to be able to create the report. It is not used anywhere else and does not provide anything useful to the rest of the project. Why would you like to see it moved?

src/Stryker.Core/Stryker.Core/StrykerRunner.cs
```diff
@@ -40,6 +44,15 @@ public StrykerRunResult RunMutationTest(StrykerOptions options)
             var stopwatch = new Stopwatch();
             stopwatch.Start();
 
+            // Create output dir with gitignore
+            string strykerOutputDirectory = Path.Combine(options.BasePath, "StrykerOutput");
+            _fileSystem.Directory.CreateDirectory(strykerOutputDirectory);
+            _fileSystem.File.Create(Path.Combine(strykerOutputDirectory, ".gitignore")).Close();
```
richardwerkman: Using instead of `.Close()`.
Or is this not needed? Doesn't CreateText create or update a file? So now it gets created twice.

PR #306, opened 2019-01-11
feat(unsafe blocks): recompile with unsafe=true

fabianeichfeldt:
To prevent errors during recompile, the option /allowUnsafe is set to true


richardwerkman:
@fabianeichfeldt Thanks for your contribution. Maybe it would be nice to add a test case for this?

A simple bit of unsafe code in the integration test project could do the trick. This ensures this feature will keep on working in the future! Do you have time to add this?

fabianeichfeldt:
Of course. I've updated the PR

richardwerkman:
@fabianeichfeldt Thanks a lot! With our next release you can mutate your unsafe code ;)

Issue #183, opened 2018-10-12
Integrating with the test runner


nicojs:
By far the next biggest performance improvement is integrating with the test runner. Right now, we support the `dotnet test` command. This spins up a new dotnet process. There is a lot of overhead here before dotnet even starts executing the tests:

* OS Allocation of process resources
* dotnet allocation of heap + other overhead
* Loading of all assemblies from disk
* Exiting of the process.

It would be great if we can host the child process ourselves. In this process we can load everything once and reuse that process to run tests (each time with a different ACTIVE_MUTATION environment variable). Having the overhead just once will dramatically improve performance. 

In Stryker (for JS) this is done with a test runner plugin system. The command test runner is just one of the available implementations. Other examples are: `MochaTestRunner` (stryker-mocha-runner) and the `JestTestRunner` (stryker-jest-runner). Something similar here would be great to have. Not sure if it needs to be it's own separate dependency.

If you open a solution in Visual Studio, you can see the tests on the left. It doesn't matter if they are implemented in MSTest, NUnit or XUnit. This involves some kind of TestAdapter interface. Maybe we can leverage this api as well, so we only need to implement it once, instead of for each test runner.

Mobrockers:
Useful links?:
https://github.com/Microsoft/vstest
https://github.com/JetBrains/TeamCity.VSTest.TestAdapter/wiki

simondel:
I've removed the hacktoberfest label for now because this will take a while to research, review and develop.

nicojs:
So it should get 10 hacktoberfest labels, right? 😎

richardwerkman:
During my internship I also experimented with several testrunners. Both XUnit and NUnit have nuget packages available to run test projects from code. I couldn't get the XUnit one to run from a .NET Standard 2.0 project. But the NUnit runner seemed to work perfectly and super quick.

It offers the following advantages:
- I had to pass the runner an in memory assembly, so we can run the tests all in memory. (a rerun would be super fast)
- It offers hooks like:
  - TestStarted
  - TestFinished (with result)
- Test exporation can be configured (filtering)

This would speed up Stryker.NET enormously, but of course only on NUnit test projects... 

Would implementing a custom NUnit testrunner inside Stryker.NET be worth the work, or do we want to focus on the VSTest adapter only?

Mobrockers:
[...]

PR #319, opened 2019-01-25
feat(testrunner): Integrate with vstest testrunner on windows

Mobrockers:
First step for #183


src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
```diff
@@ -134,5 +137,19 @@ public static class CLIOptions
...
             JsonKey = "files-to-exclude"
         };
+
+        public static readonly CLIOption<string> TestRunner = new CLIOption<string>
+        {
+            ArgumentName = "--test-runner",
+            ArgumentShortName = "-tr",
+            ArgumentDescription = $"Choose which testrunner should be used to run your tests. | {FormatOptionsString(_defaultOptions.TestRunner, (IEnumerable<TestRunner>)Enum.GetValues(_defaultOptions.TestRunner.GetType()))}",
+            DefaultValue = _defaultOptions.TestRunner.ToString(),
+            JsonKey = "files-to-exclude"
```
richardwerkman: json key is incorrect

src/Stryker.Core/Stryker.Core/Options/Values/DotnetFramework.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Stryker.Core.Options.Values
+{
+    public enum DotnetFramework
+    {
+        NetFull, NetCore
```
richardwerkman:
```suggestion
        NetFull, 
        NetCore
```

src/Stryker.Core/Stryker.Core/Options/Values/Reporters.cs
```diff
@@ -1,6 +1,6 @@
-﻿namespace Stryker.Core.Options
+﻿namespace Stryker.Core.Options.Values
 {
-    public enum Reporter
+    public enum Reporters
```
richardwerkman: Change back to `Reporter`

src/Stryker.Core/Stryker.Core/Options/Values/TestRunner.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Stryker.Core.Options.Values
+{
+    public enum TestRunner
+    {
+        VsTest, DotnetTest
```
richardwerkman:
```suggestion
        VsTest, 
        DotnetTest
```

src/Stryker.Core/Stryker.Core/Reporters/ReporterFactory.cs
```diff
@@ -13,22 +13,22 @@ public static IReporter Create(StrykerOptions options)
...
 
-        private static IEnumerable<IReporter> DetermineEnabledReporters(IEnumerable<Reporter> enabledReporters, IDictionary<Reporter, IReporter> possibleReporters)
+        private static IEnumerable<IReporter> DetermineEnabledReporters(IEnumerable<Options.Values.Reporters> enabledReporters, IDictionary<Options.Values.Reporters, IReporter> possibleReporters)
         {
-            return enabledReporters.Contains(Reporter.All) ? possibleReporters.Values :
-                possibleReporters.Where(reporter => enabledReporters.Contains(reporter.Key))
-                    .Select(reporter => reporter.Value);
+            return enabledReporters.Contains(Options.Values.Reporters.All) ? possibleReporters.Values :
+                possibleReporters.Where((KeyValuePair<Options.Values.Reporters, IReporter> reporter) => enabledReporters.Contains(reporter.Key))
+                    .Select((KeyValuePair<Options.Values.Reporters, IReporter> reporter) => reporter.Value);
```
richardwerkman: Does this need to be a typed select? Can't it just be:
``` csharp
.Select(reporter => reporter.Value);
```
That reads a lot easier :)

src/Stryker.Core/Stryker.Core/Stryker.Core.csproj
```diff
@@ -31,7 +31,11 @@
     <PackageReference Include="Microsoft.Extensions.Configuration" Version="2.2.0" />
     <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="2.2.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="2.2.0" />
-    <PackageReference Include="Newtonsoft.Json" Version="12.0.1" />
+    <PackageReference Include="Microsoft.TestPlatform" Version="16.0.0-preview-20181205-02" />
+    <PackageReference Include="Microsoft.TestPlatform.ObjectModel" Version="16.0.0-dev" />
+    <PackageReference Include="Microsoft.TestPlatform.Portable" Version="16.0.0-preview-20181205-02" />
```
richardwerkman: Why the preview versions?

src/Stryker.Core/Stryker.Core/TestRunners/vstest/RunEventHandler.cs
```diff
@@ -0,0 +1,61 @@
+﻿using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Client;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging;
+using System;
+using System.Collections.Generic;
+using System.Threading;
+
+namespace Stryker.Core.TestRunners.VsTest
```
richardwerkman: The namespace is called: `VsTest` but the dir is called `vstest`. I would like to see this consistent.

src/Stryker.Core/Stryker.Core/TestRunners/vstest/StrykerVstestHostLauncher.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public int LaunchTestHost(TestProcessStartInfo defaultTestHostStartInfo, CancellationToken cancellationToken)
+        {
+            var processInfo = new ProcessStartInfo(
+                                  defaultTestHostStartInfo.FileName,
+                                  defaultTestHostStartInfo.Arguments)
+            {
+                WorkingDirectory = defaultTestHostStartInfo.WorkingDirectory
+            };
+            processInfo.EnvironmentVariables["ActiveMutation"] = _activeMutation.ToString();
```
richardwerkman: This would look better:
``` csharp
var processInfo = new ProcessStartInfo(defaultTestHostStartInfo.FileName, defaultTestHostStartInfo.Arguments)
{
    WorkingDirectory = defaultTestHostStartInfo.WorkingDirectory,
    EnvironmentVariables["ActiveMutation"] = _activeMutation.ToString()
};
```

src/Stryker.Core/Stryker.Core/TestRunners/vstest/VsTestRunner.cs
```diff
@@ -0,0 +1,167 @@
...
+            consoleWrapper = new VsTestConsoleWrapper(vsTestToolPath, new ConsoleParameters { TraceLevel = DetermineTraceLevel(), LogFilePath = Path.Combine(_options.BasePath, "StrykerOutput", "logs", "vstest-log.txt") });
+
+            consoleWrapper.StartSession();
+            consoleWrapper.InitializeExtensions(new List<string> { testBinariesPath, vsTestExtensionsPath });
+
+            var testCases = DiscoverTests(new List<string>() { testAssemblyPath }, consoleWrapper);
+
+            var testResults = RunAllTests(consoleWrapper, new List<string> { testAssemblyPath }, activeMutationId);
+
+            // For now we need to throw an OperationCanceledException when a testrun has timed out. We know the testrun has timed out because there is no result for the test.
```
richardwerkman:
```suggestion
            // For now we need to throw an OperationCanceledException when a testrun has timed out. We know the testrun has timed out because there are less results for the test, than test cases.
```

src/Stryker.Core/Stryker.Core/TestRunners/vstest/VsTestRunner.cs
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        public TestRunResult RunAll(int? timeoutMS, int? activeMutationId)
+        {
+            _timeoutMS = timeoutMS ?? 0;
+            GenerateRunSettings();
+
+            var testBinariesPath = FilePathUtils.ConvertPathSeparators(Path.Combine(_options.BasePath, "bin", "Debug", _projectInfo.TargetFramework));
+            var testAssemblyPath = FilePathUtils.ConvertPathSeparators(Path.Combine(testBinariesPath, _projectInfo.TestProjectFileName.Replace("csproj", "dll")));
+            //var vsTestToolPath = _vsTestHelper.GetVsTestToolPaths()[_runnerFramework];
```
richardwerkman: Remove commented code

src/Stryker.Core/Stryker.Core/TestRunners/vstest/VsTestRunner.cs
```diff
@@ -0,0 +1,167 @@
...
+                    return TraceLevel.Info;
+                default:
+                    return TraceLevel.Off;
+            }
+        }
+
+        private void GenerateRunSettings()
+        {
+            string targetFramework = _projectInfo.TargetFramework;
+            if (targetFramework.Contains("netcoreapp") || targetFramework.Contains("netstandard"))
```
richardwerkman: The test project can't target `netstandard`.

src/Stryker.Core/Stryker.Core/TestRunners/vstest/VsTestRunner.cs
```diff
@@ -0,0 +1,167 @@
...
+                _runnerFramework = DotnetFramework.NetCore;
+            }
+
+            string targetFrameworkVersion = Regex.Replace(targetFramework, @"[^.\d]", "");
+            switch (targetFramework)
+            {
+                case string s when s.Contains("netcoreapp"):
+                    targetFrameworkVersion = $".NETCoreApp,Version = v{targetFrameworkVersion}";
+                    break;
+                case string s when s.Contains("netstandard"):
```
richardwerkman: Same here

src/Stryker.Core/Stryker.Core/ToolHelpers/VsTestHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+        public VsTestHelper(StrykerOptions options, IFileSystem fileSystem = null)
+        {
+            _options = options;
+            _fileSystem = fileSystem ?? new FileSystem();
+        }
+
+        public Dictionary<DotnetFramework, string> GetVsTestToolPaths()
+        {
+            Dictionary<DotnetFramework, string> vsTestPaths = new Dictionary<DotnetFramework, string>();
+            string versionString = "16.0.0-preview-20181205-02";
```
richardwerkman: Can we get rid of the hard coded version number?

src/Stryker.Core/Stryker.Core/ToolHelpers/VsTestHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+            foreach (string nugetPackageFolder in nugetPackageFolders)
+            {
+                if (dllFound && exeFound)
+                {
+                    break;
+                }
+
+                string portablePackageFolder = _fileSystem.Directory.GetDirectories(nugetPackageFolder, portablePackageName, SearchOption.AllDirectories).First();
+
+                string dllPath = FilePathUtils.ConvertPathSeparators(
```
richardwerkman: Can't we search the whole `portablePackageFolder` for files with the name: `vstest.console` as a filter. That way even if the path insided the package ever changes we can still locate vstest

src/Stryker.Core/Stryker.Core/Stryker.Core.csproj
Mobrockers: Because there has not been a full release for a while, and the change we will need for vstest.console.dll integration will most likely be a preview version as well. This is also why vstest is not yet the default test runner.

src/Stryker.Core/Stryker.Core/TestRunners/vstest/StrykerVstestHostLauncher.cs
Mobrockers: This is not possible. EnvironmentVariables is a System.Collections.Specialized.StringDictionary. You cannot instantiate this type in that way. I would also be hesitant to overwrite the whole environment variables 'dictionary' so I would prefer not to instantiate this property. Therefore we need to add to the dictionary outside of the instantiation of ProcessStartInfo.

src/Stryker.Core/Stryker.Core/ToolHelpers/VsTestHelper.cs
Mobrockers: I have not been able to come up with a better way for now..

src/Stryker.Core/Stryker.Core/ToolHelpers/VsTestHelper.cs
Mobrockers: We need those specific versions of vstest.console. If the location changes we will have to investigate whether we are still compatible anyhow and our integration tests will fail on it so I do not see any problem in keeping it how it is.

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
```diff
@@ -0,0 +1,161 @@
...
+            IVsTestConsoleWrapper consoleWrapper = null;
+
+            consoleWrapper = new VsTestConsoleWrapper(vsTestToolPath, new ConsoleParameters { TraceLevel = DetermineTraceLevel(), LogFilePath = Path.Combine(_options.BasePath, "StrykerOutput", "logs", "vstest-log.txt") });
+
+            consoleWrapper.StartSession();
+            consoleWrapper.InitializeExtensions(new List<string> { testBinariesPath, vsTestExtensionsPath });
+
+            var testCases = DiscoverTests(new List<string>() { testAssemblyPath }, consoleWrapper);
+
+            var testResults = RunAllTests(consoleWrapper, new List<string> { testAssemblyPath }, activeMutationId);
```
richardwerkman: I think all code above this line in this method can be moved to a other method. I think this needs to only be executed once, not for every mutant. This might speed up things too?

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
```diff
@@ -0,0 +1,161 @@
...
+            var testResult = new TestRunResult
+            {
+                Success = testResults.All(tr => tr.Outcome == TestOutcome.Passed),
+                ResultMessage = string.Join(
+                    Environment.NewLine,
+                    testResults.Where(tr => !string.IsNullOrWhiteSpace(tr.ErrorMessage)).Select(tr => tr.ErrorMessage)),
+                TotalNumberOfTests = testCases.Count()
+            };
+
+            consoleWrapper.EndSession();
```
richardwerkman: We should somehow call this when all mutants have been tested. That way we keep the session open for the whole stryker run.

src/Stryker.Core/Stryker.Core/TestRunners/TestRunnerFactory.cs
```diff
@@ -14,20 +18,31 @@ public TestRunnerFactory()
...
             {
-                case("dotnet test"):
-                    testRunner = new DotnetTestRunner(path, new ProcessExecutor(), new TotalNumberOfTestsParser());
+                _logger.LogWarning("Using testrunner {0} because OS is not windows. Stryker does not currently support vstest on non-windows OS.", TestRunner.DotnetTest.ToString());
+                return new DotnetTestRunner(projectInfo.TestProjectPath, new ProcessExecutor(), new TotalNumberOfTestsParser());
+            }
+
+            switch (options.TestRunner)
+            {
+                case TestRunner.DotnetTest:
```
richardwerkman: This is a duplicate of the default. It could be removed.

src/Stryker.Core/Stryker.Core/Stryker.Core.csproj
richardwerkman: The current implementation only works on windows. Wouldn't it be best to use the last stable packages for now? When we add vstest for linux and mac we can update the package version.

Mobrockers: Agreed. Downgrading for now.

Mobrockers:
PR Merging is currently blocked because our way of running multiple stryker runs at the same time seems to break vstest in some way. A solution needs to be found first. A possible temporary solution is to limit stryker execution to one thread when vstest is used as testrunner. Because we can re-use the same vstest instance for all tests this might actually still be faster than dotnet test without even running multi-threaded.

Mobrockers:
@richardwerkman Ready for re-review and merge

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunnerFactory.cs
```diff
@@ -0,0 +1,56 @@
+﻿using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using Stryker.Core.Initialisation;
+using Stryker.Core.Options;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+
+namespace Stryker.Core.TestRunners.VsTest
+{
+    public class VsTestRunnerFactory : ITestRunner
```
richardwerkman: I find the name `VsTestRunnerFactory` a bit confusing. Maybe `VsTestRunnerObjectPool` covers the subject better?

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
```diff
@@ -0,0 +1,212 @@
...
+            TestRunResult testResult = new TestRunResult() { Success = false };
+            lock (runLock)
+            {
+                Running = true;
+
+                var testResults = RunAllTests(activeMutationId, GenerateRunSettings(timeoutMS ?? 0));
+
+                // For now we need to throw an OperationCanceledException when a testrun has timed out. 
+                // We know the testrun has timed out because we received less test results from the test run than there are test cases in the unit test project.
+                if (testResults.Count() < _testCases.Count())
```
richardwerkman: For every testrun all testcases are being count. It looks like a simple integer property would do the job too.

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
```diff
@@ -0,0 +1,212 @@
...
+                    throw new OperationCanceledException();
+                }
+
+                testResult = new TestRunResult
+                {
+                    Success = testResults.All(tr => tr.Outcome == TestOutcome.Passed),
+                    ResultMessage = string.Join(
+                        Environment.NewLine,
+                        testResults.Where(tr => !string.IsNullOrWhiteSpace(tr.ErrorMessage)).Select(tr => tr.ErrorMessage)),
+                    TotalNumberOfTests = _testCases.Count()
```
richardwerkman: Same here

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
```diff
@@ -0,0 +1,212 @@
...
+                case string s when s.Contains("netstandard"):
+                    throw new Exception("Unsupported targetframework detected. A unit test project cannot be netstandard!: " + targetFramework);
+                default:
+                    targetFrameworkVersion = $".NETFramework = v{targetFrameworkVersion}";
+                    break;
+            }
+
+            return $@"<RunSettings>
+  <RunConfiguration>
+    <MaxCpuCount>1</MaxCpuCount>
```
richardwerkman: Why is this set to 1? We might get even more performance if we set this higher?

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunnerFactory.cs
```diff
@@ -0,0 +1,56 @@
...
+            for (var i = 0; i < options.ConcurrentTestrunners; i++)
+            {
+                _availableRunners.Add(new VsTestRunner(options, projectInfo, testCases));
+            }
+        }
+
+        public TestRunResult RunAll(int? timeoutMS, int? activeMutationId)
+        {
+            TestRunResult result = null;
+            while (result is null)
```
richardwerkman: Is this busy waiting needed? Can't we solve this with event handlers? Like, set a handler when an object is free again and wait here till the object is free.

src/Stryker.Core/Stryker.Core/ToolHelpers/VsTestHelper.cs
```diff
@@ -0,0 +1,140 @@
...
+        public VsTestHelper(StrykerOptions options, IFileSystem fileSystem = null)
+        {
+            _options = options;
+            _fileSystem = fileSystem ?? new FileSystem();
+        }
+
+        public string GetCurrentPlatformVsTestToolPath()
+        {
+            var paths = GetVsTestToolPaths();
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
richardwerkman: Suggestion:

``` csharp
foreach(var path in paths)
{
     if (RuntimeInformation.IsOSPlatform(path.Key))
     {
         return path.Value;
     }
}
throw new PlatformNotSupportedException(
     $"The current OS is not any of the following supported: { OSPlatform.Windows.ToString() }, { OSPlatform.Linux.ToString() } or { OSPlatform.OSX.ToString() }");
```

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
Mobrockers: I set the max cpu count to 1 because the testruns are already distributed over multiple threads with different test runners by stryker. I don't believe it would be beneficial to performance if vstest is then also allowed to spawn even more threads.

Issue #18, opened 2018-06-08
Large projects


richardwerkman:
On large projects, it might occur that not all compile errors are returned by Roslyn after the first compile try. If this happens, not all "broken" mutations will be rollbacked.

For now this problem has not occurred yet. Stryker should be tested on a large project to discover if this is an existing bug or not.

Mobrockers:
@richardwerkman Do you feel like we have sufficiently fixed this now, or should we try this on even larger projects?

richardwerkman:
I think this will work for extremely large projects, if not people can open new issues for their case.

PR #323, opened 2019-01-27
fix(Missed compilation erros): Rollback two times instead of one

### LINKED ISSUE #304: Why is "a"+"b" mutated to "a"-"b" and compiles?
How comes that the mutation report has lines like this:
`[Survived] Binary expression mutation on line 26: '"" + toString(x)' ==> '"" - toString(x)'`

Mutating `"a"+"b"` to `"a"-"b"` is not valid C# code if I am not totally wrong?  So why is this a valid mutation that even passes the execution of my tests?

richardwerkman:
On large projects not all compilation errors are returned by the Roslyn compiler on the first compilation. The rollback process does not rollback all broken mutations on the rirst try. This PR adds a seconds rollback step to also catch all missed compilation errors.

closes #304 and possibly #18

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Compiling/CompilingProcess.cs
+++ b/src/Stryker.Core/Stryker.Core/Compiling/CompilingProcess.cs
@@ -71,6 +71,19 @@ namespace Stryker.Core.Compiling
                 emitResult = rollbackProcessResult.Compilation.Emit(ms);
             }
 
+            if (!emitResult.Success)
+            {
+                LogEmitResult(emitResult);
+                // remove any left over broken mutations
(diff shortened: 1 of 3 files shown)


src/Stryker.Core/Stryker.Core/Compiling/CompilingProcess.cs
```diff
@@ -71,6 +71,19 @@ public CompilingProcessResult Compile(IEnumerable<SyntaxTree> syntaxTrees, Memor
                 emitResult = rollbackProcessResult.Compilation.Emit(ms);
             }
 
+            if (!emitResult.Success)
```
Mobrockers: This should be extracted into a method since we are repeating ourselves.

PR #350, opened 2019-02-22
Support DeclarationPattern

dupdob:
Fix issues on Declaration pattern + provides location detail when rollback fails.
Changes are minimal

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_IfStatementsShouldBe_Nested_IN.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_IfStatementsShouldBe_Nested_IN.cs
@@ -30,6 +30,11 @@ namespace StrykerNet.UnitTest.Mutants.TestResources
             {
                 return i + 1;
             }
+
+            if (i is int x)
+            {
+                return x + 1;
+            }
         }
 
         private bool Out(out string test)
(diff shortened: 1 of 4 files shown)


richardwerkman:
Thank you for yet another contribution! Looks good to me 👍

dupdob:
hello @richardwerkman: I just realised that the number of Mutants for NFluent has been roughly halved (1660 => 930). I don't think this relates to my PR (I tested before pulling latest version and it gave me ok results).
Is this expected from your point of view?

Mobrockers:
We have removed some linq mutations that never actually worked. Could you diff your results and see what type of mutations you are missing or do you not store the json result?

richardwerkman:
I'm curious to what kind of mutations are missing indeed. I did a major rewrite on the mutation logic in #336 so maybe that has some side effects that can only be noticed on larger projects?

On the small integration test project the number of mutations has been the same.

dupdob:
I think I only have a handful of Linq mutants, so the difference is somewhere else. 
I will do runs to see what changed

dupdob:
It did some analysis and I am pretty sure mutants are missing (string mutation, but probably not only).
Here is a bit of code which is not mutated but I would expect to have mutants for constant strings.
```csharp
namespace Test 
{
    using System;
    public static class CharCheckExtensions
    {
        public static string ChainCalls(string check)
        {
            return check.Replace("ab", "cd")
                .Replace("12", "34")
                .PadLeft(12)
                .Replace("12", "34");
        }
    }
}
```

dupdob:
I now understand the cause of the problem, I will create an issue to capture it.

Issue #353, opened 2019-02-24
Original syntaxnode does not contain full parent syntax tree for return statement nodes


Mobrockers:
This results in being unable to correctly report the mutation and original node location in source code as the line number is always 0 (as the return statement is the full syntax tree and the mutation is on the return statement value/expression).

Mobrockers:
Traced the issue to here: https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs#L129-L134

```
// No statement found yet, search deeper in the tree for statements to mutate
var children = currentNode.ChildNodes().ToList(); // Children have parent here
var childCopy = currentNode.TrackNodes(children); // Childcopy has lost it's parent
foreach (var child in children)
{
    var mutatedNode = Mutate(child);
    var originalNode = childCopy.GetCurrentNode(child); // So original node here has current node as parent, but not it's parent. Syntaxtree is broken.
```

richardwerkman:
Do you have an piece of code available that results in a mutation with a wrong span? I could look into this if I have a reproduction project.

Mobrockers:
Our integrationtest StringMagic.Demo method on the return true -> false and DummyMath.Add method on the checked(a + b) -> a + b both have this behaviour

richardwerkman:
The object `currentNode` still has the parent node available. We need a way to return the copy node with the parent of the currentNode.

``` cs
// No statement found yet, search deeper in the tree for statements to mutate
var children = currentNode.ChildNodes().ToList();
var childCopy = currentNode.TrackNodes(children);
foreach (var child in children)
{
    var mutatedNode = Mutate(child);
    var originalNode = childCopy.GetCurrentNode(child);
    if (!mutatedNode.IsEquivalentTo(originalNode))
    {
         childCopy = childCopy.ReplaceNode(originalNode, mutatedNode);
    }
}
// currentNode still has the parent here.
// childCopy has parent null
// pseudocode:
// childCopy.parent = currentNode.parent
return childCopy;
```

richardwerkman:
I have to review my last comment. It is not of importance that the node that is returned has a parent node. What is of importance is that the currentNode in `Mutate` always has a parent node. 

When does this not occur? If we call `replaceNode` on a tree and then call `Mutate` on that replaced tree. The method replaceNode removes the parent node of the tree. 

So we have to introduce a new thumb rule for the whole MutantOrchestrator: 

[...]

PR #373, opened 2019-03-03
fix(Missing mutations): Improve mutation algorithm to find more mutants

dupdob:
# Main change
## Other changes
- identify slightly more mutants that in the current release
- avoid generating known to fail mutants (e.g. default value for parameter or string concatenation)
- dump CompileError mutants in devmode

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_IN.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_IN.cs
@@ -4,6 +4,7 @@
     public static class CharCheckExtensions
     {
         
+        // Following code lines have been extracted from NFluent source code (www.n-fluent.net)
         private static void ImplementContains<T>(ICheckLogic<IEnumerable<T>> checker, IEnumerable<T> otherEnumerable)
         {
             IList<object> notFoundValues = null;
@@ -15,6 +16,7 @@
                 .OnNegate("The {0} contains all the given values whereas it must not.")
                 .EndCheck();
         }
+
         public static ICheckLink<ICodeCheck<T>> LastsLessThan<T>(
             this ICodeCheck<T> check,
             double threshold,
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_IN.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_IN.cs
@@ -1,0 +1,26 @@
+﻿namespace Test 
+{
+    using System;
+    public static class CharCheckExtensions
+    {
+        [Obsolete("thismustnotbemutated")]
+        public static ICheckLink<ICheck<char>> IsALetter(this ICheck<char> check)
+        {
+            var x=1;
+            x *=x+2;
+            return ExtensibilityHelper.BeginCheck(check).
+                FailWhen(sut => !IsALetter(sut), "The {0} is not a letter.").
+                OnNegate("The {0} is a letter whereas it must not.").
+                EndCheck();
+        }
+
+        public static string ChainCalls(string check)
+        {
+            return check.Replace("ab", "cd")
+                .Replace("12", "34")
+                .PadLeft(12)
+                .Replace("12", "34")
+                .Normalize();
+        }
+    }
+}
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_IN.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_IN.cs
@@ -3,6 +3,18 @@
     using System;
     public static class CharCheckExtensions
     {
+        
+        private static void ImplementContains<T>(ICheckLogic<IEnumerable<T>> checker, IEnumerable<T> otherEnumerable)
+        {
+            IList<object> notFoundValues = null;
+            checker
+                .FailWhen((sut) => sut == null && otherEnumerable != null, "The {0} is null and thus, does not contain the given expected value(s).")
+                .DefineExpectedValues(otherEnumerable, otherEnumerable.Count())
+                .Analyze((sut, _) => notFoundValues = ExtractNotFoundValues(sut, otherEnumerable))
+                .FailWhen((_) => notFoundValues.Any(), string.Format("The {{0}} does not contain the expected value(s):" + Environment.NewLine + "\t{0}", notFoundValues.ToStringProperlyFormatted().DoubleCurlyBraces()))
+                .OnNegate("The {0} contains all the given values whereas it must not.")
+                .EndCheck();
+        }
         public static ICheckLink<ICodeCheck<T>> LastsLessThan<T>(
             this ICodeCheck<T> check,
             double threshold,
@@ -30,6 +42,5 @@
                 OnNegate("The {0} is a letter whereas it must not.").
                 EndCheck();
         }
-
     }
 }
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_IN.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_IN.cs
@@ -3,24 +3,33 @@
     using System;
     public static class CharCheckExtensions
     {
-        [Obsolete("thismustnotbemutated")]
+        public static ICheckLink<ICodeCheck<T>> LastsLessThan<T>(
+            this ICodeCheck<T> check,
+            double threshold,
+            TimeUnit timeUnit)
+            where T : RunTrace
+        {
+            var durationThreshold = new Duration(threshold, timeUnit);
+
+            ExtensibilityHelper.BeginCheck(check).
+                SetSutName("code").
+                CheckSutAttributes(sut =>  new Duration(sut.ExecutionTime, timeUnit), "execution time").
+                FailWhen((sut) => sut > durationThreshold, "The {checked} was too high.").
+                DefineExpectedValue(durationThreshold, "less than", "more than").
+                OnNegate("The {checked} was too low.").
+                EndCheck();
+
+            return ExtensibilityHelper.BuildCheckLink(check);
+        }
+
         public static ICheckLink<ICheck<char>> IsALetter(this ICheck<char> check)
         {
-            var x=1;
-            x *=x+2;
+
             return ExtensibilityHelper.BeginCheck(check).
                 FailWhen(sut => !IsALetter(sut), "The {0} is not a letter.").
                 OnNegate("The {0} is a letter whereas it must not.").
                 EndCheck();
         }
 
-        public static string ChainCalls(string check)
-        {
-            return check.Replace("ab", "cd")
-                .Replace("12", "34")
-                .PadLeft(12)
-                .Replace("12", "34")
-                .Normalize();
-        }
     }
 }
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_OUT.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_OUT.cs
@@ -9,9 +9,9 @@
             IList<object> notFoundValues = null;
             checker
                 .FailWhen((sut) => Stryker.ActiveMutationHelper.ActiveMutation==2?sut == null && otherEnumerable == null:Stryker.ActiveMutationHelper.ActiveMutation==1?sut != null && otherEnumerable != null:Stryker.ActiveMutationHelper.ActiveMutation==0?sut == null || otherEnumerable != null:sut == null && otherEnumerable != null, Stryker.ActiveMutationHelper.ActiveMutation==3?"":"The {0} is null and thus, does not contain the given expected value(s).")
-                .DefineExpectedValues(otherEnumerable, Stryker.ActiveMutationHelper.ActiveMutation==4?otherEnumerable.Sum():otherEnumerable.Count())
+                .DefineExpectedValues(otherEnumerable, otherEnumerable.Count())
                 .Analyze((sut, _) => notFoundValues = ExtractNotFoundValues(sut, otherEnumerable))
-                .FailWhen((_) => notFoundValues.Any(), Stryker.ActiveMutationHelper.ActiveMutation==8?string.Format("The {{0}} does not contain the expected value(s):" + Environment.NewLine + "", notFoundValues.ToStringProperlyFormatted().DoubleCurlyBraces()):Stryker.ActiveMutationHelper.ActiveMutation==7?string.Format(""+ Environment.NewLine + "\t{0}", notFoundValues.ToStringProperlyFormatted().DoubleCurlyBraces()):string.Format("The {{0}} does not contain the expected value(s):" + Environment.NewLine + "\t{0}", notFoundValues.ToStringProperlyFormatted().DoubleCurlyBraces()))
+                .FailWhen((_) => notFoundValues.Any(), string.Format(Stryker.ActiveMutationHelper.ActiveMutation==6?"The {{0}} does not contain the expected value(s):" + Environment.NewLine + "":Stryker.ActiveMutationHelper.ActiveMutation==5?""+ Environment.NewLine + "\t{0}":"The {{0}} does not contain the expected value(s):" + Environment.NewLine + "\t{0}", notFoundValues.ToStringProperlyFormatted().DoubleCurlyBraces()))
                 .OnNegate(Stryker.ActiveMutationHelper.ActiveMutation==9?"":"The {0} contains all the given values whereas it must not.")
                 .EndCheck();
         }
@@ -37,18 +37,9 @@
         public static ICheckLink<ICheck<char>> IsALetter(this ICheck<char> check)
         {
 
-            return Stryker.ActiveMutationHelper.ActiveMutation==23?ExtensibilityHelper.BeginCheck(check).
-                FailWhen(sut => !IsALetter(sut), "The {0} is not a letter.").
-                OnNegate("").
-                EndCheck():Stryker.ActiveMutationHelper.ActiveMutation==22?ExtensibilityHelper.BeginCheck(check).
-                FailWhen(sut => !IsALetter(sut), "").
-                OnNegate("The {0} is a letter whereas it must not.").
-                EndCheck():Stryker.ActiveMutationHelper.ActiveMutation==21?ExtensibilityHelper.BeginCheck(check).
-                FailWhen(sut => IsALetter(sut), "The {0} is not a letter.").
-                OnNegate("The {0} is a letter whereas it must not.").
-                EndCheck():ExtensibilityHelper.BeginCheck(check).
-                FailWhen(sut => !IsALetter(sut), "The {0} is not a letter.").
-                OnNegate("The {0} is a letter whereas it must not.").
+            return ExtensibilityHelper.BeginCheck(check).
+                FailWhen(sut => Stryker.ActiveMutationHelper.ActiveMutation==18?IsALetter(sut):!IsALetter(sut), Stryker.ActiveMutationHelper.ActiveMutation==19?"":"The {0} is not a letter.").
+                OnNegate(Stryker.ActiveMutationHelper.ActiveMutation==20?"":"The {0} is a letter whereas it must not.").
                 EndCheck();
         }
     }
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_OUT.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_OUT.cs
@@ -3,25 +3,33 @@
     using System;
     public static class CharCheckExtensions
     {
-        [Obsolete("thismustnotbemutated")]
+        public static ICheckLink<ICodeCheck<T>> LastsLessThan<T>(
+            this ICodeCheck<T> check,
+            double threshold,
+            TimeUnit timeUnit)
+            where T : RunTrace
+        {
+            var durationThreshold = new Duration(threshold, timeUnit);
+
+            ExtensibilityHelper.BeginCheck(check).
+                SetSutName(Stryker.ActiveMutationHelper.ActiveMutation==0?"":"code").
+                CheckSutAttributes(sut =>  new Duration(sut.ExecutionTime, timeUnit), Stryker.ActiveMutationHelper.ActiveMutation==1?"":"execution time").
+                FailWhen((sut) => Stryker.ActiveMutationHelper.ActiveMutation==3?sut >= durationThreshold:Stryker.ActiveMutationHelper.ActiveMutation==2?sut < durationThreshold:sut > durationThreshold, Stryker.ActiveMutationHelper.ActiveMutation==4?"":"The {checked} was too high.").
+                DefineExpectedValue(durationThreshold, Stryker.ActiveMutationHelper.ActiveMutation==5?"":"less than", Stryker.ActiveMutationHelper.ActiveMutation==6?"":"more than").
+                OnNegate(Stryker.ActiveMutationHelper.ActiveMutation==7?"":"The {checked} was too low.").
+                EndCheck();
+
+            return ExtensibilityHelper.BuildCheckLink(check);
+        }
+
         public static ICheckLink<ICheck<char>> IsALetter(this ICheck<char> check)
         {
-            var x=1;
-if(Stryker.ActiveMutationHelper.ActiveMutation==0){            x /=x+2;
-}else{            x *=Stryker.ActiveMutationHelper.ActiveMutation==1?x-2:x+2;
-}            return ExtensibilityHelper.BeginCheck(check).
-                FailWhen(sut => Stryker.ActiveMutationHelper.ActiveMutation==2?IsALetter(sut):!IsALetter(sut), Stryker.ActiveMutationHelper.ActiveMutation==3?"":"The {0} is not a letter.").
-                OnNegate(Stryker.ActiveMutationHelper.ActiveMutation==4?"":"The {0} is a letter whereas it must not.").
+
+            return ExtensibilityHelper.BeginCheck(check).
+                FailWhen(sut => Stryker.ActiveMutationHelper.ActiveMutation==8?IsALetter(sut):!IsALetter(sut), Stryker.ActiveMutationHelper.ActiveMutation==9?"":"The {0} is not a letter.").
+                OnNegate(Stryker.ActiveMutationHelper.ActiveMutation==10?"":"The {0} is a letter whereas it must not.").
                 EndCheck();
         }
 
-        public static string ChainCalls(string check)
-        {
-            return check.Replace(Stryker.ActiveMutationHelper.ActiveMutation==5?"":"ab", Stryker.ActiveMutationHelper.ActiveMutation==6?"":"cd")
-                .Replace(Stryker.ActiveMutationHelper.ActiveMutation==7?"":"12", Stryker.ActiveMutationHelper.ActiveMutation==8?"":"34")
-                .PadLeft(12)
-                .Replace(Stryker.ActiveMutationHelper.ActiveMutation==9?"":"12", Stryker.ActiveMutationHelper.ActiveMutation==10?"":"34")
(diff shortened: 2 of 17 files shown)


dupdob:
One mutant has not been generated during integration tests. Looking into this.
(I did not run them locally as I was unable to locate them until now)

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -121,27 +113,140 @@ public SyntaxNode Mutate(SyntaxNode currentNode)
...
+                    return newNode.ReplaceNode(originalFort, forStatement);
+                }
+                else
+                {
+                    result = MutateWithIfStatements(statement);
+                }
+
+                if (result != statement)
+                {
+                    var node = ((result as IfStatementSyntax).Else.Statement as BlockSyntax).Statements[0];
```
richardwerkman: Same here, tracknodes and getCurrentNode would be better here.

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -121,27 +113,140 @@ public SyntaxNode Mutate(SyntaxNode currentNode)
...
+                    var currentExpressionSyntax = currentNodeCopy.GetCurrentNode(expressionSyntax);
+                    SyntaxNode actualMutationNode = null;
+                    if (currentNode is AssignmentExpressionSyntax assign)
+                    {
+                        actualMutationNode = assign.Right;
+                    }
+                    else if (currentNode is ExpressionStatementSyntax)
+                    {
+                        // The expression of a ExpressionStatement cannot be mutated directly
+                        return MutateInPlace(currentNodeCopy, currentExpressionSyntax);
```
richardwerkman: Mutate should not be called with a copy of the current node. This is where the parent node is already `null`. All mutations produced by this statement will have incorrect spans (and thus incorrect locations in the report).

src/Stryker.Core/Stryker.Core/SyntaxHelper.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Stryker.Core
+{
+    // extension methods for SyntaxNode(s)
+    internal static class SyntaxHelper
+    {
+        public static bool IsString(this ExpressionSyntax node)
```
richardwerkman: I think `IsStringLiteral` is a better name. This does not reassure that the Expression is a string. An variableNameExpression can be of type string too.

richardwerkman:
After some research I found the reason all mutations have incorrect spans. I wrote my explanation in #353. If you apply the rule I mention there all spans should be correct again.

src/Stryker.Core/Stryker.Core/SyntaxHelper.cs
dupdob: Fair point, naming is incorrect here. I was not really happy with IsStringLiteral, as the check is broader.
I settled for _IsAStringExpression_ which is very close to the intent: is the expression of a string type, so we do not try arithmetic mutant on it.

dupdob:
Thanks for the pointer @richardwerkman; this does not come as a surprise.
I will address this and the issue you raised in your comments in the near future. Still focused on getting something that mutates properly, then address location and tracking issues

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -121,27 +113,140 @@ public SyntaxNode Mutate(SyntaxNode currentNode)
...
+                            var newFor = ApplyMutant(forStatement, mutant);
+
+                            newNode = MutantPlacer.PlaceWithIfStatement(statement, newFor, mutant.Id);
+                        }
+                    }
+
+                    var originalFort = forStatement;
+                    if (newNode != originalFort)
+                    {
+                        originalFort = (ForStatementSyntax) ((BlockSyntax)((IfStatementSyntax) newNode).Else.Statement).Statements[0];
```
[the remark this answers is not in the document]
dupdob: This one is ugly as hell plus it is buggy. I need to use trackNodes, but my first attempt failed, so I went for an ugly fix as stop gap solution

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
dupdob: Thanks for pointing my errors.
Fixed

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
dupdob: Fixed.
Again, thanks for the pointer.
Only **for** statement remains to be done.

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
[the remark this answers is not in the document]
dupdob: Well, I am working on the correct version and this one is tricky, as I have to generate several mutants of the for statement, a naive TrackNode approach fails (more than one node found)...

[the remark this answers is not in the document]
dupdob: This one was harder, due to the specificities of for.
Done.
At least another one remains.

dupdob:
Las known issues have been addressed.

## PR is now ready for consideration and merge!!
Remarks: due to the heavy refactoring of the MutantOrchestrator class, I advise to perform tests on as many project as possible: regression may have crept in.
I have tested it against NFluent with success as well as the integration tests, but varying coding styles projects could help identify potential issues.

And thanks @richardwerkman for his support during the stabilisation phase.

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -76,72 +216,66 @@ public SyntaxNode Mutate(SyntaxNode currentNode)
...
         private IEnumerable<Mutant> FindMutants(SyntaxNode current)
+        {
+            foreach (var mutant1 in FlatMutate(current)) yield return mutant1;
+            foreach (var mutant in current.ChildNodes().SelectMany(FindMutants))
+            {
+                yield return mutant;
+            }
+        }
+
+        private IEnumerable<Mutant> FlatMutate(SyntaxNode current)
```
richardwerkman: I find this name somewhat confusing. Maybe `FindMutants()` and `FindMutantsRecursive()` would make more sence for these two methods.

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -62,12 +64,150 @@ public IEnumerable<Mutant> GetLatestMutantBatch()
             return tempMutants;
         }
 
+
+        private bool CanBeMutated(SyntaxNode node)
```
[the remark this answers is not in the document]
richardwerkman: Or rewrite this to an extension method. It might be nice to create a folder of `SyntaxNodeExtensions` that contains `CanBeMutated` and `IsStringLiteral`. We might think of even more. This would put all SyntaxNode logic in one place and clean up this huge class a bit.

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -62,12 +64,150 @@ public IEnumerable<Mutant> GetLatestMutantBatch()
...
+                    return result;
+                }
+
+                return result.ReplaceNode(result.GetCurrentNode(expressionStatement), MutateExpressions(expressionStatement));
+            }
+
+            return MutateExpressions(expressionStatement);
+        }
+
+
```
richardwerkman: Double enter

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -31,6 +32,7 @@ public class MutantOrchestrator : IMutantOrchestrator
         private IEnumerable<IMutator> _mutators { get; set; }
         private ILogger _logger { get; set; }
 
+        private static Type[] expressionStatementNeedingIf = { typeof(AssignmentExpressionSyntax), typeof(PostfixUnaryExpressionSyntax), typeof(PrefixUnaryExpressionSyntax)};
```
richardwerkman: Why is this static?

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -62,12 +64,150 @@ public IEnumerable<Mutant> GetLatestMutantBatch()
...
+            foreach (var incrementor in forStatement.Incrementors)
+            {
+                trackedCopy = MutateSubExpressionWithIfStatements(forStatement, trackedCopy, incrementor);
+            }
+
+            var originalFor = trackedCopy;
+            if (trackedCopy != forStatement)
+            {
+                var scan = trackedCopy;
+                while (scan is IfStatementSyntax mutantIf)
```
richardwerkman: What is going on here?

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -62,12 +64,150 @@ public IEnumerable<Mutant> GetLatestMutantBatch()
...
+                mutatedIf = mutatedIf.ReplaceNode(mutatedIf.GetCurrentNode(ifStatement.Else), Mutate(ifStatement.Else));
+            }
+
+            return mutatedIf.ReplaceNode(mutatedIf.GetCurrentNode(ifStatement.Statement), Mutate(ifStatement.Statement));
+        }
+
+        private SyntaxNode MutateForStatement(ForStatementSyntax forStatement)
+        {
+            // for needs special treatments for its incrementors
+            StatementSyntax trackedCopy = forStatement;
```
richardwerkman: Why is this called `trackedCopy`? It doesn't seem to be tracked anywhere so the name is misleading.

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -62,12 +64,150 @@ public IEnumerable<Mutant> GetLatestMutantBatch()
...
+                {
+                    scan = ((BlockSyntax) mutantIf.Else.Statement).Statements[0];
+                }
+
+                originalFor = (ForStatementSyntax) scan;
+            }
+
+            // now we generate a mutant for the remainder of the for statement
+            var mutatedFor = forStatement.TrackNodes(forStatement.Condition, forStatement.Statement);
+            mutatedFor = mutatedFor.ReplaceNode(mutatedFor.GetCurrentNode(forStatement.Condition),
```
richardwerkman: `mutatedFor.GetCurrentNode` in all other methods we call the tracked nodes `Copy`. Why not here?

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -76,72 +216,66 @@ public SyntaxNode Mutate(SyntaxNode currentNode)
...
-                        }
-                        else
-                        {
-                            // If the EpxressionStatement does not contain a expression that can be mutated with conditional expression...
-                            // it should be mutated with if statements
-                            return MutateWithIfStatements(currentNode as ExpressionStatementSyntax);
-                        }
+                        actualMutationNode = assign.Right;
+                    }
+                    else if (currentNode is ExpressionStatementSyntax)
```
richardwerkman: This should be placed outside of the foreach. It checks the same node on every loop. It could be placed at the start of the method?

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -76,72 +216,66 @@ public SyntaxNode Mutate(SyntaxNode currentNode)
...
-                {
-                    var mutatedNode = Mutate(child);
-                    var originalNode = childCopy.GetCurrentNode(child);
-                    if (!mutatedNode.IsEquivalentTo(originalNode))
+
+                    if (actualMutationNode != null)
                     {
-                        childCopy = childCopy.ReplaceNode(originalNode, mutatedNode);
+                        var subNodeCopy = expressionSyntax.TrackNodes(actualMutationNode);
+                        var mutant = Mutate(actualMutationNode);
```
richardwerkman: These local variable names are very unclear. Mutant has a different meaning in the context of mutation testing. This seems off here.

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -62,12 +64,150 @@ public IEnumerable<Mutant> GetLatestMutantBatch()
...
+            var mutatedFor = forStatement.TrackNodes(forStatement.Condition, forStatement.Statement);
+            mutatedFor = mutatedFor.ReplaceNode(mutatedFor.GetCurrentNode(forStatement.Condition),
+                Mutate(forStatement.Condition));
+            mutatedFor = mutatedFor.ReplaceNode(mutatedFor.GetCurrentNode(forStatement.Statement),
+                Mutate(forStatement.Statement));
+            // and now we replace it
+            return trackedCopy.ReplaceNode(originalFor, mutatedFor);
+        }
+
+        private SyntaxNode MutateExpressions(SyntaxNode currentNode)
```
richardwerkman: This method is way to complicated. It might be good to create a new method called something like `GetSpecialTreatmentExpressionSyntax` where all special cases are gathered and returned. Just like `GetExpressionSyntax`. That would give this method some clarity I think.

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -201,10 +329,9 @@ private IEnumerable<Mutant> ApplyMutator(SyntaxNode syntaxNode, IMutator mutator
             }
         }
 
-        private T ApplyMutant<T>(T node, Mutant mutant) where T: SyntaxNode
+        private static T ApplyMutant<T>(T node, IReadOnlyMutant mutant) where T: SyntaxNode
```
richardwerkman: Why did you make this static?

src/Stryker.Core/Stryker.Core/Mutators/AssignmentStatementMutator.cs
```diff
@@ -23,13 +23,18 @@ public class AssignmentStatementMutator : MutatorBase<AssignmentExpressionSyntax
         public override IEnumerable<Mutation> ApplyMutations(AssignmentExpressionSyntax node)
         {
             var assignmentKind = node.Kind();
-            if (KindsToMutate.TryGetValue(assignmentKind, out var targetAssignementKind))
+            if (KindsToMutate.TryGetValue(assignmentKind, out var targetAssignmentKind))
             {
+                if (node.Kind() == SyntaxKind.AddAssignmentExpression &&(node.Left.IsAStringExpression() || node.Right.IsAStringExpression()))
```
richardwerkman: This line could use some formatting

src/Stryker.Core/Stryker.Core/Mutators/BinaryExpressionMutator.cs
```diff
@@ -33,11 +33,19 @@ public override IEnumerable<Mutation> ApplyMutations(BinaryExpressionSyntax node
         {
             if(_kindsToMutate.ContainsKey(node.Kind()))
             {
+                // skip string additions
+                
+                if (node.Kind() == SyntaxKind.AddExpression &&(node.Left.IsAStringExpression()||node.Right.IsAStringExpression()))
```
richardwerkman: This line could use some formatting

src/Stryker.Core/Stryker.Core/Mutators/BinaryExpressionMutator.cs
```diff
@@ -33,11 +33,19 @@ public override IEnumerable<Mutation> ApplyMutations(BinaryExpressionSyntax node
         {
             if(_kindsToMutate.ContainsKey(node.Kind()))
             {
+                // skip string additions
+
```
richardwerkman: Useless enter

src/Stryker.Core/Stryker.Core/Mutators/BinaryExpressionMutator.cs
```diff
@@ -33,11 +33,19 @@ public override IEnumerable<Mutation> ApplyMutations(BinaryExpressionSyntax node
...
+                {
+                    yield break;
+                }
+                
                 foreach(var mutationKind in _kindsToMutate[node.Kind()])
                 {
                     var replacementNode = SyntaxFactory.BinaryExpression(mutationKind, node.Left, node.Right);
                     // make sure the trivia stays in place for displaying
                     replacementNode = replacementNode.WithOperatorToken(replacementNode.OperatorToken.WithTriviaFrom(node.OperatorToken));
+
```
richardwerkman: Useless newline

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -76,72 +216,66 @@ public SyntaxNode Mutate(SyntaxNode currentNode)
...
+                return currentNodeCopy;
             }
+
+            return null;
         }
 
+
         private IEnumerable<Mutant> FindMutants(SyntaxNode current)
+        {
+            foreach (var mutant1 in FlatMutate(current)) yield return mutant1;
```
richardwerkman: This is not conform the microsoft guidelines. Please write the foreach full, and you can call this var `mutant` too. No need for the 1 I think.

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -62,12 +64,150 @@ public IEnumerable<Mutant> GetLatestMutantBatch()
...
+        }
+
+
+        private SyntaxNode MutateIfStatement(IfStatementSyntax ifStatement)
+        {
+            var mutatedIf = ifStatement.Else != null
+                ? ifStatement.TrackNodes(ifStatement.Condition, ifStatement.Statement, ifStatement.Else)
+                : ifStatement.TrackNodes(ifStatement.Condition, ifStatement.Statement);
+
+            if (!ifStatement.Condition.DescendantNodes().Any(x =>
```
richardwerkman: We could create an extensionMethod for this too like: `!ifStatement.Condition.ContainsDeclaration()`

src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_OUT.cs
```diff
@@ -0,0 +1,46 @@
+﻿namespace Test
```
Mobrockers: From what actual project is this code and do we have correct attribution /licensing to use it?

src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_IN.cs
```diff
@@ -0,0 +1,46 @@
+﻿namespace Test
```
Mobrockers: From what actual project is this code and do we have correct attribution /licensing to use it?

src/Stryker.Core/Stryker.Core.UnitTest/Stryker.Core.UnitTest.csproj
```diff
@@ -17,6 +21,21 @@
     <Compile Remove="Mutants\TestResources\Mutator_TwoMutationsInOneStatmentShouldBeMade.cs" />
   </ItemGroup>
   <ItemGroup>
+    <None Remove="Stryker.Core.UnitTest.v3.ncrunchproject" />
```
Mobrockers: What does this do

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
```diff
@@ -45,10 +45,32 @@ public RollbackProcess()
                 syntaxTreeMapping[diagnostic.Location.SourceTree].Add(diagnostic);
             }
 
-            // remove the broken mutations from the syntaxtrees
+            // remove the broken mutations from the syntax trees
             foreach(var syntaxTreeMap in syntaxTreeMapping.Where(x => x.Value.Any()))
             {
-                _logger.LogDebug("Rollbacking mutations from {0}", syntaxTreeMap.Key.FilePath);
+                if (devMode)
```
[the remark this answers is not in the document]
dupdob: Indeed, I should have done it.

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
dupdob: It is supposed to be a constant. I marked it as readonly. Why do you think it should be an instance field?

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
dupdob: new name:
forWithMutantIncrementors

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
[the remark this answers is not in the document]
dupdob: I thought about it, but then I realised that this method belongs to MutantOrchestrator as it relates to its inability to handle mutants.
To put it in another way: those node can be mutated, but they cannot be controlled by an If or a conditional expression.

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
dupdob: Done

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
dupdob: Well, I think it is more explicit as it expresses intent, vs copy which is broader. But I did not rename every other usage. So I think every other usage should be properly renamed, but I try to avoid endless naming discussions, so you tell me what you want

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
dupdob: Done. Refactoring was done with R#, did not bother to clean the mess :-)

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
dupdob: well.... I left the code as I found it, and I must say it did hurt my eyes too. Actually, there was several other constructions that were similar and I was able to fix them.
This one is harder than it looks to get right, and I had to stop refactoring the code at some stage.

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
dupdob: It is static as it is nothing more than syntactic sugar. Marking it as static makes it explicit: no side effects.

dupdob: Actually, it looks like it should be Mutant method.

src/Stryker.Core/Stryker.Core.UnitTest/Stryker.Core.UnitTest.csproj
dupdob: this is just noise. removed

src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_OUT.cs
dupdob: this is some snippets from NFluent. You want me to state it explicitly?

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
dupdob: I need to traverse through the ifs to locate the for that needs further mutation. I don't like it either, but I was not able to find any other way (including tracknodes)

src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_OUT.cs
Mobrockers: I thought it might be NFluent but I wasn't sure. If NFluent doesn't require attribution then this is fine. Since it explicitly states it is code from an actual project I just thought I would double check with you that we can use it.

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
dupdob: working on this

dupdob: got a nicer version. It needs further redesign, logic still feels a bit strange, but it is far much better than it originally was

src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_OUT.cs
dupdob: I kinda of seen this as fair use, as it was only a couple of methods. My intent was to make room to have tests that would work on 'known to have failed' code constructs.

src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_IN.cs
dupdob: NFluent, comment added

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
dupdob: Naming is hard 😉. Got rid of some intermediate steps, use _mutationCandidate_ instead of _actualMutationNode_. Sounds better to me

src/Stryker.Core/Stryker.Core.UnitTest/Mutants/TestResources/Mutator_FromActualProject_IN.cs
Mobrockers: Cool! That should avoid any confusion anyone might have :)

dupdob:
Please revert on open points (some naming and static related concerns).
The only open point that bugs me is the 'if traversing' I do to properly mutate _for_s. But I think the improvement provided overall deems this PR has good for merging.

I also understand that I probably got some methods wrong and probably twisted their original functions. My original intent was to fix a pressing issue (for me), but simple fix attempts made me understand I had to be more disruptive if I wanted to get some results

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
richardwerkman: Ah, I wanted it to be `const` but of course that doesn't work. We could add the readonly keyword?

richardwerkman: Now I see you already did, resolved!

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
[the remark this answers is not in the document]
dupdob: done done and done

dupdob:
hello @richardwerkman, sorry to ping you, but I am waiting for your feedback regarding two review comments you made on this PR. 
I am asking because I would like to focus on mutant coverage as a next feature

richardwerkman:
@dupdob Sorry I've been busy. Tonight I'll have time to look at the PR again and test it on a project.

dupdob:
Thanks for your answer

Envoyé de mon iPhone

richardwerkman:
Tested on another project and all mutations seemed correct and had their line numbers. All review points have been resolved. For me this is merge ready. @Mobrockers do you want to test this too or can this be merged?

Mobrockers:
@richardwerkman Nope, lgtm. Merging!

dupdob:
Thanks guys!!

Issue #388, opened 2019-03-05
Add coverage analysis to offer improved performance


Mobrockers:
This issue follows from #183

Generate mutant coverage (which mutants is actually executed by any test). This could provide useful metrics as of now.

Offer to skip test runs for non covered mutants (hence assumed as survivors)
==> Improve performance for projects with low test coverage.
Identify tests that do not cover any mutant, offer to skip them from each run.
==> Could improve performance for projects with many tests
Optimise executed tests according to mutant coverage (i.e. run only the tests that cover the 'current' mutant)
==> Should dramatically improve performance
Step 1 requires to able to access coverage results from Stryker
Step 2 looks simple
Step 3 requires to control which tests are run
Step 4 require analysis of coverage data to build strategy

Mobrockers:
From dupdop:

Some update: I have namedpipes running smoothly and stable, after a few fights.
I have reached maturity level 2: (my) Stryker now assumes non-covered mutants are survivors hence do not test them.

Code available here:
https://github.com/dupdob/stryker-net/tree/mutant_coverage

dupdob:
I have a branch ready, just waiting to finish with my current PR.

Mobrockers:
@dupdob Yes I've seen it. The named pipe seems like a good way to implement this. Step 3 can be achieved with the vstest integration.

To achieve step 4 I was thinking we could create a named pipe from the mutated assembly to a vstest data collector (out of proc as well) and then a named pipe from the data collector to stryker. The datacollector can execute for every single test being run and while in the middle we can receive which mutants have been hit for that testrun from the mutated assembly. Then we can send that information including the test that was run to stryker through the other named pipe. And then we can use vstest to run only those unit tests that are relevant to the mutant.

dupdob:
@Mobrockers 
thanks for the proposal, I have made an MVP using pipe communication between the assembly and Stryker, the objective reaching a proof of concept before considering improvement strategies.
Here is what I have achieved and learned so far. Spoiler alert, it is a bit **disappointing** (at least for netcore environment, it may be better with classical .Net)..
1. Overall, VsTest architecture is complex, brittle and prone to failure . I often end up having vstest.console.exe processes eating a core in some endless loop, logging nothing.
1. I have yet to achieve a complete run with NFluent due to above issues. At some point, Stryker runs out of available vstest runners.
1. **Performance are abysmal**. To get coverage info _per-test_, I launch each test in isolation. It is really slow. 
[...]

Issue #391, opened 2019-03-06
VsTest runner assumes the test assembly name


dupdob:
VsTest runner assumes the test assembly name is derived from the project name.
If the assembly name has been changed, the process fails.
related to #390

Mobrockers:
We can improve this once we have #371 merged because buildalyzer will probably be able to provide us with the necessary information about assembly names so we don't have to make assumptions anymore. It would be wasted effort to try to improve this before then.

dupdob:
good idea

PR #393, opened 2019-03-07
feat(Mutant coverage): Analyze mutant coverage to improve performance by filtering tests

dupdob:
### this PR is not to be merged for now, it is WIP
Implement mutant coverage, i.e. have the list of tests that goes through each mutant.
see issue #388 for motivation and usage.

Notes:
- control mechanism for mutant control as been revised (method call instead of a property). A method call is imperative to track mutant usage.


dupdob:
I have issues with one test with namepipe. Works in local, fails on Azure (aren't they the best ones?).
Need to dig into this.

dupdob:
### Current status
Integrated **basic** mutant coverage (i.e. which mutant is actually executed during tests) with VsTest.
Plus, this version skip testing non covered mutants.
Now I will work on detail coverage : which mutants are executed during each test.

Side note: The stability issue with named pipe tests was due to race conditions with the assertion logic. It gave me the idea for a useful helper method.

dupdob:
### Update
- added coverage capture with VsTest. Results are unstable on integration test projects: actual result is 5 mutants are covered by tests (out of 15), but sometimes I get only 3. May be related to a race condition (data is published when the appdomain is tear down).
- tests are now interrupted as soon as a test failure is received (with VsTest), should provide total time reduction by at least 2x up to 20-30x or even more. But no run made against NFluent, waiting for issue #391 to be solved

dupdob:
VsTest now capture and reports covered mutants **per test**!
Performance needs to improve.

dupdob:
Some update on my end: I have been busy fixing an issue on NFluent and preparing a release last week end, but I have been able to implement the optimisation logic: running only the appropriate tests for each mutant. 
On the integration project, it does not bring any visible difference (too small). 
I tried it on NFluent (renamed the assembly to go past issue #391): it seems to be failing at various stage. Performance turns out to be a problem as running each test in isolation is really slow.

Further work needed

dupdob:
I just pushed past days' work. It implements all 4 steps of coverage strategies.
Still issues to be ironed out, and performance kinda sucks, but this is functional.
Optimisation features are now controlled by a flag field in InitialisationProcess

dupdob:
Now I know why the coverage logic fails to send back its results: as explained in this [comment](https://github.com/Microsoft/vstest/issues/1900#issuecomment-458004993No), VsTest kills the testing process if it takes more than 100ms to shut down.

Mobrockers:
Hey @dupdob, how's it going? :) 

Would you mind if we also take a look at this some time?

dupdob:
Of course, I suggest you read my status report and revert to me regarding how to move forward


---
# summary
well, I am still working on NFluent right now and your ping makes me realise that I have been stuck for a while on this PR.
here is my current status:
I am able to reliably identify mutants that are not covered by any test, which allows to skip them during the test phase, saving some time (depending of the test coverage)

I am still unable to RELIABLY identify the mutants covered by each tests.

The main obstacles are:
- reliability issues with (my implementation of) pipes. despite intensive debugging, some lockup still occurs
- asynchronous behaviour in VsTest

## designs
I have went through three design phases/strategy. I'll share them with you:

0. prerequisite: review the active mutant selection component to use a method call (IsActive(int id)) instead of a property. It allows the code to capture which mutants are covered

1. **capture the list of tested mutants**. This is done thanks to (0) and the list of covered mutants are sent back to Stryker using a named pipe. The transmission is triggered by the appdomain shut down event. But there is some stability issue (sometimes result is not transmitted, like 3-5% of the time), probably due to timing issues (appdomain shutdown event processing is limited to two seconds and vstests uses it as well.

2. **capture the list of tested mutants on a per test basis**. I use previous design and runs each test in isolation, which provides me with the list of tested mutants for each tests. I encounter the same stability issue, but I can mitigate it by retrying the run a couple of times. At the end, I get coverage date for 99.8% of the tests (i.e. only 5-8 failures on NFluent test base), **but this approach is almost as slow as the non optimised case**, each test run taking a couple of seconds, it takes roughly 30-40 minutes to get coverage result for NFluent.

3. **capture the list of tested mutants using an inproc data collector**. This time, the tests are run normally, but I use a custom inproc data collector to get coverage date after each test is run. Performance looks very good, but this is very unstable and not usable with NFluent. The first issue was that you need to run tests on a sequential basis (at least for NUnit), otherwise you get asynchronous notification for test runs, meaning coverage data leak across tests. Plus I have severe deadlocks with pipes, I fixed two issues in my code but I am not able to spot any other, I hope the issue lies in my implementation, but there is also a risk the issue my come from async IO being not very friendly with vstest threading model.

After spending two weeks trying to fix my issue, I decided I needed to refresh my mind and work on something simpler

## possible next steps
If you have a different design/strategy in mind, try it, by all means. Otherwise, here is a proposal for you:
1. I can revert my code to design number 2 and push a PR for it. It will allow to release the feature as beta/experimental. Note that those optimisations could be controlled though Stryker options. My code use configuration flags, but those are not exposed as option yet.
1. You can also activate only the basic coverage strategy that should improve performance for many project, plus provide interesting feedback to user (this code is not covered by any test)
1. I can push my latest work is anybody is willing to give a try and find the issue(s) that block me


As usual, I will respond to questions/comments ASAP

keep up the good work

Mobrockers:
@dupdob If you could provide us with the commit id's for each step (1, 2) and push your latest changes, then I'll take a look at it and depending on my experience I'll decide with richard which route we take. Preferably I would continue working on your latest work because that is the route we discussed internally with coworkers as well :) I'll just create new branches from the commits for each step and release based on that. Sound good?

dupdob:
helloI am afraid I can no longer point to the specific commits. 
I restarted effort from scratch along the way.
Plus I fixed bug in the pipe implementations that would have to be back ported.
I am working on a PR that will provide those features. I will push something tomorrow

PR #416, opened 2019-03-20
feat(exclusions): Warn user on all mutations excluded

kkolodziejczak:
Hi, 
I'm currently starting a new project and I disabled string mutations, It happened that those were only mutations in my project. So all mutations were skipped. I was surprised when Stryker exited with code 1, instead of 0, even that I had Threshold Break set to 0. 

I think that when there are no tests to run exit code should be 0 not 1. The issue, that I found is passing mutation score as nullable decimal `decimal?` to the `StrykerRunResult` when there are no mutations to run.
```csharp
public bool IsScoreAboveThresholdBreak()
{
    // Check if the mutation score is not below the threshold break
    //          null     >=   0  this is not true.
    return MutationScore >= _options.Thresholds.Break;
}
```
I added if statement that checks if `MutationScore` is `null`, and checks if Threshold.Break is set to 0.

```csharp
public bool IsScoreAboveThresholdBreak()
{
    if (MutationScore == null)
    {
        return _options.Thresholds.Break == 0;
    }

    // Check if the mutation score is not below the threshold break
    return MutationScore >= _options.Thresholds.Break;
}
```

### DIFF
--- a/src/Stryker.CLI/Stryker.CLI.UnitTest/StrykerCLITests.cs
+++ b/src/Stryker.CLI/Stryker.CLI.UnitTest/StrykerCLITests.cs
@@ -302,8 +302,8 @@ namespace Stryker.CLI.UnitTest
             int result = target.Run(new string[] { });
 
             mock.Verify();
-            Assert.Equal(1, target.ExitCode);
-            Assert.Equal(1, result);
+            target.ExitCode.ShouldBe(1);
+            result.ShouldBe(1);
         }
 
         [Fact]
@@ -318,8 +318,8 @@ namespace Stryker.CLI.UnitTest
             int result = target.Run(new string[] { });
 
             mock.Verify();
-            Assert.Equal(0, target.ExitCode);
-            Assert.Equal(0, result);
+            target.ExitCode.ShouldBe(0);
+            result.ShouldBe(0);
         }
 
         [Fact]
@@ -334,8 +334,8 @@ namespace Stryker.CLI.UnitTest
             int result = target.Run(new string[] { });
 
             mock.Verify();
-            Assert.Equal(0, target.ExitCode);
-            Assert.Equal(0, result);
+            target.ExitCode.ShouldBe(0);
+            result.ShouldBe(0);
         }
 
         [Fact]
@@ -351,8 +351,8 @@ namespace Stryker.CLI.UnitTest
             int result = target.Run(new string[] { });
 
             mock.Verify();
-            Assert.Equal(0, target.ExitCode);
-            Assert.Equal(0, result);
+            target.ExitCode.ShouldBe(0);
+            result.ShouldBe(0);
         }
 
         [Fact]
--- a/src/Stryker.CLI/Stryker.CLI.UnitTest/StrykerCLITests.cs
+++ b/src/Stryker.CLI/Stryker.CLI.UnitTest/StrykerCLITests.cs
@@ -307,6 +307,22 @@ namespace Stryker.CLI.UnitTest
         }
 
         [Fact]
+        public void StrykerCLI_OnMutationScoreEqualToNullAndThresholdBreakEqualTo0_ShouldReturnExitCode0()
+        {
+            var mock = new Mock<IStrykerRunner>(MockBehavior.Strict);
+            StrykerOptions options = new StrykerOptions(thresholdBreak: 0);
+            StrykerRunResult strykerRunResult = new StrykerRunResult(options, null);
+            mock.Setup(x => x.RunMutationTest(It.IsAny<StrykerOptions>())).Returns(strykerRunResult).Verifiable();
+
+            var target = new StrykerCLI(mock.Object);
+            int result = target.Run(new string[] { });
+
+            mock.Verify();
+            Assert.Equal(0, target.ExitCode);
+            Assert.Equal(0, result);
+        }
+
+        [Fact]
         public void StrykerCLI_OnMutationScoreAboveThresholdBreak_ShouldReturnExitCode0()
         {
             var mock = new Mock<IStrykerRunner>(MockBehavior.Strict);
--- a/src/Stryker.CLI/Stryker.CLI.UnitTest/StrykerCLITests.cs
+++ b/src/Stryker.CLI/Stryker.CLI.UnitTest/StrykerCLITests.cs
@@ -323,6 +323,22 @@ namespace Stryker.CLI.UnitTest
         }
 
         [Fact]
+        public void StrykerCLI_OnMutationScoreEqualToNullAndThresholdBreakAbove0_ShouldReturnExitCode0()
+        {
+            var mock = new Mock<IStrykerRunner>(MockBehavior.Strict);
+            StrykerOptions options = new StrykerOptions(thresholdBreak: 40);
+            StrykerRunResult strykerRunResult = new StrykerRunResult(options, null);
+            mock.Setup(x => x.RunMutationTest(It.IsAny<StrykerOptions>())).Returns(strykerRunResult).Verifiable();
+
+            var target = new StrykerCLI(mock.Object);
+            int result = target.Run(new string[] { });
+
+            mock.Verify();
+            Assert.Equal(0, target.ExitCode);
+            Assert.Equal(0, result);
+        }
+
+        [Fact]
         public void StrykerCLI_OnMutationScoreAboveThresholdBreak_ShouldReturnExitCode0()
         {
             var mock = new Mock<IStrykerRunner>(MockBehavior.Strict);
--- a/src/Stryker.Core/Stryker.Core.UnitTest/MutationTest/MutationTestProcessTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/MutationTest/MutationTestProcessTests.cs
@@ -13,6 +13,7 @@ using System.Collections.ObjectModel;
 using System.IO;
 using System.IO.Abstractions.TestingHelpers;
 using System.Reflection;
+using Shouldly;
 using Xunit;
 
 namespace Stryker.Core.UnitTest.MutationTest
@@ -336,7 +337,7 @@ namespace Stryker.Core.UnitTest.MutationTest
             reporterMock.Verify(x => x.OnStartMutantTestRun(It.Is<IList<Mutant>>(y => y.Count == 1)), Times.Never);
             reporterMock.Verify(x => x.OnMutantTested(mutant), Times.Never);
             reporterMock.Verify(x => x.OnAllMutantsTested(It.IsAny<ProjectComponent>()), Times.Never);
-            Assert.Null(testResult.MutationScore);
+            testResult.MutationScore.ShouldBeNull();
         }
 
         [Fact]
@@ -385,7 +386,7 @@ namespace Stryker.Core.UnitTest.MutationTest
             reporterMock.Verify(x => x.OnStartMutantTestRun(It.Is<IList<Mutant>>(y => y.Count == 1)), Times.Never);
             reporterMock.Verify(x => x.OnMutantTested(It.IsAny<Mutant>()), Times.Never);
             reporterMock.Verify(x => x.OnAllMutantsTested(It.IsAny<ProjectComponent>()), Times.Never);
-            Assert.Null(testResult.MutationScore);
+            testResult.MutationScore.ShouldBeNull();
         }
 
         [Fact]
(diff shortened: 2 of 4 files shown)


src/Stryker.Core/Stryker.Core/StrykerRunResult.cs
```diff
@@ -15,6 +15,11 @@ public StrykerRunResult(StrykerOptions options, decimal? mutationScore)
 
         public bool IsScoreAboveThresholdBreak()
         {
+            if (MutationScore == null)
+            {
+                return _options.Thresholds.Break == 0;
```
richardwerkman: Why do you validate the breaking treshold here? If no mutations are made and my treshold > 0 I would get an exit code 1 right? Is this expected behaviour? 

I guess when no mutations are made you don't have a mutation score. Thus your score is neither lower or higher than the breaking treshold. I would expect a 0 exit code. @Mobrockers what do you think about this matter?

kkolodziejczak: > Why do you validate the breaking treshold here? If no mutations are made and my treshold > 0 I would get an exit code 1 right? Is this expected behaviour?

Oh, I didn't notice this scenario. That's true in this case it would return exit code 1. 
In that case 
```
if (MutationScore == null)
{
    return true;
}
```
should do the job. I think that if no tests were run, then we should have success message.

Mobrockers:
IMO if there are no mutations to be tested we should simply write out the report with skipped mutations (if any) and exit with a warning that we did not find any mutations to test. We should definitely not break but we should tell the user that there was nothing for us to test so we have no actual result.

Until we warn the user that we did in fact not actually test anything and can therefore not determine anything about the test quality, I believe that we should break on no mutations tested. We cannot end on a 'good' result while also not notifying the user that something is up.

richardwerkman:
@Mobrockers I can't agree with that. We have seen that test runners also exit with exit code 0 on no test cases found. It is to users themselves to see the stryker output and see that no mutations were made. That does not mean stryker failed in some way or the breaking score was reached.

src/Stryker.Core/Stryker.Core/StrykerRunResult.cs
richardwerkman: Yes, this would be better for me! If you could update the PR it can be merged.

Mobrockers:
@richardwerkman That other testframeworks show this behavior does not necessarily mean that it is the right one. In any case, I am fine with exiting with 0 as long as we tell the user clearly that we did not find any mutations to test. We should not leave it up to the user to decide from reading the stryker report that there were no mutations. We are capable of seeing this behavior and notifying users explicitly.

simondel:
For Stryker (js+ts) we don't run mutation testing when there are no mutants to run. Instead, we just exit and log an info message with the number of mutants that were excluded due to the user's settings:
https://github.com/stryker-mutator/stryker/blob/24467e11dcbf68ec2e6fdbb4fadb0c70e71206c3/packages/core/src/mutants/MutatorFacade.ts#L41-L53

kkolodziejczak:
Test output looks like this(at the bottom), the information about skipped mutations and mutations to run is displayed. The process itself is not stopped when there are no mutations to test.

I think that the solution would be to see if there is something to test after mutation process is done, and then exit if there are no mutations to be tested. 

The `Mutate(...)` method from `IMutationTestProcess` could return number of mutations created, instead of void, and then tests would be executed if this number is not equal to 0.
```
   _____ _              _               _   _ ______ _______
  / ____| |            | |             | \ | |  ____|__   __|
 | (___ | |_ _ __ _   _| | _____ _ __  |  \| | |__     | |
  \___ \| __| '__| | | | |/ / _ \ '__| | . ` |  __|    | |
  ____) | |_| |  | |_| |   <  __/ |    | |\  | |____   | |
 |_____/ \__|_|   \__, |_|\_\___|_| (_)|_| \_|______|  |_|
                   __/ |
                  |___/
Version 0.9.0 (beta)

[15:38:30 INF] Using C:\...\Xamarin.BetterNavigation.UnitTests.csproj as project file
[15:38:30 INF] Starting initial build
[15:38:32 INF] Initial build successful
[15:38:33 INF] Using testrunner DotnetTest
[15:38:33 INF] Initial testrun started
[15:38:35 INF] Total number of tests found in initial test run: 0
[15:38:35 INF] Initial testrun successful in 2104 ms
[15:38:35 INF] Using 33156 ms as testrun timeout
[15:38:36 INF] 27 mutants created
[15:38:38 INF] 27 mutants got status Skipped
[15:38:38 INF] 0 mutants ready for test

0 mutants have been created. Each mutant will now be tested, this could take a while.

Tests progress | ▒▒▒▒▒▒▒▒▒▒ | 0 / 0 | 100 % | NA |
Killed : 0
Survived: 0
Timeout : 0


- \Xamarin.BetterNavigation.Forms [0/0 (- %)]
--- IPageLocator.cs [0/0 (- %)]
--- NavigationService.cs [0/0 (- %)]
[15:38:38 INF] Time Elapsed 00:00:07.3945299
[15:38:38 ERR] Final mutation score:  under breaking threshold value 0.
Setting exit code to 1 (failure).
Improve the mutation score or set the `threshold-break` value lower to prevent this error in the future.
```

richardwerkman:
@kkolodziejczak I agree with the exit strategy. Could you add the logic for exiting on 0 mutants?

I think we have to follow Stryker (js) here:
- On 0 mutants exit and log a warning:
    - "It's a mutant-free world, nothing to test."
- On > 0 mutants but all with status "skip" log a warning
    - "It looks like all mutants were excluded, try a re-run with less exclusion."

I think exit code 0 is best here. By logging a warning we show something is off, but stryker did not fail.

kkolodziejczak:
I made something, I did push it to the other branch though, I'm not 100% satisfied with the result. 
I needed to modify source code a little to make it work, feedback would be awesome.
Where would be a good place to write some tests for that? 

https://github.com/kkolodziejczak/stryker-net/commit/87579a72954f8f24f1284cffc95219b85bf01cb1

richardwerkman:
I've commented on your commit, I hope you can see the comments 😄. I think that if we move the logic into the Test method of MutationTestProcess we can make things more slick. Also the MutationTestProcess would be a good place to check if no reporter is called in the case of 0 mutants 👍

src/Stryker.Core/Stryker.Core.UnitTest/MutationTest/MutationTestProcessTests.cs
```diff
@@ -288,5 +288,155 @@ public void MutationTestProcess_ShouldCallExecutorForEveryMutant()
...
+                reporterMock.Object,
+                executorMock.Object);
+
+            var testResult = target.Test(options);
+
+            executorMock.Verify(x => x.Test(mutant), Times.Never);
+            reporterMock.Verify(x => x.OnStartMutantTestRun(It.Is<IList<Mutant>>(y => y.Count == 1)), Times.Never);
+            reporterMock.Verify(x => x.OnMutantTested(mutant), Times.Never);
+            reporterMock.Verify(x => x.OnAllMutantsTested(It.IsAny<ProjectComponent>()), Times.Never);
+            Assert.Null(testResult.MutationScore);
```
richardwerkman: We try to use shouldly for assertions. This would look like:
```suggestion
            testResult.MutationScore.ShouldBeNull();
```

src/Stryker.CLI/Stryker.CLI.UnitTest/StrykerCLITests.cs
```diff
@@ -306,6 +306,38 @@ public void StrykerCLI_OnMutationScoreBelowThresholdBreak_ShouldReturnExitCode1(
...
+            var mock = new Mock<IStrykerRunner>(MockBehavior.Strict);
+            StrykerOptions options = new StrykerOptions(thresholdBreak: 0);
+            StrykerRunResult strykerRunResult = new StrykerRunResult(options, null);
+            mock.Setup(x => x.RunMutationTest(It.IsAny<StrykerOptions>())).Returns(strykerRunResult).Verifiable();
+
+            var target = new StrykerCLI(mock.Object);
+            int result = target.Run(new string[] { });
+
+            mock.Verify();
+            Assert.Equal(0, target.ExitCode);
```
richardwerkman: It would be nice to use shouldly here too. But I now see that we neither did above here. Feel free to edit or keep it the same :)

src/Stryker.Core/Stryker.Core.UnitTest/MutationTest/MutationTestProcessTests.cs
```diff
@@ -288,5 +288,155 @@ public void MutationTestProcess_ShouldCallExecutorForEveryMutant()
...
+                            Name = "SomeFile.cs",
+                            Mutants = new Collection<Mutant>() { }
+                        }
+                    }
+                    },
+                    ProjectUnderTestAssemblyName = "ExampleProject",
+                    ProjectUnderTestPath = Path.Combine(_filesystemRoot, "ExampleProject"),
+                    TargetFramework = "netcoreapp2.0",
+                },
+                AssemblyReferences = new ReferenceProvider().GetReferencedAssemblies()
```
richardwerkman: This line uses reflection to get some assembly references. It would be nice for performance to reuse its value in each test

src/Stryker.Core/Stryker.Core.UnitTest/MutationTest/MutationTestProcessTests.cs
```diff
@@ -288,5 +288,155 @@ public void MutationTestProcess_ShouldCallExecutorForEveryMutant()
...
+            var options = new StrykerOptions(fileSystem: new MockFileSystem(), basePath: basePath);
+
+            var target = new MutationTestProcess(input,
+                reporterMock.Object,
+                executorMock.Object);
+
+            var testResult = target.Test(options);
+
+            executorMock.Verify(x => x.Test(It.IsAny<Mutant>()), Times.Never);
+            reporterMock.Verify(x => x.OnStartMutantTestRun(It.Is<IList<Mutant>>(y => y.Count == 1)), Times.Never);
```
richardwerkman: I think using Any list of mutations could be better for future changes. 
```suggestion
            reporterMock.Verify(x => x.OnStartMutantTestRun(It.IsAny<IList<Mutant>>(), Times.Never);
```

If the code somehow calls OnStartMutantTestRun with an empty list this test would not fail.

src/Stryker.Core/Stryker.Core.UnitTest/MutationTest/MutationTestProcessTests.cs
```diff
@@ -288,5 +288,155 @@ public void MutationTestProcess_ShouldCallExecutorForEveryMutant()
...
+            var options = new StrykerOptions(fileSystem: new MockFileSystem(), basePath: basePath);
+
+            var target = new MutationTestProcess(input,
+                reporterMock.Object,
+                executorMock.Object);
+
+            var testResult = target.Test(options);
+
+            executorMock.Verify(x => x.Test(It.IsAny<Mutant>()), Times.Never);
+            reporterMock.Verify(x => x.OnStartMutantTestRun(It.Is<IList<Mutant>>(y => y.Count == 2)), Times.Never);
```
richardwerkman: Same here

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
```diff
@@ -138,6 +138,18 @@ public void Mutate(StrykerOptions options)
         public StrykerRunResult Test(StrykerOptions options)
         {
             var mutantsNotRun = _input.ProjectInfo.ProjectContents.Mutants.Where(x => x.ResultStatus == MutantStatus.NotRun).ToList();
+            if (!mutantsNotRun.Any())
+            {
+                if (_input.ProjectInfo.ProjectContents.Mutants.Any(x => x.ResultStatus == MutantStatus.Skipped))
```
richardwerkman: I think we can just use Any here? If there are > 0 mutants but no mutants to run, it means the other mutants were excluded.

```suggestion
                if (_input.ProjectInfo.ProjectContents.Mutants.Any()
```

kkolodziejczak: Well other mutations were excluded, that's correct but what if the other mutations did not compile?
Then we would say that all mutations were excluded, is this expected behavior? Since user is not able to re-run with less exclusion, compile error mutations.

Current logic(any => skipped): 

|mutations to run | skipped | compile error| message|
|-------------------|----------|----------------|----------|
|0|0|0|It's a mutant-free world, nothing to test.|
|0|1|0|It looks like all mutants were excluded, try a re-run with less exclusion.|
|0|0|1|It's a mutant-free world, nothing to test.|

Current logic(any): 

|mutations to run | skipped | compile error| message|
|-------------------|----------|----------------|----------|
|0|0|0|It's a mutant-free world, nothing to test.|
|0|1|0|It looks like all mutants were excluded, try a re-run with less exclusion.|
|0|0|1|It looks like all mutants were excluded, try a re-run with less exclusion.|

PR #423, opened 2019-03-23
feat(compiling): Add support for properties in shared projects

### LINKED ISSUE #317: Missing shared project if using csproj imports with properties in file name
Stryker throws `System.IO.FileNotFoundException: Missing shared project $(BuildDirectory)Foo.targets` if some of the `Imports` have properties in them (it is likely those properties are set in some of the MSBuild explicit imports as part of more complex build).

## Steps to reproduce

Add an `Import` with a `Project` containing a property (like the following snippet) to the `csproj` file of the project under test.

```xml
<Import Project="$(BuildDirectory)Foo.targets" />
```

## Related issues

* #235 
* #230

shadow-cs:
Adds support for properties in imported project.

Integration tests were modified in order to cover more complex project setups.

Fixes #317

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
+++ b/src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
@@ -163,7 +163,7 @@ namespace Stryker.Core.Initialisation
             {
                 foreach (var sharedProject in new ProjectFileReader().FindSharedProjects(xDocument))
                 {
-                    var sharedProjectName = ReplaceMsbuildProperties(sharedProject, projectAnalyzerResult.Properties);
+                    var sharedProjectName = ReplaceMsbuildProperties(sharedProject, projectAnalyzerResult);
 
                     if (!_fileSystem.File.Exists(_fileSystem.Path.Combine(projectDirectory, sharedProjectName)))
                     {
@@ -178,18 +178,23 @@ namespace Stryker.Core.Initialisation
             return folders;
         }
 
-        private static string ReplaceMsbuildProperties(string value, IReadOnlyDictionary<string, string> properties)
+        private static string ReplaceMsbuildProperties(string projectReference, ProjectAnalyzerResult projectAnalyzerResult)
         {
             var propertyRegex = new Regex(@"\$\(([a-zA-Z_][a-zA-Z0-9_\-.]*)\)");
+            var properties = projectAnalyzerResult.Properties;
 
-            try
-            {
-                return propertyRegex.Replace(value, m => properties[m.Groups[1].Value]);
-            }
-            catch (KeyNotFoundException e)
-            {
-                throw new StrykerInputException($"Cannot replace all properties in {value}", e);
-            }
+            return propertyRegex.Replace(projectReference,
+                m =>
+                {
+                    var property = m.Groups[1].Value;
+                    if (properties.TryGetValue(property, out var propertyValue))
+                    {
+                        return propertyValue;
+                    }
+
+                    var message = $"Missing MSBuild property ({property}) in project reference ({projectReference}). Please check your project file ({projectAnalyzerResult.ProjectFilePath}) and try again.";
+                    throw new StrykerInputException(message);
+                });
         }
     }
 }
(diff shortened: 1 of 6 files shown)


shadow-cs:
I didn't add support for calling function or using items as it would require more processing and is a job that would possibly needed to be done by buildalyzer. I don't think anyone is using that though as it would make the build difficult to maintain. Using properties on the other hand is very common (I actually set-up my build by learning from Moq ;-) ).

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -161,18 +166,31 @@ private IEnumerable<string> ExtractProjectFolders(string projectFilePath, bool f
...
+
+        private static string ReplaceMsbuildProperties(string value, IReadOnlyDictionary<string, string> properties)
+        {
+            try
+            {
+                return PropertyRegex.Replace(value, m => properties[m.Groups[1].Value]);
+            }
+            catch (KeyNotFoundException e)
+            {
+                throw new FileNotFoundException($"Cannot replace all properties in {value}", e);
```
Mobrockers: I think this should rather be a stryker input exception, as something is wrong in the input provided to stryker to run the test

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -24,6 +25,9 @@ public interface IInputFileResolver
     /// </summary>
     public class InputFileResolver : IInputFileResolver
     {
+        private static readonly Regex PropertyRegex = new Regex(@"\$\(([a-zA-Z_][a-zA-Z0-9_\-.]*)\)");
```
Mobrockers: Since this and the EmptyProperties member are only used in ReplaceMsbuildProperties I would rather have them be local to that function

shadow-cs:
Do you prefer fixups or separate commits during review?

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
shadow-cs: Understood I jus wanted to prevent parsing of the expression for every call of the method.

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
shadow-cs: Will do, I wasn't sure which exception to use, in the end I didn't want the change the API surface so I ended up using whichever was thrown before the change...

Mobrockers:
I would prefer new commits over amends. The contents and messages for the commits don't matter much though, as we squash merge to master.

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
Mobrockers: That is understandable, I don't believe we have properly documented the exception strategy.

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
shadow-cs: Done, I removed the `EmptyProperties` as it was no longer needed the tests were fine after adding the `Properties` in the test cases. Buildalyzer will never put `null` there (right?) so we should be safe.

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
shadow-cs: Updated to `StrykerInputException`, I took the liberty and added `innerException` to `StrykerInputException` ctor, if you don't want to keep track of the inner exception I'll revert it back.

Mobrockers:
StrykerInput exception should not have an inner exception I don't think. It
should have a clean error message that goes straight to the user.
@richardwerkman thoughts?

richardwerkman:
Yes, the idea behind StrykerInputException is that we can let users know what they did wrong without showing a stacktrace. Thus an innerexception is not needed. 

For this case a message like `Missing MSBuild property {propertyname} in projectReference {projectreference}. Please check your projectFile and try again.` would do the trick.

shadow-cs:
> For this case a message like `Missing MSBuild property {propertyname} in projectReference {projectreference}. Please check your projectFile and try again.` would do the trick.

@richardwerkman  on it

shadow-cs:
Hope the message is OK with the detailed info in braces @richardwerkman . If not, please suggest the changes.

Mobrockers:
Thanks!

Issue #425, opened 2019-03-25
Improve message of exception when directory contains more than one csproj


robertlyson:
Hi,

Just encountered an error when a directory where I try to run Stryker contains more than one csproj:
```
   _____ _              _               _   _ ______ _______
  / ____| |            | |             | \ | |  ____|__   __|
 | (___ | |_ _ __ _   _| | _____ _ __  |  \| | |__     | |
  \___ \| __| '__| | | | |/ / _ \ '__| | . ` |  __|    | |
  ____) | |_| |  | |_| |   <  __/ |    | |\  | |____   | |
 |_____/ \__|_|   \__, |_|\_\___|_| (_)|_| \_|______|  |_|
                   __/ |
                  |___/


Version 0.9.0 (beta)

[20:58:45 INF] Time Elapsed 00:00:00.1017350
Stryker.NET failed to mutate your project. For more information see the logs below:

Expected exactly one .csproj file, found more than one. Please fix your project contents
```

Would be great if the exception message could contain details about csprojs found. That could help with a faster investigation with what is wrong.

Mobrockers:
Fixed #426

PR #426, opened 2019-03-25
feat(input file resolver): directory contains more than one csproj

robertlyson:
Hi, just thought maybe this error formatting 
```
   _____ _              _               _   _ ______ _______
  / ____| |            | |             | \ | |  ____|__   __|
 | (___ | |_ _ __ _   _| | _____ _ __  |  \| | |__     | |
  \___ \| __| '__| | | | |/ / _ \ '__| | . ` |  __|    | |
  ____) | |_| |  | |_| |   <  __/ |    | |\  | |____   | |
 |_____/ \__|_|   \__, |_|\_\___|_| (_)|_| \_|______|  |_|
                   __/ |
                  |___/


Version 0.9.0 (beta)

[21:19:14 INF] Time Elapsed 00:00:00.1008880
Stryker.NET failed to mutate your project. For more information see the logs below:

Expected exactly one .csproj file, found more than one:
/Users/robertlyson/projects/TestProject2/TestProject2/TestProject2.csproj
/Users/robertlyson/projects/TestProject2/TestProject2/published/TestProject2.csproj

Please fix your project contents
```
would be a good solution for https://github.com/stryker-mutator/stryker-net/issues/425.

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
+++ b/src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.IO.Abstractions;
 using System.Linq;
+using System.Text;
 using System.Xml.Linq;
 
 namespace Stryker.Core.Initialisation
@@ -141,7 +142,13 @@ namespace Stryker.Core.Initialisation
             _logger.LogTrace("Scanned the current directory for *.csproj files: found {0}", projectFiles);
             if (projectFiles.Count() > 1)
             {
-                throw new StrykerInputException("Expected exactly one .csproj file, found more than one. Please fix your project contents");
+                var sb = new StringBuilder();
+                sb.AppendLine("Expected exactly one .csproj file, found more than one:");
+                foreach (var file in projectFiles)
+                    sb.AppendLine(file);
+                sb.AppendLine();
+                sb.AppendLine("Please fix your project contents");
+                throw new StrykerInputException(sb.ToString());
             }
             else if (!projectFiles.Any())
             {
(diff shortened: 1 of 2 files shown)


richardwerkman:
Hi, thanks for yet another contribution! This is actually not behaviour I would expect. I thought the check for project files was only for the top level directory. As I now see we scan for project files in sub-directories too. 

Maybe it would be good to investigate first if we still need to do that. In your example, I guess you want to run on the project file in the top level directory. And Stryker can run fine with that extra csproj in a sub-directory I think. I wrote the check to validate that only 1 project was found to run agains. In your case there is only 1 project in the directory so I don't think Stryker should stop/crash.

Could you investigate this a little more?

robertlyson:
@richardwerkman that make sense, will have a look further. Do you think we should leave the improved message in place just in case someone has two csprojs in the same directory?

richardwerkman:
If it is indeed true that we only have to scan the top directory a message like:
```
Expected exactly one .csproj file in the project root directory, found more than one:
/Users/robertlyson/projects/TestProject2/TestProject2/TestProject1.csproj
/Users/robertlyson/projects/TestProject2/TestProject2/TestProject2.csproj

Please fix your project contents.
```

In your specific case you should see no error and Stryker should just continue :+1:

robertlyson:
I've pushed change and also added one test cases for a scenario where the current directory contains another directory with csproj inside.

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -137,11 +138,17 @@ private void MarkInputFilesAsExcluded(FolderComposite root, List<string> pathsTo
...
         {
-            var projectFiles = _fileSystem.Directory.GetFiles(currentDirectory, "*.csproj", SearchOption.AllDirectories);
+            var projectFiles = _fileSystem.Directory.GetFiles(currentDirectory, "*.csproj");
             _logger.LogTrace("Scanned the current directory for *.csproj files: found {0}", projectFiles);
             if (projectFiles.Count() > 1)
             {
-                throw new StrykerInputException("Expected exactly one .csproj file, found more than one. Please fix your project contents");
+                var sb = new StringBuilder();
+                sb.AppendLine("Expected exactly one .csproj file, found more than one:");
+                foreach (var file in projectFiles)
```
Mobrockers: Please add the braces for this foreach and i'll merge the pr

robertlyson: Done, thanks for your comment.

PR #428, opened 2019-03-26
feat(compiling mutations): Add support for signed assemblies

decipherer:
The mutated assembly is signed with the same key as the original assembly, this is what the unittest project expects.

Besides the added support for signed assemblies it contains a fix for certain projects that buildalyzer is not able to analyze well without passing the solution file to ProjectFileReader.AnalyzeProject. In our case the references of the project file were not found by buildalyzer without passing the solution file. With this fix, even though the references are found, buildalyzer still says analyzing fails, I've added a warning when that happens.


src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
```diff
@@ -27,22 +27,13 @@ public ProjectFileReader()
...
-            {
-                _logger.LogDebug("Analyzing solution file {0}", solutionFilePath);
-                manager = new AnalyzerManager(solutionFilePath);
-            }
-            analyzerResult = manager.GetProject(projectFilePath).Build().First();
-            _logger.LogDebug("Analyzing project file {0}", projectFilePath);
+            _logger.LogDebug("Analyzing solution file {0}, project file {1}", solutionFilePath, projectFilePath);
+            var manager = new AnalyzerManager(solutionFilePath);
+            var analyzerResult = manager.GetProject(projectFilePath).Build().First();
+            if (!analyzerResult.Succeeded)
```
Mobrockers: We don't use shorthand if statements

src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs
```diff
@@ -45,13 +45,6 @@ public MutationTestInput Initialize(StrykerOptions options)
             // initial build
             _initialBuildProcess.InitialBuild(projectInfo.FullFramework, options.BasePath, options.SolutionPath, Path.GetFileName(projectInfo.TestProjectAnalyzerResult.ProjectFilePath));
 
-            // If project is full framework, we need to reload projectinfo and rebuild
-            if (projectInfo.FullFramework)
```
Mobrockers: Have you tested stryker with this removed? We have not been able to get buildalyzer to recognize all project dependencies without building and analyzing twice, because buildalyzer cleans during build.

src/Stryker.Core/Stryker.Core/FilePathUtils.cs
```diff
@@ -14,4 +14,13 @@ public static string ConvertPathSeparators(string filePath)
 			return filePath.Replace(windowsDirectorySeparator, Path.DirectorySeparatorChar);
 		}
 	}
+
+
+    public static string ToFullPath(this string path)
+    {
+        if (path == null)
```
Mobrockers: We don't use shorthand if's

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
```diff
@@ -27,22 +27,13 @@ public ProjectFileReader()
             _logger = ApplicationLogging.LoggerFactory.CreateLogger<ProjectFileReader>();
         }
 
-        public ProjectAnalyzerResult AnalyzeProject(string projectFilePath, string solutionFilePath = null)
+        public ProjectAnalyzerResult AnalyzeProject(string projectFilePath, string solutionFilePath)
         {
-            AnalyzerResult analyzerResult = null;
-            AnalyzerManager manager = null;
-
-            if (solutionFilePath is null)
```
Mobrockers: Using a solution is optional as dotnet core does not need the solution file to build the project. If you have a case where you do need the solution file for a proper dotnet core build, then you will still need to handle cases where a solution file is not given by the user. I would also expect the solution file commandline flag documentation to be updated if a solution file can be used on dotnet core projects.

src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs
decipherer: I was wondering why this was in there :) I've run multiple sessions after removing this and did not notice any difference. I will restore this though.

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
decipherer: Ok, will fix

src/Stryker.Core/Stryker.Core/FilePathUtils.cs
decipherer: Ok, will fix

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
decipherer: I will handle the case where a solutionfile is not present and change the command line flag documentation to reflect the change.

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
decipherer: Fixed

src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs
decipherer: Restored

src/Stryker.Core/Stryker.Core/FilePathUtils.cs
decipherer: Fixed

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
decipherer: A solutionfile is no longer mandatory, as it was before. Documentation has been updated.

decipherer:
Thank you for reviewing (especially so fast)!

I made all requested changes and made sure all checks pass.
I hade some trouble getting assembly signing to work on non-windows systems. It turned out that the DesktopStrongNameProvider does not support non-windows in version 2.10.0. I decided to use the internal type PortableStrongNameProvider from 2.10.0, rather than upgrading to 3.0.0-beta4 and using the public DesktopStrongNameProvider. This should be easy to change once 3.0.0 is out of beta.

Please have another look :)

PR #443, opened 2019-04-05
feat(report json): support html tags in source code and update html report to 1.0.0

### LINKED ISSUE #400: Remove deprecated health property from json report

### LINKED ISSUE #442: Break open html tags in mutation report json
Because we load the report json in a javascript powered html report, html tags in source code could break json parsing in the html report. To combat this we should break open all tags in the json so that they are not parsable as html.

`string.Replace(repotJson, "<", "<\" + \"")`

`<\" + \"` should be parsed back to < by the json parser eventually and thus this should not have any impact on the report itself.

Mobrockers:
Fixes #442 
Fixes #400


Mobrockers:
Broken report example (remove .txt extension to view as html): 
[mutation-report.html.txt](https://github.com/stryker-mutator/stryker-net/files/3051527/mutation-report.html.txt)

Mobrockers:
Tested on real project and works.

Issue #410, opened 2019-03-15
Metadata urls are being deprecated by nuget starting with licenseUrl


Mobrockers:
https://github.com/NuGet/Announcements/issues/32

Nuget push already warns us that licenseUrl is deprecated and should be replaced by providing the license inside the nuget package. We already have the license in our source repo so we should be able to package that in the azure pipeline instead of using the licenseUrl. We should also check what the future plans are because it looks like other metadata url's are also going to be deprecated in the future such as package icon.

Mobrockers:
Closed by #445 

New issue can be opened once NuGet has implemented deprecating the remaining metadata information

PR #445, opened 2019-04-06
chore(nuget): license

robertlyson:
Hi 👋,

PR to address https://github.com/stryker-mutator/stryker-net/issues/410.

At a first glance this looks a little bit clumsy(LICENS*)
```
  <ItemGroup>
    <None Include="..\..\..\LICENS*" Pack="true" PackagePath="" />
  </ItemGroup>
```
but this is a work around for [known issue](https://github.com/NuGet/Home/issues/7601) allowing you to include license file which doesn't have an extension.


Mobrockers:
I checked the resulting nuget package and nusepec file and it looks all good. Thanks!

Issue #444, opened 2019-04-05
Negate mutator


richardwerkman:
Currently Stryker mutates expressions from negative to non negative [here](https://github.com/stryker-mutator/stryker-net/blob/8b1b900513edcf55e07cafbf2824d7d4ecaae091/src/Stryker.Core/Stryker.Core/Mutators/PrefixUnaryMutator.cs#L21). But not the other way around. 

![image](https://user-images.githubusercontent.com/14224500/55628620-7f6a1f80-57b1-11e9-997d-1c86ba2fbe33.png)
This statement is not being mutated by stryker. It could be negated.

## Why?
It would be nice to see stryker mutate this expression into:
``` cs
if (!scan.MoveNext()) {
```

## The catch
We have to prevent double mutations from happening. For example:
![image](https://user-images.githubusercontent.com/14224500/55628698-ac1e3700-57b1-11e9-947d-8a9a24b8ca34.png)
Should not be mutated more than once. 

This means the negate mutator should specificly target expressions that do not contain any other mutations. This is hard to detect since our mutators have no knowledge of other mutations. We have to think of specific SyntaxNodes that should be mutated.

dupdob:
It may be simple: if we cannot generate mutant for the condition then Stryker add the negated mutant

It should do the trick

Bearing in mind that there is a method dedicated to if statement mutations..

Envoyé de mon iPhone

richardwerkman:
@dupdob don't we want to mutate more than if statements? I can imagine this is useful in where expressions too. Or even in method call arguments. 

We could state that every expression that contains a method call without any arguments should be negated. That way we can create a regular mutator and don't have to change the mutantorchestrator. What do you think?

dupdob:
@richardwerkman : well, that sounds hard to me: this mutator works only on boolean expression. The only places where we can be sure of that are:
- conditions within statements (if, for, while)
- boolean literals (true, false)
you add to that the logic that requires mutating expressions that cant be mutated in any other way..

dupdob:
Let me push an in interesting question: how shall we mutate this:

```csharp
if (Check(true))
{
...
}
```
does it make sens to have both:
```csharp
if (!Check(false))
{
...
}
```
and
```csharp
if (Check(true))
{
...
}
```

Hence my proposal to focus on *if, while, for (maybe)*, at least for a first try. 
That being said, you may look at other mutation framework to see their strategies for this situation.

richardwerkman:
[...]

PR #450, opened 2019-04-07
feat(mutations): negate mutator

robertlyson:
PR for https://github.com/stryker-mutator/stryker-net/issues/444.
Based on suggestion made by @dupdob(https://github.com/stryker-mutator/stryker-net/issues/444#issuecomment-480286555).

Craeting nagated if condition looks a little bit odd right now
`SyntaxFactory.ParseExpression($"!{ifStatement.Condition}");`
but I couldn't find any better way to express `!` with Roslyn. Maybe you guys will come up with something better :)

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/MutantOrchestratorTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/MutantOrchestratorTests.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.CodeAnalysis;
+﻿using System;
+using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Shouldly;
 using Stryker.Core.Mutants;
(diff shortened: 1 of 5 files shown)


Mobrockers:
Thank you for taking a look at this @robertlyson :)
I am afraid you're looking in the wrong place though. The mutant orchestrator is the component in which way we place the mutation switch (it determines with which kind of if statement we turn mutations on and off so we can place all mutations from the start without recompiling between every run). So this is not actually an if statement mutator. This just places an if with in it the actual mutation.

You can find the actual mutations we place here: https://github.com/stryker-mutator/stryker-net/tree/master/src/Stryker.Core/Stryker.Core/Mutators
Here is the boolean mutator: https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/Mutators/BooleanMutator.cs
Dupdob's suggestion is that we need a new mutator that looks for if statements.

I kindly suggest you revert your current changes, as they are unfortunately not in the right direction.

robertlyson:
@Mobrockers thanks for all of the feedback and suggestions!

PR #451, opened 2019-04-07
feat(mutations): negate mutator

robertlyson:
Issue: https://github.com/stryker-mutator/stryker-net/issues/444.
Continuation of this MR https://github.com/stryker-mutator/stryker-net/pull/450.

I added additional test into `MutantOrchestratorTests` to check how it behaves in general.


Mobrockers:
I tested this build on one of my projects and it resulted in every single mutant being killed:

image

Result from master build:
image

robertlyson:
Interesting, will have a look. Thanks for commenting.

Mobrockers:
Don't look too long. Thinking about it some more I think I know what the
problem is and I'm fairly certain the problem is in one of my tests in the
project I used. I'll look into it tomorrow morning and let you know.

Mobrockers:
I have found the issue and it is  in your code after all. I didn't catch this before but because you use ApplyMutant directly without placing the mutatant with an if structure, the mutant is **always** active and **always** causes test failures because of structures like this:

```cs
if (string.IsNullOrWhiteSpace(_apiOptions.EnvironmentName))
{
	_logProvider.Log(LogLevel.Error, Resources.ShouldNotBeNull, Guid.Empty);
	throw new ArgumentNullException("EnvironmentName", Resources.ShouldNotBeNull);
}
```

Your code mutates this to:

```cs
if (!string.IsNullOrWhiteSpace(_apiOptions.EnvironmentName))
{
	_logProvider.Log(LogLevel.Error, Resources.ShouldNotBeNull, Guid.Empty);
	throw new ArgumentNullException("EnvironmentName", Resources.ShouldNotBeNull);
}
```

Which always fails. Because we use mutation switching, the mutation should actually look like this:

```cs
if ((ActiveMutationHelper.ActiveMutation == 21) ? (!string.IsNullOrWhiteSpace(_apiOptions.EnvironmentName)) : (string.IsNullOrWhiteSpace(_apiOptions.EnvironmentName)))
{
	_logProvider.Log(LogLevel.Error, Resources.ShouldNotBeNull, Guid.Empty);
	throw new ArgumentNullException(nameof(_apiOptions.EnvironmentName), Resources.ShouldNotBeNull);
}
```

See:

https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs#L31-L40

And:

https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs#L64-L74

For how we place mutants using mutation switching.

robertlyson:
@Mobrockers thanks for looking into it! Will try to make it working :)

src/Stryker.Core/Stryker.Core/Mutators/IfStatementMutator.cs
```diff
@@ -0,0 +1,25 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Stryker.Core.Mutants;
+
+namespace Stryker.Core.Mutators
+{
+    public class IfStatementMutator : MutatorBase<IfStatementSyntax>, IMutator
+    {
+        public override IEnumerable<Mutation> ApplyMutations(IfStatementSyntax node)
```
richardwerkman: I'm not sure if this mutator should only target if statements. What about:

``` cs
While (person.IsAdult()) {
...
}
```
We could negate this too right? I think we can find more cases that could be worth negating. It might be better to look at the `ExpressionSyntax` node or some other Expression node type.

src/Stryker.Core/Stryker.Core/Mutators/IfStatementMutator.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            if (node.IsKind(SyntaxKind.IfStatement))
```
richardwerkman: node is of type IfStatementSyntax, thus it is always of kind IfStatement ;)

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -84,7 +85,17 @@ public SyntaxNode Mutate(SyntaxNode currentNode)
             }
             else if (currentNode is IfStatementSyntax ifStatement)
             {
-                return MutateIfStatement(ifStatement);
+                var mutateIfStatement = MutateIfStatement(ifStatement);
```
richardwerkman: I don't think we should change our mutating logic for this mutator. If we target expressions in the mutator, our current mutating logic will call the mutator at the right nodes for us.

src/Stryker.Core/Stryker.Core/Mutators/IfStatementMutator.cs
```diff
@@ -0,0 +1,25 @@
...
+            {
+                yield return new Mutation()
+                {
+                    OriginalNode = node,
+                    ReplacementNode = node.WithCondition(SyntaxFactory.ParseExpression($"!{node.Condition}")),
```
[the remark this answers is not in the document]
robertlyson: Looks much better, thanks.

src/Stryker.Core/Stryker.Core/Mutators/IfStatementMutator.cs
robertlyson: I think I see it now, good idea :+1:

richardwerkman: See #444 for my thoughts! I think we can target `ExpressionSyntax` nodes and look in their parent nodes. If the parent is a `IfStatementSyntax` or `WhileStatementSyntax` we could mutate the `EpxressionSyntax` node :)

robertlyson:
Just pushed some changes, let me know what do you think.

src/Stryker.Core/Stryker.Core/Mutators/ConditionStatementMutator.cs
```diff
@@ -0,0 +1,61 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Stryker.Core.Mutants;
+
+namespace Stryker.Core.Mutators
+{
+    public class ConditionStatementMutator : MutatorBase<ExpressionSyntax>, IMutator
+    {
+        public override IEnumerable<Mutation> ApplyMutations(ExpressionSyntax node)
+        {
+            if (ContainsArguments(node)) yield break;
```
Mobrockers: We don't use short form if's

src/Stryker.Core/Stryker.Core/Mutators/ConditionStatementMutator.cs
```diff
@@ -0,0 +1,61 @@
...
+            
+            if (node.Parent is IfStatementSyntax ifStatementSyntax)
+            {
+                if(!IsInvocationExpression(ifStatementSyntax.Condition)) yield break;
```
Mobrockers: We don't use short form if's

src/Stryker.Core/Stryker.Core/Mutators/ConditionStatementMutator.cs
```diff
@@ -0,0 +1,61 @@
...
+                    OriginalNode = node,
+                    ReplacementNode = replacement,
+                    DisplayName = "If statement mutation",
+                    Type = Mutator.Statement
+                };
+            }
+
+            if (node.Parent is WhileStatementSyntax whileStatementSyntax)
+            {
+                if(!IsInvocationExpression(whileStatementSyntax.Condition)) yield break;
```
Mobrockers: We don't use short form if's

src/Stryker.Core/Stryker.Core/Mutators/ConditionStatementMutator.cs
robertlyson: valid point, remember it from last time, thanks

src/Stryker.Core/Stryker.Core/Mutators/ConditionStatementMutator.cs
```diff
@@ -0,0 +1,61 @@
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Stryker.Core.Mutants;
+
+namespace Stryker.Core.Mutators
+{
+    public class ConditionStatementMutator : MutatorBase<ExpressionSyntax>, IMutator
```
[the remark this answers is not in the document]
robertlyson: yes, was struggling for a while with a name for the new mutator, but `NegateConditionMutator` sounds good

src/Stryker.Core/Stryker.Core/Mutators/Mutator.cs
```diff
@@ -26,7 +26,9 @@ public enum Mutator
         [Description("Linq methods")]
         Linq,
         [Description("String literals")]
-        String
+        String,
+        [Description("Statements")]
+        Statement
```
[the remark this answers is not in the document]
robertlyson: Thanks for pointing this out.

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
richardwerkman: Actually it turned out there was no way of doing this without changing the orchestrator...

richardwerkman:
@robertlyson thanks for your contribution! The mutator will be included in the next release 👍

Issue #462, opened 2019-04-18
"Original project under test cannot be loaded" and "Unable to rollback mutation" when BuildOutputDir is changed relative to project dir


derrickstolee:
Hello, I'm very new to using Stryker, and I'm having trouble that is likely related to the way [my project](https://github.com/microsoft/vfsforgit) puts build outputs outside of the source directory.

First, the error (and the request for me to create an issue):

```
[15:55:18 WRN] Original project under test C:\_git\VFSforGit\src\GVFS\BuildOutput\GVFS.Common\bin\AnyCPU\Debug\netcoreapp2.1\GVFS.Common.dll could not be loaded.
Embedded Resources might be missing.
System.IO.FileNotFoundException: Could not find file 'C:\_git\VFSforGit\src\GVFS\BuildOutput\GVFS.Common\bin\AnyCPU\Debug\netcoreapp2.1\GVFS.Common.dll'.
File name: 'C:\_git\VFSforGit\src\GVFS\BuildOutput\GVFS.Common\bin\AnyCPU\Debug\netcoreapp2.1\GVFS.Common.dll'
   at System.IO.FileStream.ValidateFileHandle(SafeFileHandle fileHandle)
   at System.IO.FileStream.CreateFileOpenHandle(FileMode mode, FileShare share, FileOptions options)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
   at Mono.Cecil.ModuleDefinition.ReadModule(String fileName, ReaderParameters parameters)
   at Stryker.Core.Initialisation.EmbeddedResourcesGenerator.LoadModule(String assemblyPath, ILogger logger) in C:\_git\stryker-net\src\Stryker.Core\Stryker.Core\Initialisation\EmbeddedResourcesGenerator.cs:line 42
[15:55:19 ERR] Unable to rollback mutation for node FileAttributes with error: Ambiguity between 'FileProperties.FileAttributes' and 'FileProperties.FileAttributes' (at C:\_git\VFSforGit\src\GVFS\GVFS.Common\FileSystem\FileProperties.cs, line: 7, col: 80).
[15:55:19 ERR] This should not happen, please report this as an issue on github with the previous error message.
```

[...]

PR #473, opened 2019-04-22
feat(rollback mutations): Recompile 100 times

derrickstolee:
The CompilingProcess has a hard-coded three retries to fix any
compiler errors that are fixed by recompiling. Instead, loop until
the recompile works, or until the recompile fails 100 times in a row.

Thanks to @Mobrockers for the assistance in with the variable
names to properly run the recompile.

Resolves one issue for #462.


Mobrockers:
I fixed the one failing unit test for you and added some logging so we can diagnose how many recompilations we use. Hope you don't mind.

derrickstolee:
> I fixed the one failing unit test for you and added some logging so we can diagnose how many recompilations we use. Hope you don't mind.

I absolutely don't mind. I remember seeing the unit test that was associated when I first made my change and completely forgot about it when changing the infinite loop to the recommended finite loop. Thanks for all the help here!

PR #479, opened 2019-04-26
fix(shared projects): Only include .projitems files in shared projects import

### LINKED ISSUE #462: "Original project under test cannot be loaded" and "Unable to rollback mutation" when BuildOutputDir is changed relative to project dir

Mobrockers:
For #462


src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InputFileResolverTests.cs
```diff
@@ -153,6 +153,131 @@ public void InputFileResolver_InitializeShouldResolveImportedProject()
...
+    <ItemGroup>
+    </ItemGroup>
+               
+     <Import Project=""../SharedProject/Example.props"" Label=""Shared"" />
+
+</Project>";
+
+            var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData>
+                {
+                    { Path.Combine(_filesystemRoot, "SharedProject", "Example.props"), new MockFileData("")},
```
richardwerkman: This is misleading. The folder "SharedProject" does not contain a sharedproject in this case :)

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InputFileResolverTests.cs
```diff
@@ -153,6 +153,131 @@ public void InputFileResolver_InitializeShouldResolveImportedProject()
...
+                });
+            var target = new InputFileResolver(fileSystem, projectFileReaderMock.Object);
+
+            var result = target.ResolveInput(new StrykerOptions(fileSystem: fileSystem, basePath: _basePath));
+
+            result.ProjectContents.GetAllFiles().Count().ShouldBe(2);
+        }
+
+        [Fact]
+        public void InputFileResolver_InitializeShouldOnlyResolveImportedSharedProjectsFileType()
```
richardwerkman: This test seems to be (almost) the same as the test 2 above here.

PR #482, opened 2019-04-27
feat(compiling): Compile assembly version information

### LINKED ISSUE #480: Assembly version missing in mutated assembly
When running stryker on [vfsforgit](https://github.com/microsoft/vfsforgit) for #462  we encountered a lot of false positives. We noticed this has to do with missing assembly version info in the mutated assembly. Some code in the project reads the assembly version, which fails.

Stryker should produce an assembly that has exactly the same assembly version as the original assembly to ensure tests cannot fail because of this issue.

Mobrockers:
Fix #480 
Improve #462


src/Stryker.Core/Stryker.Core/Compiling/CompilingProcess.cs
```diff
@@ -50,25 +51,40 @@ public CompilingProcessResult Compile(IEnumerable<SyntaxTree> syntaxTrees, Memor
...
                                                       cryptoKeyFile: analyzerResult.SignAssembly ? analyzerResult.AssemblyOriginatorKeyFile : null,
                                                       strongNameProvider: analyzerResult.SignAssembly ? new DesktopStrongNameProvider() : null),
-
                 references: _input.AssemblyReferences);
+
             RollbackProcessResult rollbackProcessResult = null;
 
             // first try compiling
-            var emitResult = compiler.Emit(ms, manifestResources: analyzerResult.Resources);
+            var emitResult = compilation.Emit(ms,
```
richardwerkman: Maybe we could try to call the `Emit` method only once in this file. I don't like that we have two spots where we call the actual compiler. If we merge this we are sure the emit result is always correct.

src/Stryker.Core/Stryker.Core.UnitTest/Compiling/CompilingProcessTests.cs
```diff
@@ -239,5 +238,58 @@ public int Subtract(int first, int second)
...
+            var rollbackProcessMock = new Mock<IRollbackProcess>(MockBehavior.Strict);
+
+            var target = new CompilingProcess(input, rollbackProcessMock.Object);
+
+            using (var ms = new MemoryStream())
+            {
+                var result = target.Compile(new Collection<SyntaxTree>() { syntaxTree }, ms, false);
+                result.Success.ShouldBe(true);
+
+                Assembly.Load(ms.ToArray()).GetName().Version.ToString().ShouldBe("0.0.0.0");
```
richardwerkman: Should we not assert for GetVersionInfo as well? Since that is the example we saw in VFSForGit

https://github.com/Microsoft/VFSForGit/blob/25d21da9e286eef33db635833a92cad0055923da/GVFS/GVFS.Common/ProcessHelper.cs#L46

Mobrockers: Yes, but we can't because to do so we have to write to disk as FileVersionInfo.GetVersionInfo only accepts a string path and thus cannot use our mock filesystem

src/Stryker.Core/Stryker.Core/Compiling/CompilingProcess.cs
Mobrockers: Good idea, we keep having to call emit in multiple places every time we change it

PR #488, opened 2019-05-04
feat(Xamarin support): Run stryker.net on xamarin projects

richardwerkman:
We have to skip xaml.cs files while resolving files. The roslyn compiler cannot handle these files.


src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -102,14 +102,17 @@ private FolderComposite FindInputFiles(string path, List<string> filesToExclude,
             }
             foreach (var file in _fileSystem.Directory.GetFiles(folderComposite.FullPath, "*.cs", SearchOption.TopDirectoryOnly))
             {
-                var fileName = Path.GetFileName(file);
-                folderComposite.Add(new FileLeaf()
+                if (!file.EndsWith(".xaml.cs"))
```
Mobrockers: I would like some comment on why we are excluding xaml.cs files here so that we can find it back later

Issue #322, opened 2019-01-26
Specify what files to mutate


richardwerkman:
## Introduction
Currenty it is possible to exclude files. This makes filtering possible so we can specify a few files not to mutate if these for example contain logic that aren't needed to be tested (like a `startup.cs`). 

## Issue
We wish to expand this feature by specifying what files should be mutated. All other files should be excluded. This can be useful on large projects. When a developer made changes in only a few files he could want to do a quick run instead of a full run on all files.

## Requirements
- A option `--files-to-mutate` / `-ftm` should be added which expects a json array of strings.
- All files not specified should be excluded
- When a folder is specified, all files found in this folder should be mutated
- When a file/folder is specified that is not found a warning should be logged
- When this feature is enabled, file exclusion may not be enabled. Also the other way around.

royto:
Have you considered using using [Glob](https://en.wikipedia.org/wiki/Glob_(programming)) ? 

It can be more productive and clean solution.

Glob patterns specify sets of filenames with wildcard characters.

It will allow to specify files more concisely instead of specifying all files ... using wildcard, ...

We could merge the 2 options  --files-to-mutate and --files-to-exclude into one as glob support exclusion via !

example: 

`--files-to-mutate ["**/services/*.cs", "!**/services/Azure/*"]`

[DotNet.Glob](https://github.com/dazinator/DotNet.Glob) could be a solution for this purpose

royto:
This is the way it is implemented in [JS stryker mutator ](https://github.com/stryker-mutator/stryker/blob/master/packages/stryker/README.md#mutate-string)

royto:
After digging into DotNet.Glob, it currently does not support Extended globbing which allow negate feature nativelly.

I think it can be achieve in negating match if pattern starts with an `!`

richardwerkman:
Great idea! I think it will be a little more complex, but also better in usage.

We should think about cases like, what if a file is both included and excluded. And what if a file is included in an excluded folder? Or the other way around.

By default it should be that all files are mutated. And when one or more globs like: `"/services/**/*.cs"` are presented, all other files should be excluded.

anhaehne:
Could this be extended to also specify which segments of the files should be included/excluded?

This could be useful in a pull request build where you only want to include mutations for the changed code to minimize the burden on the developer, since he only has to worry about his own changes. Also this would reduce the amount of mutation stryker would have to check and thus improve the performance.

[...]

Issue #459, opened 2019-04-12
Multi project runs: Mutate more projects at once


SierraNL:
In a situation where you use both normal unittests and SpecFlow tests, you can end up with two test projects, testing the same implementation project.

Now you have to run stryker twice and you end up with two results. Which probably have a certain amount of overlap.

It would be helpful if you can merge the results, or even better, run both projects at once, so you have only one result.

richardwerkman:
This would indeed be a cool feature! I'd like to see a way to tell stryker what projects to mutate and what test projects to run. Performance wise this would be a killing feature. We may need to think about ways to minimize the number of testruns.

For example: `project.x` has two test projects: `testproject.x` and `testproject.y`. And `project.z` has one test project: `testproject.z`.

That way we know we won't have to run `testproject.z` for mutations made in `project.x`. 

With this feature we should immediately build the merging of the html reports.

richardwerkman:
I discussed this with @Mobrockers today and we see two aproaches for this. 

* Always run from the solution file and run all unit tests and mutate all logic
    * On no configuration, all logic will be mutated
* Always run from a test project and include more unit test projects or more logic projects
    * On no configuration, only 1 logic project will be mutated

The first option seems to have some advantages over the second one:
* Buildalyzer is more precise when the solution path is always given
* With no configuration, a full report will be generated

The second option also has an advantage:
* Running from the solution file can be hard to configure on large projects (10+ csproj files) and be very slow

richardwerkman:
Today I've created a design for the multi project run feature.

It looks like this:

![image](https://user-images.githubusercontent.com/14224500/58338393-11a0b280-7e48-11e9-8031-06a8babded85.png)

The idea is that we'll place a config file in each project. This makes excluding files possible in each project that should be mutated. This also keeps the current way of running stryker possible. That way we stay backwards compatible.

I'll create smaller issues to get to the final design in smaller steps.

OskarKlintrot:
Is there any documentation on how to make this work? I added a `stryker-config.json` to my root (where my .sln-file is) and one to each project that contains which test projects to run. When I then run `dotnet stryker` or `dotnet stryker --solution MySolution.sln` in the root it says:
```
No .csproj file found, please check your project directory at C:\...
```

rouke-broersma:
Hi @OskarKlintrot 

Unfortunately we cannot recommend using solution runs at this time, because you will almost guaranteed hit #1665 if you do.

[...]

PR #512, opened 2019-05-14
feat(Mutate file only): Specify what files to mutate

richardwerkman:
This PR aims to improve file exclusion. It works as described in #322.


src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
```diff
@@ -90,17 +90,16 @@ public static class CLIOptions
         public static readonly CLIOption<int> MaxConcurrentTestRunners = new CLIOption<int>
         {
             ArgumentName = "--max-concurrent-test-runners",
-            ArgumentShortName = "-m <integer>",
+            ArgumentShortName = "-c <integer>",
```
Mobrockers: This is super breaking, we should consider this carefully and I would prefer to do this in a separate pr so we can go over and fix all flags we deem wrong at once.

richardwerkman:
This should change to getting the config from the project under test. This has to do with the future plans described in #459

PR #528, opened 2019-05-24
break(Configuration): Update option names

richardwerkman:


src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
```diff
@@ -55,7 +55,7 @@ public static class CLIOptions
             ArgumentName = "--dev-mode",
             ArgumentShortName = "-dm",
```
Mobrockers: We were going to change this to -dev

PR #530, opened 2019-05-26
use "y" counter instead of "i" in nested loop in tests

DamirAinullin:
Inside nested loop we create 4 FileLeafs with the same file name. I suggest using nested loop variable for filename.


Mobrockers:
Thank you!

PR #588, opened 2019-06-18
feat(Specify test project): Specify test project at any relative path. 

### LINKED ISSUE #575: Determine unity testproject and project under test projectfile locations

MadsPB:
This PR aims to allow Stryker to be executed from any directory, with any amount of project files in the directory by specifying the path to a specified test project. It also aims to execute Stryker in a directory containing both the test project and project under test.

The PR is a step on the road to enable Stryker-net to be used with Unity solutions. (fixes #575 )
All Unity project files are generated in the root solution folder.

I have left one to do as I was not sure how to handle OS specific path separators. Especially the C:// vs C:/
I hope to get some help here.


src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
```diff
@@ -241,5 +244,10 @@ private string ValidateSolutionPath(string basePath, string solutionPath)
 
             return solutionPath;
         }
+
+        public void UpdateBasePath(string newBaseDirectory)
```
richardwerkman: Do we have to update the `BasePath` later than the creation of the options? Why not set the `BasePath` with a value that is passed. If it was not passed set it to the current directory (should  be the test project directory). That way the config doesn't have to be updated later. And validation can just be done like all other configs.

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialisationProcessTests.cs
```diff
@@ -67,6 +66,60 @@ public void InitialisationProcess_ShouldCallNeededResolvers()
...
+                initialTestProcessMock.Object,
+                testRunnerMock.Object,
+                assemblyReferenceResolverMock.Object);
+
+            var options = new StrykerOptions();
+
+            var result = target.Initialize(options);
+
+            inputFileResolverMock.Verify(x => x.ResolveInput(It.IsAny<StrykerOptions>()), Times.Once);
+            options.BasePath.ShouldBe( "C:/Example/Dir/ProjectFolder/TestFolder"); //ToDo: expect OS specific path
```
richardwerkman: You can use the PathUtils for the todo job :+1:

src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
MadsPB: That's a great idea. I was considering it a bit, but it was too late(in the night) to consider the scope of use of the base path.

I thought the Options would be immutable, i would also consider it a value object, but then i noticed the TestRunner being get set as well.

I will change this to set the basepath to the provided test project path if one is provided.

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialisationProcessTests.cs
MadsPB: I couldn't really make it work with PathUtils.
my input path was C://one/two/three
but the path detected by the system. Path.PathDirectory 
would be C:/one/two/three
missing one /

I'll try again (after moving this to the CLI)

src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
richardwerkman: You are right, the `TestRunner` is a bit of a flaw in the design. It is set later to VsTest if we detect some values in the project file. It would be better to not change the option there but the actual testrunner object that is used. For now this was the easy solution.

src/Stryker.CLI/Stryker.CLI.sln
```diff
@@ -17,10 +17,10 @@ Global
 		Release|Any CPU = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
-		{7B35D5F5-F983-4034-8A15-952E8C415010}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{7B35D5F5-F983-4034-8A15-952E8C415010}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{7B35D5F5-F983-4034-8A15-952E8C415010}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{7B35D5F5-F983-4034-8A15-952E8C415010}.Release|Any CPU.Build.0 = Release|Any CPU
+		{7B35D5F5-F983-4034-8A15-952E8C415010}.Debug|Any CPU.ActiveCfg = TestProject|Any CPU
```
Mobrockers: What is this config used for?

src/Stryker.CLI/Stryker.CLI/Stryker.CLI.csproj
```diff
@@ -7,6 +7,8 @@
     <OutputType>Exe</OutputType>
     <PackAsTool>true</PackAsTool>
     <ToolCommandName>dotnet-stryker</ToolCommandName>
+    <Configurations>Debug;Release;TestProject</Configurations>
```
Mobrockers: What is this configuration used for?

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialisationProcessTests.cs
```diff
@@ -67,7 +66,7 @@ public void InitialisationProcess_ShouldCallNeededResolvers()
 
             inputFileResolverMock.Verify(x => x.ResolveInput(It.IsAny<StrykerOptions>()), Times.Once);
         }
-
+
```
Mobrockers: Can you revert the whitespace changes in this file

src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
```diff
@@ -86,6 +86,14 @@ public static class CLIOptions
             ArgumentDescription = @"Used for matching the project references when finding the project to mutate. Example: ""ExampleProject.csproj""",
             JsonKey = "project-file"
         };
+       
+        public static readonly CLIOption<string> TestProjectFileName = new CLIOption<string>
```
richardwerkman: I'm not sure if this should be the exact name or a filter like the `project-file` param. For the project to mutate we filter on the name passed by the user. It could for example be:

```
--project-file "myProject"
```
if the project file is `C:/myproject/myproject.csproj`

But the following would also be accepted:
```
--project-file "myproject/myProject"
```
which gives user a less strict experience.

@Mobrockers what do you think? Should this be the exact name or a filter?

src/Stryker.CLI/Stryker.CLI/OptionsBuilder.cs
```diff
@@ -40,10 +42,20 @@ public class OptionsBuilder
                         .Build().GetSection("stryker-config");
             }
 
+            var testProjectName = testProjectNameFilter.Value();
+            if (testProjectNameFilter.HasValue())
```
richardwerkman: Is this neccacary? If we just pass the filter as it is we can later determine during the run if the filter should be used and how it should be used. I don't think the `OptionsBuilder` should be responsible for modifying the value.

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -143,11 +146,25 @@ private void MarkInputFilesAsExcluded(FolderComposite root, List<string> pathsTo
...
+        {
+            var testProjectFile = _fileSystem.Directory.GetFiles(currentDirectory, $"{testProjectNameFilter}");
+
+            if (!testProjectFile.Any())
+            {
+                throw new StrykerInputException(
+                    $"No .csproj file found, matching testProjectParameter: {testProjectNameFilter}");
+            }
+            
+            return testProjectFile.First();
```
Mobrockers: If we find more than one project matching the name filter we should throw an input exception. We should not make assumptions about the first we find being the correct one. This will lead to unexpected results for users..

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -157,7 +174,7 @@ public string ScanProjectFile(string currentDirectory)
                     sb.AppendLine(file);
                 }
                 sb.AppendLine();
-                sb.AppendLine("Please fix your project contents");
+                sb.AppendLine("Please specify a testProjectFile.");
```
Mobrockers: You could add examples here as we do in find project under test since we already know all projectFiles the user could choose.

src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
Mobrockers: @richardwerkman If we only change the actual object then the used configuration will differ from what the user chose in an implicit way. This means in other parts of stryker we will not be able to determine what testrunner is used (as we only have the options object in most places). That would not be ideal and could lead to bugs.

src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
Mobrockers: I would prefer a filter, as it is implemented currently.

src/Stryker.CLI/Stryker.CLI/OptionsBuilder.cs
Mobrockers: testProjectNameFilter.Value() should simply be passed to GetOption() and handled just like all other options, agreed.

src/Stryker.CLI/Stryker.CLI.sln
MadsPB: Actually, it is not used for anything.
I was trying to follow the guide to run the CLI.
```Clone the repository https://github.com/stryker-mutator/stryker-net.git
Open Stryker.CLI.sln
On Stryker.CLI open properties > Debug
Create a new Debug profile
Set Launch as Project
Set WorkingDirectory as your local installation dir, pointing to a UnitTest project example: (C:\Repos\stryker-net\src\Stryker.Core\Stryker.Core.UnitTest)
Run the program with Stryker.CLI as the startup project with the newly created Debug profile```

I should not have committed this.

File: src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs (hunk 8 above)
Comment (MadsPB) [reply src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs]: Good catch

File: src/Stryker.CLI/Stryker.CLI.sln (hunk 3 above)
Comment (richardwerkman) [reply src/Stryker.CLI/Stryker.CLI.sln]: Did you succeed in running the CLI? It looks like you created a SolutionConfiguration instead of a debug configuration.

File: src/Stryker.Core/Stryker.Core/TestRunners/TestRunnerFactory.cs [hunk 10]
```diff
@@ -25,13 +26,13 @@ public ITestRunner Create(StrykerOptions options, ProjectInfo projectInfo)
             switch (options.TestRunner)
             {
                 case TestRunner.DotnetTest:
-                    testRunner = new DotnetTestRunner(options.BasePath, new ProcessExecutor(), new TotalNumberOfTestsParser());
+                    testRunner = new DotnetTestRunner(Path.GetDirectoryName(projectInfo.TestProjectAnalyzerResult.ProjectFilePath), new ProcessExecutor(), new TotalNumberOfTestsParser());
```
Mobrockers: I would prefer the callee becomes responsible for correctly handling getting the directory name from the path since we are dealing with strings and have no mechanism to validate the string input so cannot force the caller to make the right decision.

src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs
```diff
@@ -44,7 +44,8 @@ public MutationTestInput Initialize(StrykerOptions options)
             var projectInfo = _inputFileResolver.ResolveInput(options);
 
             // initial build
-            _initialBuildProcess.InitialBuild(projectInfo.FullFramework, options.BasePath, options.SolutionPath, Path.GetFileName(projectInfo.TestProjectAnalyzerResult.ProjectFilePath));
+            _initialBuildProcess.InitialBuild(projectInfo.FullFramework, Path.GetDirectoryName(projectInfo.TestProjectAnalyzerResult.ProjectFilePath),
```
Mobrockers: I would prefer the callee becomes responsible for correctly handling getting the directory name from the path since we are dealing with strings and have no mechanism to validate the string input so cannot force the caller to make the right decision.

src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs
```diff
@@ -44,7 +44,8 @@ public MutationTestInput Initialize(StrykerOptions options)
             var projectInfo = _inputFileResolver.ResolveInput(options);
 
             // initial build
-            _initialBuildProcess.InitialBuild(projectInfo.FullFramework, options.BasePath, options.SolutionPath, Path.GetFileName(projectInfo.TestProjectAnalyzerResult.ProjectFilePath));
+            _initialBuildProcess.InitialBuild(projectInfo.FullFramework, Path.GetDirectoryName(projectInfo.TestProjectAnalyzerResult.ProjectFilePath), 
+                options.SolutionPath, Path.GetFileName(projectInfo.TestProjectAnalyzerResult.ProjectFilePath));
```
Mobrockers: Same here but for the filename. I see we implemented it like this ourselves before, now that I notice I disagree with our implementation. Could you fix it here as well?

src/Stryker.Core/Stryker.Core/TestRunners/TestRunnerFactory.cs
```diff
@@ -25,13 +26,13 @@ public ITestRunner Create(StrykerOptions options, ProjectInfo projectInfo)
...
                     break;
                 case TestRunner.VsTest:
                     testRunner = new VsTestRunnerPool(options, projectInfo);
                     break;
                 default:
-                    testRunner = new DotnetTestRunner(options.BasePath, new ProcessExecutor(), new TotalNumberOfTestsParser());
+                    testRunner = new DotnetTestRunner(Path.GetDirectoryName(projectInfo.TestProjectAnalyzerResult.ProjectFilePath), new ProcessExecutor(), new TotalNumberOfTestsParser());
```
Mobrockers: Same here

Mobrockers:
Hey @MadsPB , are you still able to work on this?

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -152,10 +153,12 @@ private void MarkInputFilesAsExcluded(FolderComposite root, List<string> pathsTo
             }
         }
 
-        public string ScanProjectFile(string currentDirectory)
+        public string FindProjectFile(string currentDirectory, string testProjectNameFilter)
         {
-            var projectFiles = _fileSystem.Directory.GetFiles(currentDirectory, "*.csproj");
-            _logger.LogTrace("Scanned the current directory for *.csproj files: found {0}", projectFiles);
+            var projectFiles = _fileSystem.Directory.GetFiles(currentDirectory, testProjectNameFilter);
```
richardwerkman: What will happen when the path `../../myproject.csproj` is passed as filter?Is this a valid filter for Directory.GetFiles? Or the path `C:/dev/myproject/mytestproject/mytestproject.csproj`? 

Also, do we validate what kind of filter is passed? I think we can already rule out wrong filters like a missing `csproj` in the end.

Mobrockers: A filter without csproj would still work as long as the partial file name matches only one file

Mobrockers: The docs say the following:

```
searchPattern cannot end in two periods ("..") or contain two periods ("..") followed by DirectorySeparatorChar or AltDirectorySeparatorChar, nor can it contain any invalid characters. You can query for invalid characters by using the GetInvalidPathChars method.
```

So it seems like relative paths are not supported

Mobrockers: @richardwerkman I fixed the relative and full paths by converting relative paths with navigation characters to full paths in StrykerOptions so the filter always contains either a full path or a relative path without navigation characters. In the FindProjectFile method I remove the basePath from the full path if it exists, so it will at that point be converted to a relative path without navigation. This should fix all cases I think.

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -165,17 +170,33 @@ public string ScanProjectFile(string currentDirectory)
...
             return projectFiles.Single();
         }
 
+        private string BuildTestProjectFilter(string basePath, string testProjectNameFilter)
+        {
+            // Make sure the filter is relative to the base path otherwise we cannot find it
+            var filter = FilePathUtils.ConvertPathSeparators(testProjectNameFilter.Replace(basePath, "", StringComparison.InvariantCultureIgnoreCase));
+
+            // If the filter starts with directory separator char, remove it
+            filter = filter.Replace("*", "").StartsWith(Path.DirectorySeparatorChar) ?
```
richardwerkman: For readability this could be made an if statement instead of conditional.

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
richardwerkman: Cool! Looks good :+1:

PR #612, opened 2019-07-01
feat(update checks): Add version updates check

### LINKED ISSUE #524: Notify users about update
Since stryker is downloaded as a global tool it is unclear when to update stryker.

When stryker starts it could lookup if a new version exists on nuget. If a newer version exists a warning/info log could notify users about updating using the `dotnet tool update dotnet-stryker  -g` command to update.

**Additional context**
The lookup for newer versions should be non blocking and happen in a background thread. If it fails stryker should of course keep on running.

Mobrockers:
Fix #524


src/Stryker.CLI/Stryker.CLI/StrykerNugetFeedInfo.cs
```diff
@@ -0,0 +1,58 @@
...
+            }
+        }
+
+        public static StrykerNugetFeedInfo Create()
+        {
+            using (var httpclient = new HttpClient())
+            {
+                try
+                {
+                    var json = httpclient.GetStringAsync("https://api.nuget.org/v3/registration3/dotnet-stryker/index.json").Result;
```
richardwerkman: This is a blocking call. What if somebody runs stryker with no internet connection? Stryker should run just as expected and not crash because this check fails. Can this be made an asynchonous task?

Mobrockers: Almost all our processes are synchronous, there's is nothing that would run while waiting on this callback so it would be blocking anyway. If it crashes it simply returns null. Perhaps the timeout could be shorter. Or am I missing something?

Mobrockers: I just checked with no internet connection available the call immediately causes an exception which causes the version check to be skipped.

PR #615, opened 2019-07-02
Fix: coverage failing on net full framework

dupdob:
This PR contains only the fix for the problem with code coverage capture and Net Full Framework.
The problem seems to be that the name built for the pipe looked like an URL and triggered some security checks that ultimately prevented the pipe to be created.


Mobrockers:
It works, awesome! Went from 25 minutes to 3 minutes on one of my projects 400 mutations skipped 🙈

dupdob:
Glad to hear that!

Envoyé de mon iPhone

PR #624, opened 2019-07-10
Coverage for statics: fix #623 + mutation logic refactor

### LINKED ISSUE #579: Mutations for Linq methods don't work
**Describe the bug**
This code with the Linq method doesn't mutate. 
```
public int TestLinq(int count)
{ 
    var list = Enumerable.Range(1, count);
    return list.Last();
}
```
Also I tried to move Linq method from return statement.
```
public int TestLinq(int count)
{
   var list = Enumerable.Range(1, count);
   var last = list.Last();
   return last;
}
```
And also I tried another methods (First, FirstOrDefault).
Unit test for this method.
```
[Test]
public void TestLinq_One_Equal()
{
    var r = _sl.TestLinq(5);
    Assert.AreEqual(5, r);
}
```

**Expected behavior**
Linq method should be mutated Last to LastOrDefault (etc) and vice versa.

 - OS: Windows 10
 - Type of project: core
 - Framework Version: netcoreapp2.2
 - Stryker Version: 0.11.0 (beta)

**Additional context**
Others mutation for another methods works well.

### LINKED ISSUE #623: Static fields may not be properly tested when using coverage analysis
**Describe the bug**
Static fields and static constructors may have surviving mutants whereas they are properly tested for. This happen when using coverage analysis to speed up tests. It is because only the first test accessing the field/triggering the static constructor is identified as covering the mutants.

**Expected behavior**
Stryker should compensate for this phenomenon and run the correct tests.

 - OS: Any
 - Type of project Any
 - Framework Version Any
 - Stryker Version: beta 0.12.0

dupdob:
Fix issue #623:
mutation on static fields and static constructors are flagged during mutation process and are tested against ALL tests, ignoring (incomplete) coverage data.

Refactor MutantOrchestractor:
This was due since the kludgy fix I provided for v0.9 (I think). I hope you will be happy with the result; I have been able to revert to a design that is close to the original design. This refactor has the following (non breaking) impacts:
1 - Mutant generation order is slightly changed, so ID will be different between this version and previous ones. I don't think it has any importance, plus it was not feasible to keep previous behaviour
2 - It is able to generate a few mutants more, as the previous logic was so convoluted that it sometimes skipped mutating part of the code.

EDIT: plus I have somewhat improved the error message for assertion on mutated file, so that it highlights the first different lines, to ease diagnostic
This also probably fixes #579.


src/Stryker.Core/Stryker.Core/Mutants/MutationContext.cs
```diff
@@ -0,0 +1,49 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+//  <copyright file="MutationContext.cs" company="Stryker-Mutator.Net">
```
Mobrockers: Is this necessary? I personally would prefer not having this clutter in source files. Copyright does not need to be defined explicitly to be valid, you of course hold copyright on any material you write and we don't try to take this from you in any way. It's all publicly visible in the github logs. I will accept it if you do feel it's necessary of course since it is your work.

dupdob: Is it necessary? Actually it is, if you want the source code to be compliant with Apache 2 license (any license actually). Otherwise, the file is assumed to be unlicensed. 

My IDE automatically add this when I create a new file and I only made very basic cleanup. 

I don’t really care for my name being there or not. No prolific you want me to take the banner out, but I urge you to consider having one in your source files.

Mobrockers: Mhm I was not aware this is recommended so I looked it up. The way I read it, it doesn't mean the file is unlicensed but it is recommended to add it explicitly so the license information is more likely to be included when a single file is shared. The shared file would still fall under apache2 but the sharer would not necessarily know they are non-compliant with our license. Is that correct? In any case I'm not falling over whichever name is in the text I just believe it's unnecessary clutter. I am not a legal expert though so now idk if we should have it or not..

dupdob: I understand  this part as the recommendation to enclose some information within each file. But this is a should, not a must.
>The text should be enclosed in the appropriate
>     comment syntax for the file format. We also recommend that a
>    file or class name and description of purpose be included on the
 >   same "printed page" as the copyright notice for easier
 >  identification within third-party archives.

I found this [thread](https://softwareengineering.stackexchange.com/questions/125836/do-you-have-to-include-a-license-notice-with-every-source-file) on this very topic.
I just pushed a commit without the header. 
I still think you should have some minimal header info, such as project 'StrykerMutator.Net', license Apache 2 and maybe a year

Mobrockers: I think it could be a good idea. I think I'd like this to be automated maybe. I'll add it to the backlog. Thanks!

dupdob:
Warning: there is a small issue.
I just finished a run on NFluent and I am going through the report: some of the mutants are ill placed. Which means the mutation logic fails to preserve the dependence in one of the methods.
Will look into it later today

dupdob:
I have added a failing test for the location issue and implemented a fix, which ironically implies to revert to the original mutant numbering scheme.
Algorithm is also even simpler than before

Full test with NFluent is ongoing.
I have also added a unit test for the LinqMutator.

dupdob:
The run on NFluent worked as a charm. Digging through the mutants show no (new) anomaly. So this PR can be reviewed and merged.

Now that the mutant orchestrator logic is understandable again, I will make a careful analysis of the mutants, because it looks like Stryker is missing mutation opportunities, especially on long arithmetic expressions (x+y+z+t+u+v...). But this is nothing new.

dupdob:
I forgot an important message: NFluent went from ~1680 mutants up to ~1780 with the refactoring.

richardwerkman:
@dubdop Thanks for yet another great PR! I'll take a look at this tomorrow. 

What do you think about flagging the static mutations to test with all unit tests? Do you think this would outperform running each unit test in isolation during analyse? 

For example, lets say we have a test project with 50 unit tests and 8 mutations static. Could we win time by running the tests in isolation to still get coverage for the static mutations, or would we only loose time by this?

dupdob:
@richardwerkman: thanks for the feedback.
Just to clarify: this PR runs 'static mutants' against all unit tests, and this is not configurable (as of now)
As of now, the option to capture coverage per isolated test is not exposed. And it would also means controlling the current behaviour to benefits from accurate coverage info.
Assuming both options are available, my personal experience screams that this PR approach is the good one: running each test in isolation takes **one hour** with NFluent. And then you have to perform the actual test runs.
Whereas, the full run takes around 40 minutes with this PR.

My gut feeling is that 'running all tests' is better than 'coverage in isolation'. That is unless each unit tests takes multiple seconds to run AND static mutants are covered by a fraction of the tests.

On my machine, it looks like the fixed cost of running a test session is around 3 seconds, which covers starting up the required processes and runners as well as loading and analysing the test assembly.

Most unit test runs are sub second, which leads the coverage capture process to spend most of this time on this fixed costs. NFluent has 1700 tests, assuming 3 fixed cost per run => 5100 seconds/2 (number of runners) => 2550 seconds,  i.e. 40 minutes.

dupdob:
Re: isolated tests vs 'statics' detection
I forgot one important issue with 'statics' detection that makes it functionally inferior to the isolated tests:
current algorithm only detects static fields and static initialiser. It does not identify mutants inside methods that are called by static initialiser. Those still trigger false positives (survivors) due to insufficient testing. It turns out that my last (fake) survivor is such a mutant.

As such, the current algorithm is only an incremental improvement, but it probably covers 80%+ of situations.
Meaning, the PR makes sense.

I have two ideas to detect those mutants during the coverage capture phase:
1 - analyse stack trace during coverage capture phase to detect a static initialiser. Pros: non intrusive, cons: significant slow down, since stack walking would happen for each call.
2 - inject marking logic in static initialisers. Pros: fast, cons: intrusive

I think approach 2 is the more sensible, so I will try this one first

src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs
```diff
@@ -52,13 +52,25 @@ private static SyntaxNode RemoveByIfStatement(IfStatementSyntax ifStatement)
...
             // place the mutated statement inside the if statement
-            return SyntaxFactory.ParenthesizedExpression(
+            var result = SyntaxFactory.ParenthesizedExpression(
                 SyntaxFactory.ConditionalExpression(
                 condition: GetBinaryExpression(mutantId),
                 whenTrue: mutated,
                 whenFalse: original))
                 // Mark this node as a MutationConditional node. Store the MutantId in the annotation to retrace the mutant later
                 .WithAdditionalAnnotations(new SyntaxAnnotation(MutationConditional, mutantId.ToString()));
+            // provides the new location of the original node
```
richardwerkman: What does this comment mean? It doesn't add much clarity for me. Also the variable declaration doesn't seem necessary here.

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
```diff
@@ -200,12 +200,20 @@ public void Optimize(TestCoverageInfos coveredMutants)
...
+                else
+                {
+                    mutant.CoveringTest = tests.Select(x => x.ToString()).ToList();
+                }
             }
 
             report.AppendJoin(',', nonTested.Select(x => x.Id));
             Logger.LogInformation(nonTested.Count == 0
                 ? "Congratulations, all mutants are covered by tests!"
+
```
richardwerkman: This can be removed?

src/Stryker.Core/Stryker.Core/Mutators/LinqMutator.cs
```diff
@@ -40,30 +40,54 @@ public LinqMutator()
                 { LinqExpression.ThenBy, LinqExpression.ThenByDescending },
                 { LinqExpression.ThenByDescending, LinqExpression.ThenBy }
             };
+            RequireFilter = new Dictionary<LinqExpression, bool>
```
richardwerkman: The name RequireFilter doesn't cover what this does. Maybe:
```
RequireArgumentsFilter = new Dictionary<LinqExpression, bool>
```

richardwerkman: This could also be an array instead of dictionary. Just add only the Expression types that have a "true" value and check if it is present in the array. That would shorten this collection a lot!

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
dupdob: This is not new. What is your concern with this message?

src/Stryker.Core/Stryker.Core/Mutators/LinqMutator.cs
dupdob: done

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
richardwerkman: Sorry it was not about the message but the newline.

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
```diff
@@ -401,7 +401,15 @@ private void Dispose(bool disposing)
             {
                 if (disposing)
                 {
-                    _vsTestConsole.EndSession();
+                    try
+                    {
+                        _vsTestConsole.EndSession();
```
Mobrockers: Perhaps the VsTestRunnerPool should check whether the VsTestRunner agent is still connected before releasing the agent from the pool and if it is not connected it should remove the disconnected agent and spin up a new one. I believe vsTestConsole exposes the connection status to the consumer so we can pass that through to the pool.

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunnerPool.cs
```diff
@@ -110,12 +112,21 @@ private VsTestRunner TakeRunner()
                     Monitor.Wait(_lck);
                 }
             }
-
+            // should not happen, placed to chase an elusive bug on MacOS
+            if (runner == null)
```
Mobrockers: We could also check whether the VsTestRunner is actually still connected with VsTest because I believe the VsTestConsole exposes this information. If we detect the VsTestConsole session has ended for some reason in TakeRunner we could delete this instance and create a new one. In case vstest crashes or something we don't end up supplying the testrun with an invalid test agent.

dupdob:
@Mobrockers : sorry, I force pushed a commit to fix a stupid mistake in the previous commit. Re handling of exception on Dispose:
1. I have not seen any properties or methods giving access to a status
2. in any way, since we are disposing it, there is no point spinning up a new one, isn't there?

Mobrockers:
> @Mobrockers : sorry, I force pushed a commit to fix a stupid mistake in the previous commit. Re handling of exception on Dispose:
> 
> 1. I have not seen any properties or methods giving access to a status
> 2. in any way, since we are disposing it, there is no point spinning up a new one, isn't there?

Yea no I realised that I placed the comment on the wrong file. I meant on TakeRunner we can check that the runner is actually connected before we give the runner to the consumer.

PR #627, opened 2019-07-12
fix(AbortTestOnFail): Abort test on fail option from commandline

richardwerkman:
Makes the `--abort-test-on-fail` command available from command line


src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
```diff
@@ -100,12 +100,13 @@ public static class CLIOptions
             JsonKey = "coverage-analysis"
         };
 
-        public static readonly CLIOption<bool> AbortOnFailTest = new CLIOption<bool>
+        public static readonly CLIOption<bool> AbortTestOnFail = new CLIOption<bool>
         {
             ArgumentName = "--abort-test-on-fail",
             ArgumentShortName = "-atof",
-            DefaultValue = false,
+            DefaultValue = _defaultOptions.Optimizations.HasFlag(OptimizationFlags.AbortTestOnKill),
```
Mobrockers: Why is this enum value different from the option name?

Mobrockers:
The unit test Stryker.Core.UnitTest.TestRunners.CoverageCaptureTests.CanConnect fails on *nix and not on windows for some reason

PR #629, opened 2019-07-12
fix(VsTestRunner NPE): Use ConcurrentBag with eventhandler instead of queue with monitor

richardwerkman:
Tries to fix the build on mac


Mobrockers:
@dupdob Could you give some context as to the problem you were trying to fix by refactoring the VsTestRunnerPool from a ConcurrentQueue with eventhandler to Queue with lock?

We have noticed on the OSX pipeline that the build occasionally fails now with a NPE on TakeRunner(). Obivously if you found a problem with the old implementation we do not want to reintroduce it.

dupdob:
hi guys, thanks for asking my opinion about this.
As a rule of thumb, 99.99% of concurrence/asynchronous/parallelism problems can be solved with C# primitives. It is a well documented fact that the [Monitor pattern](https://en.wikipedia.org/wiki/Monitor_(synchronization)) is the most versatile and efficient to design for concurrency.

ConcurrentQueue is designed for heavy duty usage and is overkill for this situation, which is pretty basic.
Plus ConcurrentQueue does not alleviate the need to wait for when a Runner is available.

Can you please share a link/stack trace to the NPE? I would gladly help. As it turns out, I have ~15 years of experience working on multithreaded code... I am always glad to put them on good use

Mobrockers:
@dupdob 

```
[17:50:28 ERR] An error occurred during the mutation test run |  
-- | --
  | System.NullReferenceException: Object reference not set to an instance of an object. |  
  | at Stryker.Core.TestRunners.VsTest.VsTestRunnerPool.RunAll(Nullable`1 timeoutMs, Nullable`1 mutationId) |  
  | at Stryker.Core.Initialisation.InitialTestProcess.InitialTest(ITestRunner testRunner) |  
  | at Stryker.Core.Initialisation.InitialisationProcess.InitialTest(StrykerOptions options) |  
  | at Stryker.Core.StrykerRunner.RunMutationTest(StrykerOptions options)
```

Build: https://dev.azure.com/stryker-mutator/Stryker/_build/results?buildId=2296&view=logs&j=5eedf032-e7d0-569b-c682-175f652593e0
Another one: https://dev.azure.com/stryker-mutator/Stryker/_build/results?buildId=2259&view=logs&j=5eedf032-e7d0-569b-c682-175f652593e0
Another one:
https://dev.azure.com/stryker-mutator/Stryker/_build/results?buildId=2252&view=logs&j=5eedf032-e7d0-569b-c682-175f652593e0
Another one:
https://dev.azure.com/stryker-mutator/Stryker/_build/results?buildId=2248&view=logs&j=5eedf032-e7d0-569b-c682-175f652593e0

It seems to only fail on macosx

Mobrockers:
> Plus ConcurrentQueue does not alleviate the need to wait for when a Runner is available.

I used [WaitHandle.WaitOne](https://docs.microsoft.com/en-us/dotnet/api/system.threading.waithandle.waitone?view=netframework-4.8), does that not do exactly the same thing as the Monitor, except maybe marginally slower because it uses win32 api's? Or is that not what you meant.

dupdob:
Remember that **WaitHandle.WaitOne** will not be using Win32 APIs on Mac OS 😀.
That's one of the reason I try to use native .net primitives, as it will avoid subtle variations across OS.
That being said, working code trumps the alternative, so if you fix this with WaitHandle(s) and ConcurrentQueue, great.
The monitor pattern has the great advantage of being transactional: notification and dequeue happen within the same lock. While using ConcurrentQueue + WaitHandle forces you to enqueue before notify and there is always I small chance someone introduces a regression down the road that could lead to queue items being stoled between WaitOne and Dequeue.

dupdob:
I had a look at those build in errors. I understand you think the problems comes from takerunner.
Just a few remarks:
1. As there is no line number showing, the issue can be somewhere else. Especially if the test is run using the release version. Inlining could happen and other optimisations. I am trying to defend my code, but when dealing with concurrency, intuitions are not our best friends
2. The issue could be triggered by having a null runner added to the queue. The problems I had when working on coverage was due to the code enqueuing the SAME runner multiple times. Having a couple of checks in Take and release runner could help (if runner == null => throw argumentexception, if returning null runner => throw invalidoperation exception

I will add this in my branch, just to see what happens

dupdob:
It could also be that for some other bug the runner are being disposed while being used...

dupdob:
Sorry, it looks like my merging with the latest main kinda went wrong.
need to work on the underlying issue

Mobrockers:
> The monitor pattern has the great advantage of being transactional: notification and dequeue happen within the same lock. While using ConcurrentQueue + WaitHandle forces you to enqueue before notify and there is always I small chance someone introduces a regression down the road that could lead to queue items being stoled between WaitOne and Dequeue.

True, didn't think of that

Mobrockers:
@richardwerkman I restored the original implementation. Now running the build a couple times to see if works more reliable on osx.

Mobrockers:
@richardwerkman  5x build without issue so far

Mobrockers:
Just realized this is most likely not a problem only on OSX, but rather OSX is the only hosted microsoft agent with more than one core available so our OSX integration test is the only one using multiple cores.

dupdob:
Maybe related, but also, I use it daily on a multi core machine and the issue never occurred once!

PR #631, opened 2019-07-12
feat(zero coverage): Warn users when zero mutants are covered by a test

richardwerkman:
Rework mutant coverage to support the no coverage result. Instead of testing all not covered mutants, just skip to displaying the report.


src/Stryker.Core/Stryker.Core/Mutants/MutantStatus.cs
```diff
@@ -6,6 +6,7 @@ public enum MutantStatus
         Killed,
         Survived,
         Timeout,
+        NoCoverage,
```
Mobrockers: I would like this status to be added to the report as well

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
```diff
@@ -174,21 +181,17 @@ public StrykerRunResult Test(StrykerOptions options)
...
+
             Logger.LogDebug("Optimize test runs according to coverage info.");
-            var report = new StringBuilder();
-            var nonTested = Input.ProjectInfo.ProjectContents.Mutants.Where(x =>
-                x.ResultStatus == MutantStatus.NotRun && !covered.Contains(x.Id)).ToList();
-            const MutantStatus mutantResultStatus = MutantStatus.Survived;
-            foreach (var mutant in nonTested)
+
+            var nonCoveredMutants = Input.ProjectInfo.ProjectContents.Mutants.Where(x => x.ResultStatus == MutantStatus.NotRun && !covered.Contains(x.Id)).ToList();
+            const MutantStatus mutantResultStatus = MutantStatus.NoCoverage;
```
Mobrockers: I don't think it's neccesary to const declare an enum value especially when the variable name is less descriptive than the enum name

src/Stryker.Core/Stryker.Core/Mutants/MutantStatus.cs
richardwerkman: Isn't it already? I've seen that the Json reporter just uses ToString on this enum. So when a mutant has status NoCoverage this will be the string that ends up in the json right?

Mobrockers: Woops, I was thinking of an older iteration of the report. I forgot I made the new one a lot prettier 😅

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
Mobrockers: @richardwerkman The const is also used on line 210

Mobrockers:
https://github.com/stryker-mutator/stryker-net/blob/fe5a1fc0f1fa6815179d99dfdafd23098139c892/src/Stryker.Core/Stryker.Core/ProjectComponents/ProjectComponent.cs#L16

This should be edited to include NoCoverage mutants

richardwerkman:
Superseded by #675

PR #632, opened 2019-07-13
chore(Console Reporting): Rename 'Test Progress' to 'Mutant Progress'

Mobrockers:
To make it more clear that we are reporting on the status of the mutants we are testing and not the unit test progress.


src/Stryker.Core/Stryker.Core/Reporters/Progress/ProgressBarReporter.cs
```diff
@@ -14,10 +14,10 @@ public class ProgressBarReporter : IProgressBarReporter
         private const int MaxProgressBar = 10;
         private const char ProgressBarDoneToken = '\u2588';
         private const char ProgressBarLeftToken = '-';
-        private const string LoggingFormat = "Tests progress | {0} | {1} / {2} | {3} % | {4} |";
+        private const string LoggingFormat = "Mutant progress | {0} | {1} / {2} | {3} % | {4} |";
```
richardwerkman: Maybe something like:

```suggestion
        private const string LoggingFormat = "Testing mutants | {0} | {1} / {2} | {3} % | {4} |";
```
Covers even more what we are doing?

PR #662, opened 2019-08-13
feat(MutateArgument): Allow the user to specify which files to mutate.

anhaehne:
I created a new `IMutantFilter` that allow to specify which files should be in- or excluded as described in #322.

The `files-to-excluded` parameter is now marked as deprecated but the changes should be fully backwards compatible. If `files-to-excluded` is specified their paths will be converted to exclude file patterns in the `StrykerOptions`.

New configuration section: 
## Mutate
To specify which files should be mutated you can use file pattern to in- or excluded files or even only parts of a files. By default all files are included.

```
dotnet stryker --mutate "['C:/Repos/MyProject/MyFile.cs']"
dotnet stryker -m "['C:/Repos/MyProject/MyFile.cs']"
```

The patterns support [globbing syntax](https://en.wikipedia.org/wiki/Glob_(programming)) to allow wildcards.

```
dotnet stryker --mutate "['**/*Services.cs']"
dotnet stryker -m "['**/*Services.cs']"
```

You can add an `!` in front of the pattern to exclude all files that match the pattern.

```
dotnet stryker --mutate "['!**/*Factory.cs']"
dotnet stryker -m "['!**/*Factory.cs']"
```
When only exclude patterns are provided, all files will be included that do not match any exclude pattern. If both, include and exclude patterns, are provided, only the files that match an include pattern but not also an exclude pattern will be included. The order of the patterns is irrelevant.

### Example:

| Patterns  | File                      | Will be mutated   |
| ----------| ------------------------- | ----------------- |
| []            | MyFolder/MyFactory.cs    | Yes               |
| []            | MyFolder/MyService.cs | Yes               |
| ['\*\*/\*.\*']   | MyFolder/MyFactory.cs    | Yes               |
| ['\*\*/\*.\*']   | MyFolder/MyService.cs | Yes               |
| ['!\*\*/MyFactory.cs']   | MyFolder/MyFactory.cs    | No        |
| ['!\*\*/MyFactory.cs']   | MyFolder/MyService.cs | Yes       |
| ['!\*\*/MyFactory.cs', '\*\*/My\*.cs']   | MyFolder/MyFactory.cs    | No        |
| ['!\*\*/MyFactory.cs', '\*\*/My\*.cs']   | MyFolder/MyService.cs | Yes       |

To allow more fine grained filtering you can also specify the span of text that should be in- or excluded. A span is defined by the indices of the first character and the last character.

```
dotnet stryker --mutate "['MyFolder/MyService.cs{10..100}']"
dotnet stryker -m "['MyFolder/MyService.cs{10..100}']"
```


src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
```diff
@@ -178,7 +178,22 @@ public static class CLIOptions
...
             DefaultValue = null,
-            JsonKey = "files-to-exclude"
+            JsonKey = "files-to-exclude",
+            Deprecated = true,
+            DeprecatedMessage = "Use '--file-pattern' instead."
+        };
+
+        public static readonly CLIOption<string[]> FilePatterns = new CLIOption<string[]>
+        {
+            ArgumentName = "--file-patterns",
```
Mobrockers: We seem to not have updated the original issue, but we decided on Mutate for the option, as this is what other stryker frameworks use as well:

https://github.com/stryker-mutator/stryker-net/blob/2e479d9d181624c9edccb91431ee9ffad4849be8/src/Stryker.CLI/Stryker.CLI/CLIOptions.cs#L134

https://github.com/stryker-mutator/stryker-net/blob/2e479d9d181624c9edccb91431ee9ffad4849be8/src/Stryker.CLI/Stryker.CLI/CLIOptions.cs#L134

src/Stryker.CLI/Stryker.CLI/OptionsBuilder.cs
```diff
@@ -63,37 +71,51 @@ public class OptionsBuilder
...
-                        return ConvertTo(configValue, option);
+                        value = ConvertTo(configValue, option);
+                        hasValue = true;
                     }
                 }
             }
 
             // Unable to get value from user, return default value
-            return option.DefaultValue;
+            if (!hasValue)
```
Mobrockers: brackets.

Feel free to add that .editorconfig if that helps you keep resharper in check.

src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
```diff
@@ -99,6 +103,11 @@ private void RunStryker(StrykerOptions options)
             PrintStykerASCIIName();
             _ = PrintStrykerVersionInformationAsync();
 
+            // Configure logging
+            ApplicationLogging.ConfigureLogger(options.LogOptions);
```
Mobrockers: Moving the logging configuration here means that whatever is used to bootstrap the stryker run (Stryker.CLI) now becomes responsible to set up logging that Stryker.Core uses. Our thoughts for Stryker.CLI are that it should basically only handle user interaction and is not actually responsible for any setup, Stryker.Core owns all it's configuration and does final validation.

How about we pass some 'initial logging' to RunMutationTest instead to allow bootstrappers to pass their logging onto stryker which stryker can then handle as it sees fit?

src/Stryker.Core/Stryker.Core/Options/FilePattern.cs
```diff
@@ -0,0 +1,180 @@
...
+    /// <summary>
+    /// Contains information about which files and which parts of a file should be in- or excluded.
+    /// </summary>
+    public class FilePattern
+    {
+        private static readonly Regex _textSpanGroupRegex = new Regex("(\\{(\\d+)\\.\\.(\\d+)\\})+$");
+        private static readonly Regex _textSpanRegex = new Regex("\\{(\\d+)\\.\\.(\\d+)\\}");
+        private static readonly TextSpan _textSpanMaxValue = new TextSpan(0, int.MaxValue);
+
+        public FilePattern(Glob glob, bool isExclude, IReadOnlyCollection<TextSpan> textSpans)
```
Mobrockers:
```suggestion
        public FilePattern(Glob glob, bool isExcluded, IReadOnlyCollection<TextSpan> textSpans)
```

src/Stryker.Core/Stryker.Core/Options/FilePattern.cs
```diff
@@ -0,0 +1,180 @@
...
+        {
+            Glob = glob;
+            IsExclude = isExclude;
```
Mobrockers:
```suggestion
            IsExcluded = isExcluded;
```

src/Stryker.Core/Stryker.Core/Options/FilePattern.cs
```diff
@@ -0,0 +1,180 @@
...
+
+        /// <summary>
+        /// Gets the <see cref="Glob"/> matching the file path.
+        /// </summary>
+        public Glob Glob { get; }
+        
+        /// <summary>
+        /// Gets whether the file and text spans should be in- or excluded.
+        /// </summary>
+        public bool IsExclude { get; }
```
Mobrockers:
```suggestion
        public bool IsExcluded { get; }
```

src/Stryker.Core/Stryker.Core/ProjectComponents/IReadOnlyInputComponent.cs
```diff
@@ -13,8 +13,6 @@ public interface IReadOnlyInputComponent
         IEnumerable<IReadOnlyMutant> ReadOnlyMutants { get; }
         IEnumerable<IReadOnlyMutant> TotalMutants { get; }
         IEnumerable<IReadOnlyMutant> DetectedMutants { get; }
-        IReadOnlyCollection<FileLeaf> ExcludedFiles { get; }
-        bool IsExcluded { get; }
```
Mobrockers: We do not want to lose this information as we will want to explicitly display to users in reporting that they excluded a given file.

Mobrockers: Nevermind, I forgot you added that to the Mutant.

src/Stryker.Core/Stryker.Core/Options/FilePattern.cs
anhaehne: Wouldn't that imply that the pattern itself is excluded? The property/argument specifies that matched files should be excluded so the file pattern is an 'exclude pattern'. 

I could rename the argumen and property to `IsExcludePattern`.

Mobrockers: My bad, I completely misread the intention of the property. I think IsExcludePattern would prevent confusion a bit more but it isn't neccesary, I should've caught that.

src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
anhaehne: The log messages are now passed into the stryker runner and from there into the `ApplicationLogging.ConfigureLogging` method.

src/Stryker.CLI/Stryker.CLI/LogBuffer.cs
```diff
@@ -1,79 +1,39 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Reflection;
 using Microsoft.Extensions.Logging;
 
-namespace Stryker.CLI
+namespace Stryker.Core.Logging
```
Mobrockers: Shouldn't this namespace still be Stryker.CLI?

nicojs:
Is see you kept the argument name and argument in sync with Stryker js. Thanks for that. 🌹

nicojs:
I don't think using file spans to exclude parts of files is a good solution. What happens when someone adds a line at the top of the file? It would be a nightmare to maintain.

Better to remove it here and add another issue for excluding parts of files using exclusion comments or something. You can use https://github.com/stryker-mutator/stryker/issues/1472  as an example.

Mobrockers:
@anhaehne Could you fix the last unit tests? I it seems like you're not converting the parth seperator characters properly with FilePathUtils.ConvertPathSeperatorChars. This is causing the tests to fail on *NIX systems.

anhaehne:
Sorry for the delay, I've been on vacation. I will try to resolve the remaining issue next week. Thanks for the tip with the FilePathUtil.

Mobrockers:
Of course, no problem :) Just checking in because sometimes contributors disappear :p 
Usually we just finish the work for them in that case but we prefer if you do it yourself of course 😄

anhaehne:
For some reason the integration tests work on my machine and not in azure. Could someone try to run the integration tests locally and try to reproduce why the `EndlessLoop.cs` is not skipped in the .netStandard test?

Mobrockers:
@anhaehne It's not excluded for me when running locally, are you sure you're running locally with the same settings as on the CI? Did you check your launchSettings?

Mobrockers:
@anhaehne The problem is this:

image

You're not normalizing the filePattern, so it will never accurately match.

@richardwerkman I'm feeling like our decision to normalize all filepaths to the current filesystem is maybe the wrong choice. Perhaps we should agree on one style of filepaths to be used internally by stryker (including in filters and such) and then only normalize to the current filesystem filepaths when we're interacting with the filesystem. Then in the rest of stryker we can be sure what filepaths should look like, and always convert to "StrykerInternalPathSeperator" or something.

anhaehne:
So should we just normalize all paths to forwardslash ('/') since both windows and *nix understand those?

Mobrockers:
> So should we just normalize all paths to forwardslash ('/') since both windows and *nix understand those?

If that works, I am okay with that solution. How about you @richardwerkman ?

It does mean we need to normalize every single path or path filter going in and out of stryker. We may be missing some cases right now that accidentally work fine but might break if we change how we normalize.

anhaehne:
I've looked at the code of `System.IO.Path` and microsoft seems to just replace all forward and backward slashes with backward slashes on windows [[1]].(https://github.com/dotnet/corefx/blob/d58a51f911efb3c98beca21b6cf08cc703424fdf/src/Common/src/CoreLib/System/IO/PathInternal.Windows.cs)

On unix it's just forwards slashes since backwards slashes are not allowed anyways. [[2]].(https://github.com/dotnet/corefx/blob/d58a51f911efb3c98beca21b6cf08cc703424fdf/src/Common/src/CoreLib/System/IO/PathInternal.Unix.cs)

When normalizing we should probably just replace both with `Path.DirectorySeparatorChar` so we can still use methods like `Path.Combine` without the need to normalize the result everytime.

Mobrockers:
> I've looked at the code of `System.IO.Path` and microsoft seems to just replace all forward and backward slashes with backward slashes on windows [[1]](https://github.com/dotnet/corefx/blob/d58a51f911efb3c98beca21b6cf08cc703424fdf/src/Common/src/CoreLib/System/IO/PathInternal.Windows.cs).
> 
> On unix it's just forwards slashes since backwards slashes are not allowed anyways. [[2]](https://github.com/dotnet/corefx/blob/d58a51f911efb3c98beca21b6cf08cc703424fdf/src/Common/src/CoreLib/System/IO/PathInternal.Unix.cs).
> 
> When normalizing we should probably just replace both with `Path.DirectorySeparatorChar` so we can still use methods like `Path.Combine` without the need to normalize the result every time.
> 
> So `path.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar)` should work for every platform.

Ah I did not realize Path.AltDirectorySeparatorChar was / on windows. That is almost what we do already in FilePathUtils.Convert but it would be would be more clean to use AltDirectorySeparatorChar.

anhaehne:
Ok i found the reason one integration test did work locally on my machine. I renamed a file but only changed the casing, for which git obviously didn't care about. 😄 So if there are no more change request we can complete this PR.

Mobrockers:
Awesome! 👍 Will take a look as soon as I have time.

Mobrockers:
@anhaehne While testing locally on a full framework project from my company stryker crashed in an unexpected way while mutating. The latest released version of stryker does not have this problem. I'm currently investigating whether the problem is with this PR or with master.

anhaehne:
If you got a log or a way to reproduce it, let me know.

Mobrockers:
The issue seems to be with master, but since I can't validate whether this change works on my internal project I'm not gonna merge it until we fix master. I don't see any changes neccesary in the PR at this time though so it should get merged after.

integrationtest/TargetProjects/NetCoreTestProject.XUnit/NetCoreTestProject.XUnit.csproj
```diff
@@ -14,4 +14,9 @@
   <ItemGroup>
     <ProjectReference Include="..\TargetProject\TargetProject.csproj" Properties="TargetFramework=netcoreapp2.1" />
   </ItemGroup>
+
+  <ItemGroup>
```
Mobrockers: These should be removed

PR #674, opened 2019-08-23
Update Configuration.md

dupdob:
Provide details and advices on coverage analysis options


Mobrockers:
👍 

Also we have had a discussion this week and we feel we can soon turn coverage-analysis perTest on as the default. Do you agree?

dupdob:
This will happen eventually. All my runs haven been perfect for a few weeks now, so if you have a similar experience, then by all means, make it default. It should make people happy!

PR #705, opened 2019-09-05
fix: Dont fail compiling on warning diagnostics

Mobrockers:
Stryker failed to compile but also had a warning diagnostics error with location None. This warning was not the cause of the compilation failure. Stryker should not fail on warning diagnostics.

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Compiling/CompilingProcess.cs
+++ b/src/Stryker.Core/Stryker.Core/Compiling/CompilingProcess.cs
@@ -2,10 +2,10 @@
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.Emit;
 using Microsoft.Extensions.Logging;
+using Stryker.Core.Exceptions;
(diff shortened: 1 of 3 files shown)


src/Stryker.Core/Stryker.Core/Compiling/CompilingProcess.cs
```diff
@@ -68,11 +68,13 @@ public CompilingProcessResult Compile(IEnumerable<SyntaxTree> syntaxTrees, Memor
             EmitResult emitResult;
             var retryCount = 1;
             (rollbackProcessResult, emitResult, retryCount) = TryCompilation(ms, compilation, null, devMode, retryCount);
-            if (!emitResult.Success && emitResult.Diagnostics.Any(diag => diag.Location == Location.None))
+
+            // If compiling failed and the diagnostics failure is an error, log and throw exception
```
richardwerkman:
```suggestion
            // If compiling failed and the error has no location, log and throw exception.
```

PR #708, opened 2019-09-06
feat(Git diff): Mutate only changed files based on git diff

### LINKED ISSUE #692: Git diff check to skip mutations
PR builds with Stryker.NET enabled take too long.

Stryker should fetch the git diff between the PR branch and master and only mutate the files (and linespans) that were changed.

This could look like:
`dotnet stryker --diff`

Or if you would like the diff against another branch:
`dotnet stryker --diff develop`

Stryker should mark all mutations that are not inside the diff span as skipped. This will speed up stryker in PR builds as only the changed code will be mutated and tested.

We should see if this works fine together with the `mutate` option. What if I exclude a file in `mutate` and git diff sees a change in that file?

**Additional context**
This features depends on #322.

Edit: `dotnet stryker --gitdiff` --> `dotnet stryker --diff`

richardwerkman:
fixes #692 

Usage: 
`dotnet stryker --diff`

Will check the diff against master. If no changes were made, all files are excluded.

`dotnet stryker --diff --git-source "development"`

Will check the diff against the development branch.

This initial version only excludes files as a whole. Not parts of files. Changed line information is not available (easy) through LibGit2Sharp. When a testcase has been changed the whole project will be mutated as the change could impact all files.

In a followup PR we could try to get the changed line info from LibGit2Sharp and only mutate the changed lines. And we could try to map the changed tests back to their covered mutants. We could only test the mutants covered by the changed tests and skip all other mutants.


Mobrockers:
@nicojs @simondel what do you think about our implementation? Is that how you see it used in strykerjs as well?

anhaehne:
So this will only filter on a per file basis?

Mobrockers:
@anhaehne for now, yes. Libgit2sharp doesn't seem to have a way to provide us with usable spans at this time.

richardwerkman:
@nicojs Thanks for your feedback! Our git implementation is mainly focussed on usage in pull request builds. But it could prove useful locally as wel.

> I guess you can supply any branch, commit or tree-ish as an argument?

No for now you can only pass a branch name. But commit name support could be added later of course!

> As for the default, maybe using the current uncommitted changes (either staged or unstaged) is better than using a specific branch?

It also uses uncommited changes. For example: you run locally with --diff on your master branch. Stryker will get the diff beween master and master, but including uncommited changes. This also applies when checking the diff between a feature branch and master.

> We were thinking to use incremental analysis based on the PIT implementation. It basically writes state to an arbitrary file and thous the diffing itself.

We were thinking about this too! Basically we have now added the `--diff` option, but want to add the `--diffprovider` option later. That way we can easily add more ways to get the diff from. We were thinking about:
- git
- baseline (like PIT)
- filesystem (files changed since)
- maybe even more?

anhaehne:
This is my (realy rough) code that parses a git diff and returns the spans that are affected.
https://gist.github.com/anhaehne/22bcd10e5c3d67e342340cdf9ecb88a7

richardwerkman:
@anhaehne does your code expect something like:
```
diff --git a/TestApp.UnitTests/UnitTest1.cs b/TestApp.UnitTests/UnitTest1.cs
index 354535f..ab306f9 100644
--- a/TestApp.UnitTests/UnitTest1.cs
+++ b/TestApp.UnitTests/UnitTest1.cs
@@ -23,7 +23,7 @@ namespace TestApp.UnitTests
             Assert.False(a.InvokeIfNotNull());
             var wasCalled = false;
             a = () => { wasCalled = true; };
-            Assert.True(a.InvokeIfNotNull());
+            Assert.True(a.InvokeIfNotNull()); 
             Assert.True(wasCalled);
         }
```

I got this string from libgit2sharp. So maybe we could combine your script with libgit2sharp?

anhaehne:
Yes. This is just the default output of git diff. Is there a way to pass additional arguments to the diff? I added -U0 (removes the lines around the actual change that are printed for context) and --minimal (makes sure the smallest possible diff is found). These arguments make the diff faster to parse and more precise.

richardwerkman:
I'm not sure if that is possible. I've only been able to check the difference between two branches using:
``` cs
repository.Diff.Compare<Patch>(sourceBranch.Tip.Tree, DiffTargets.Index | DiffTargets.WorkingDirectory)
```

I dont see a way to pass arguments to that method. It accepts an options object but it contains no option for `minimal` or `u0`...

After looking into the [source code of libgit2sharp](https://github.com/libgit2/libgit2sharp/blob/9cdc883f26704a0c02aea8141385b700583828c3/LibGit2Sharp/Core/GitDiff.cs#L289) I see some classes for line diff info but they are marked as internal so we can't access them. It might be cool to contribute to libgit2sharp to make this info accessable for us!

anhaehne:
There is an old issue from 2017 with comment that proposes a Regex to parse the string output. 
[How to get lines modified in a commit #1490](https://github.com/libgit2/libgit2sharp/issues/1490#issuecomment-435657964)
If haven't tried it myself but if it works, it would be way simpler than my approach.
 
As a compromise i would recommend to use this as an intermediate solution until LibGit2Sharp support line diff to complete this feature in a reasonable time frame.

richardwerkman:
@anhaehne I've looked into the regex and it kinda works. However it only returns the first change of a file. Secondly, it gives very limited info about the change. Only the line that the change was placed in (starting from the namespace declaration). I can't seem to get an exact span till the end of the change.

Then the second challenge would be to map these lines to the span inside a syntax tree. We could check if the start or end line from the original code from the mutation are in the range of a change span. However I'm not sure if the lines in a syntax tree also start at the namespace declaration or at the start of the file (usings). If they start at the usings creating a mapping between changes and mutations could prove very hard.

Maybe we should just accept this PR for now and think about how we want to tackle this in a seperate issue.

src/Stryker.Core/Stryker.Core/DiffProviders/GitDiffProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public GitDiffProvider(StrykerOptions options)
+        {
+            _options = options;
+            _logger = ApplicationLogging.LoggerFactory.CreateLogger<GitDiffProvider>();
+        }
+
+        public DiffResult ScanDiff()
+        {
+            if (!_options.DiffEnabled)
```
Mobrockers: If we don't do this check, the options don't need to be passed to this class. The caller should decide whether or not to call the scandiff

src/Stryker.Core/Stryker.Core/DiffProviders/GitDiffProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+                _logger.LogWarning("Could not locate git repo. Unable to determine git diff to filter mutants.");
+            }
+            var diffResult = new DiffResult()
+            {
+                ChangedFiles = new Collection<string>(),
+                TestsChanged = false
+            };
+
+            // A git repository has been detected, calculate the diff to filter
+            using (var repository = new Repository(repositoryPath))
```
Mobrockers: What happens if there is no git repo?

anhaehne:
This pull request seems to have been abandoned. Is there a way a can help move this feature forward?

richardwerkman:
@anhaehne I don't think you have push rights on this branch. But you could pull the code and test it on some projects to see if it works fine. We have been short on time lately and gave priority to some other PRs. This PR will be merged soon and we'll continue with improving this feature.

You could also look further into how we can get the line spans for the changes. We could contribute to LibGit2Sharp and make this info available through their API. Or we could try to get the linespans with the regex you found. I was not able to get an exact linespan yet. Maybe you have some fresh insights?

src/Stryker.Core/Stryker.Core/DiffProviders/GitDiffProvider.cs
Mobrockers: We should probably cancel out of gitdiff if we cannot find a git repo to make sure we don't unexpectedly crash due to user error.

Mobrockers: As we also do if the given branch cannot be found

PR #746, opened 2019-09-27
build(integration test): Upgrade integration test to dotnet core 3

richardwerkman:
Upgrade the .net core integration test project to .net core 3.0


integrationtest/TargetProjects/NetStandardTestProject.XUnit/NetStandardTestProject.XUnit.csproj
```diff
@@ -12,7 +12,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\TargetProject\TargetProject.csproj" Properties="TargetFramework=netstandard2.0" />
```
Mobrockers: If we remove the TargetFramework property, we are not testing netstandard. We should decide whether we actually need a separate netstandard integrationtest or not. And if so, we do need to set the TargetFramework property.

integrationtest/TargetProjects/NetStandardTestProject.XUnit/NetStandardTestProject.XUnit.csproj
```diff
@@ -12,7 +12,7 @@
...
+    <ProjectReference Include="..\TargetProject\TargetProject.csproj" Properties="TargetFramework=netstandard2.1" />
```
Mobrockers: We no longer need the netstandard testproject as we removed the integrationtest

integrationtest/TargetProjects/TargetProject/TargetProject.csproj
```diff
@@ -1,9 +1,11 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;netstandard2.0</TargetFrameworks>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <LangVersion>7.3</LangVersion>
+    <NullableContextOptions>enable</NullableContextOptions>
```
Mobrockers: NullableContextOptions is the old property name and has been replaced by Nullable

Issue #667, opened 2019-08-20
Speed up pipeline with pipeline cache beta


Mobrockers:
**Is your feature request related to a problem? Please describe.**
Our builds are fairly slow, using nuget caching will speed them up a bit

**Describe the solution you'd like**
https://docs.microsoft.com/en-us/azure/devops/pipelines/caching/?view=azure-devops#netnuget

https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#enabling-lock-file

lpraz:
Azure DevOps pipeline caching is in preview at the moment. Should this issue be held off until it comes out of preview, or is it still acceptable to implement now?

Mobrockers:
I don't have any problem with trying this out while in preview.

PR #767, opened 2019-10-13
build(pipeline): Add pipeline package caching

lpraz:
Add CacheBeta task to azure-pipelines.yml to enable NuGet package caching
and hopefully speed up CI builds.
Enable creation of packages.lock.json for projects built in pipeline to
enable above.

Resolves issue #667. So far the only cache task that was added is immediately after the `dotnet build` in the "Run Sonar Analysis" job, let me know if this misses out on any packages.


Mobrockers:
The run sonar stage is only executed for non-fork builds so it won't actually be triggered by your pr. The deploy integration test stage makes more sense I think.

Mobrockers:
Now that I think about it, all stages probably should have a cache task as they are all new agents.

Also your current implementation never has a cache update so new package versions will not be cached because of the following part of the docs: 

>Note
>Caches are immutable, meaning that once a cache is created, its contents cannot be changed. See Can I clear a cache? in the Q & A section for additional details.

lpraz:
>Now that I think about it, all stages probably should have a cache task as they are all new agents.

Does this include the Sonar stage?

Mobrockers:
Yes, the sonar stage can use the cache if it's not a fork pr.

Also disregard the cache update comment, I did not realize a filepath hit on the key would actually hash the file contents and use that as the key.

Mobrockers:
I would also like libman to be cached, since the cdn we use is unreliable and results in failed builds. The libman file as well as the destination dir is found here: https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/libman.json

lpraz:
>I would also like libman to be cached

Do you know of any documentation for libman? I've only been able to find anything about clearing the cache, rather than specifying where it should be kept.

pipeline-templates/populate-cache-steps.yml
```diff
@@ -0,0 +1,11 @@
+steps:
+  - task: CacheBeta@0
+    displayName: 'Cache NuGet packages'
+    inputs:
+      key: nuget | **/packages.lock.json
+      path: $(NUGET_PACKAGES)
+  - task: CacheBeta@0
+    displayName: 'Cache Library Manager libraries'
+    inputs:
+      key: src/Stryker.Core/Stryker.Core/Reporters/HtmlReporter/Files/mutation-testing-elements.js
```
Mobrockers: I think the key should probably be similar to the nuget key, like so:

```suggestion
      key: libman | **/libman.json
```

pipeline-templates/prepare-integration-test-steps.yml
```diff
@@ -17,10 +17,17 @@ steps:
...
+
   - task: DotNetCoreCLI@2
     displayName: 'Install integrationtest dotnet tool'
     condition: eq(variables['System.PullRequest.IsFork'], 'False')
     inputs:
       command: custom
       custom: tool
       arguments: install dotnet-stryker -g --version $(IntegrationTestVersion) --add-source $(AzureArtifactFeedUri)
+
+  - task: CacheBeta@0
```
Mobrockers: Can the pipeline template not be used here?

lpraz: Sorry, missed a task earlier. Should be resolved in commit e45136f

pipeline-templates/populate-cache-steps.yml
lpraz: Resolved in commit d348a45.

lpraz:
Currently running into NU1403 (Package content hash validation failed) errors from Azure Pipelines for three packages:
- runtime.native.System.IO.Compression.4.3.0
- runtime.osx.10.10-x64.runtime.native.System.Security.Cryptography.OpenSsl.4.3.0
- runtime.ubuntu.14.04-x64.runtime.native.System.Security.Cryptography.OpenSsl.4.3.0
Tried making the cache key OS-specific (since all of the packages are native), but to no avail. Is there anything else missing from my implementation?

Mobrockers:
We seem to be hitting this issue: https://github.com/microsoft/azure-pipelines-tasks/issues/11449
Which to me suggests this is an issue with nuget. There are workarounds here:

https://github.com/NuGet/Home/issues/7921#issuecomment-478152479 <- don't add the nowarn, I don't think that is neccesary

Mobrockers:
Hey @lpraz 

It looks like $(Pipeline.Workspace)/.librarymanager is not the correct location for the libman cache. If I look at this pr on libman: https://github.com/aspnet/LibraryManager/pull/527/files it seems the location depends on the OS.. And it also seems like they are changing the cache location soon. Perhaps for now remove the libman caching so I can merge your PR and you or we can add the libman caching back when the libman cache location PR has been changed. Sounds good?

Mobrockers:
The cache location should become $(LocalAppData)/.librarymanager/cache in the future.
We could maybe fix it with the current cache location with something like this:

```
variables:
    LibManCacheWindows: $(USERPROFILE)/.librarymanager/cache
    LibManCacheNonWindows: $(HOME)/.librarymanager/cache
```

```
- task: CacheBeta@0
    displayName: 'Cache Library Manager libraries windows'
    inputs:
      key: libman | **/libman.json
      path: $(LibManCacheWindows) 
    condition: eq(variables['Agent.OS'], 'Windows_NT')
- task: CacheBeta@0
    displayName: 'Cache Library Manager libraries non windows'
    inputs:
      key: libman | **/libman.json
      path: $(LibManCacheNonWindows) 
    condition: ne(variables['Agent.OS'], 'Windows_NT')
```

lpraz:
Tried making the suggested changes, still getting this:
`libman.json : error LIB002: The "mutation-testing-elements@1.1.1" library could not be resolved by the "unpkg" provider [d:\a\1\s\src\Stryker.Core\Stryker.Core\Stryker.Core.csproj]`
Should I remove/comment out the libman caching entirely for now?

Mobrockers:
That is an expected error and is one of the reasons we want to have pipeline caching! Unpkg is very unreliable causing the build to fail.

Libman caching seems to work now. We do have another issue with nuget caching.. I'll look into it today.

Mobrockers:
There seem to be a thousand issues with cross-platform lock files and nuget..

Ex:

https://github.com/NuGet/Home/issues/8493

I'm starting to think that we can't realistically use nuget caching right now due to our cross-platform setup...

Mobrockers:
Thank you for your PR! With your starting point I managed to make it work. I was waiting on this feature very much so I went ahead and created a new branch with your changes and fixed it from there. That means that this PR is no longer needed so I'll be closing it.

If you were participating in hacktober, I do not believe closing this PR should have any impact on your t-shirt eligibility so don't worry about that.

Your help is much appreciated :)

lpraz:
Sorry for the wait, I wasnt able to get to work on my PR in the past few days. Good to hear you were able to get it sorted out, and thanks! I don't have much experience with using YAML files for Azure Pipelines, so thanks for helping me out as well!

Mobrockers:
Not a problem at all, I was just too excited to use this to wait :d

PR #785, opened 2019-11-01
feat(core3.1): Upgrade to dotnet core 3.1 LTS

richardwerkman:


Mobrockers:
Reminder: Use fancy c# 8 switch expression for ProjectComponent.CheckHealth

src/Stryker.CLI/Stryker.CLI.UnitTest/Stryker.CLI.UnitTest.csproj
```diff
@@ -7,6 +7,7 @@
 
   <PropertyGroup>
     <xunit>2.4.1</xunit>
+    <LangVersion>8.0</LangVersion>
```
Mobrockers, line 10: The LangVersion is set in our Directory.Build.Props. c#8 is default for netcore 3.1 so removing the langversion from the directory.build.props should just work

pipeline-templates/prepare-integration-test-steps.yml
```diff
@@ -1,13 +1,8 @@
 steps:
   - task: UseDotNet@2
-    displayName: 'Install dotnet core 3.0.x because macos is not up to date'
+    displayName: 'Install dotnet core 3.1.x to make dotnet tool path available in path'
```
Mobrockers, line 3: This is possibly no longer needed, this was only still needed because the macosx agent was not up to date

Issue #781, opened 2019-10-31
FTL: The name 'AppDomain' does not exist in the current context (Source code: AppDomain)


riezebosch:
**Describe the bug**

**Logs**
For us to best assist you in diagnosing the problem we would ask you to run stryker with the option `--log-file` and to attach the resulting logs to the issue.

This is broken? 
```
~ dotnet stryker -f stryker.log
Specify --help for a list of available options and commands.
```

```
Version: 0.13.0 (beta)

[14:34:44 INF] The project /Users/manuelriezebosch/git/Unmockable/Unmockable.Intercept/Unmockable.Intercept.csproj will be mutated
[14:34:46 INF] Started initial build using dotnet build
[14:34:48 INF] Initial build successful
[14:34:54 INF] Using testrunner VsTest
[14:34:54 INF] Total number of tests found: 48
[14:34:54 INF] Initial testrun started
[14:34:55 INF] Using 7536 ms as testrun timeout
[14:34:58 WRN] Stryker.NET encountered an compile error in Stryker.Core.InjectedHelpers.MutantControl.cs with message: The name 'AppDomain' does not exist in the current context (Source code: AppDomain)
[14:34:58 WRN] Safe Mode! Stryker will try to continue by rolling back all mutations in method. This should not happen, please report this as an issue on github with the previous error message.
[14:34:58 FTL] Stryker.NET could not compile the project after mutation. This is probably an error for Stryker.NET and not your project. Please report this issue on github with the previous error message.
[14:34:58 ERR] An error occurred during the mutation test run 
System.ApplicationException: Internal error due to compile error.
   at Stryker.Core.Compiling.RollbackProcess.RemoveMutantIfStatements(SyntaxTree originalTree, ICollection`1 diagnosticInfo, Boolean devMode)
   at Stryker.Core.Compiling.RollbackProcess.Start(CSharpCompilation compiler, ImmutableArray`1 diagnostics, Boolean devMode)
   at Stryker.Core.Compiling.CompilingProcess.TryCompilation(MemoryStream ms, CSharpCompilation compilation, EmitResult previousEmitResult, Boolean devMode, Int32 retryCount)
   at Stryker.Core.Compiling.CompilingProcess.Compile(IEnumerable`1 syntaxTrees, MemoryStream ms, Boolean devMode)
   at Stryker.Core.MutationTest.MutationTestProcess.Mutate()
[...]

PR #789, opened 2019-11-04
fix(coverage analysis): 'AppDomain' and named pipes are not available in all dotnet frameworks

dupdob:
Fix several incorrect log messages that where found while working on #781 . 
Note that my previous PR regarding Linq  mutators also appears as a difference, which is unexpected

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialTestProcessTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialTestProcessTests.cs
@@ -1,9 +1,8 @@
-﻿using System;
-using Moq;
+﻿using Moq;
 using Stryker.Core.Exceptions;
 using Stryker.Core.Initialisation;
-using Stryker.Core.Options;
 using Stryker.Core.TestRunners;
+using System;
 using Xunit;
 
 namespace Stryker.Core.UnitTest.Initialisation
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialTestProcessTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialTestProcessTests.cs
@@ -1,4 +1,5 @@
-﻿using Moq;
+﻿using System;
+using Moq;
 using Stryker.Core.Exceptions;
 using Stryker.Core.Initialisation;
 using Stryker.Core.Options;
@@ -23,9 +24,21 @@ namespace Stryker.Core.UnitTest.Initialisation
             testRunnerMock.Setup(x => x.RunAll(It.IsAny<int?>(), null)).Returns(new TestRunResult { Success = false });
             testRunnerMock.Setup(x => x.CaptureCoverage(false, false))
                 .Returns(new TestRunResult { Success = false });
-            testRunnerMock.Setup(x => x.DiscoverNumberOfTests()).Returns(0);
+            testRunnerMock.Setup(x => x.DiscoverNumberOfTests()).Returns(1);
 
             var exception = Assert.Throws<StrykerInputException>(() => _target.InitialTest(testRunnerMock.Object));
+        }
+
+        [Fact]
+        public void InitialTestProcess_ShouldThrowWhenNoTest()
+        {
+            var testRunnerMock = new Mock<ITestRunner>(MockBehavior.Strict);
+            testRunnerMock.Setup(x => x.RunAll(It.IsAny<int?>(), null)).Returns(new TestRunResult { Success = false });
+            testRunnerMock.Setup(x => x.CaptureCoverage(false, false))
+                .Returns(new TestRunResult { Success = false });
+            testRunnerMock.Setup(x => x.DiscoverNumberOfTests()).Returns(0);
+
+            var exception = Assert.Throws<InvalidOperationException>(() => _target.InitialTest(testRunnerMock.Object));
         }
 
         [Fact]
@@ -35,7 +48,7 @@ namespace Stryker.Core.UnitTest.Initialisation
             testRunnerMock.Setup(x => x.RunAll(It.IsAny<int?>(), null)).Returns(new TestRunResult { Success = true });
             testRunnerMock.Setup(x => x.CaptureCoverage(false, false))
                 .Returns(new TestRunResult { Success = true });
-            testRunnerMock.Setup(x => x.DiscoverNumberOfTests()).Returns(0);
+            testRunnerMock.Setup(x => x.DiscoverNumberOfTests()).Returns(2);
 
             var result = _target.InitialTest(testRunnerMock.Object);
         }
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialTestProcessTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialTestProcessTests.cs
@@ -1,6 +1,7 @@
 ﻿using Moq;
 using Stryker.Core.Exceptions;
 using Stryker.Core.Initialisation;
+using Stryker.Core.Options;
 using Stryker.Core.TestRunners;
 using Xunit;
 
@@ -20,7 +21,7 @@ namespace Stryker.Core.UnitTest.Initialisation
         {
             var testRunnerMock = new Mock<ITestRunner>(MockBehavior.Strict);
             testRunnerMock.Setup(x => x.RunAll(It.IsAny<int?>(), null)).Returns(new TestRunResult { Success = false });
-            testRunnerMock.Setup(x => x.CaptureCoverage())
+            testRunnerMock.Setup(x => x.CaptureCoverage(false, false))
                 .Returns(new TestRunResult { Success = false });
             testRunnerMock.Setup(x => x.DiscoverNumberOfTests()).Returns(0);
 
@@ -32,7 +33,7 @@ namespace Stryker.Core.UnitTest.Initialisation
         {
             var testRunnerMock = new Mock<ITestRunner>(MockBehavior.Strict);
             testRunnerMock.Setup(x => x.RunAll(It.IsAny<int?>(), null)).Returns(new TestRunResult { Success = true });
-            testRunnerMock.Setup(x => x.CaptureCoverage())
+            testRunnerMock.Setup(x => x.CaptureCoverage(false, false))
                 .Returns(new TestRunResult { Success = true });
             testRunnerMock.Setup(x => x.DiscoverNumberOfTests()).Returns(0);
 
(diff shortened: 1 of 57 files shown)


Mobrockers:
You made the changes on the linq mutator on the master branch, we then squashed the changes into one commit when accepting your merge request. Your master doesn't understand that the squash commit contains the same changes so it tries to reapply your previous commits. That's why they're showing up here.

dupdob:
This PR is ready for review

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -177,7 +178,7 @@ private IEnumerable<string> ExtractProjectFolders(ProjectAnalyzerResult projectA
             var folders = new List<string>();
             var projectDirectory = _fileSystem.Path.GetDirectoryName(projectFilePath);
             folders.Add(projectDirectory);
-            if (!fullFramework)
+           // if (!fullFramework)
```
dupdob: What is the point of this if? I disabled it while testing for compatibility with NFluent on Net 4.5 (otherwise, the project is seen as empty by Stryker).
Can you please give the reason why Stryker disregards share project for full framework? otherwise I suggest we get rid of this condition altogether.

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
```diff
@@ -167,7 +167,7 @@ private SyntaxTree RemoveMutantIfStatements(SyntaxTree originalTree, ICollection
                 if (mutationIf == null)
                 {
                     var errorLocation = diagnostic.Location.GetMappedLineSpan();
-                    _logger.LogWarning("Stryker.NET encountered an compile error in {0} with message: {1} (Source code: {2})", errorLocation.Path, diagnostic.GetMessage(), brokenMutation);
+                    _logger.LogWarning($"Stryker.NET encountered an compile error in {errorLocation.Path} (at {errorLocation.StartLinePosition.Line}:{errorLocation.StartLinePosition.Character}) with message: {diagnostic.GetMessage()} (Source code: {brokenMutation})");
```
dupdob: I added more location information in case of build error so we can provide diagnostic faster.

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
Mobrockers: At some point when shared projects were added we thought only dotnet core had shared projects. I guess we forgot to remove this check when we realized this isn't true. The if should be removed

Mobrockers:
Hey @dupdob I created a fresh branch in which I cherry-picked your actual change commits so the history makes more sense for the PR, do you mind if I force push the changes to your branch? You can see the fixed history here: #810

dupdob:
ok for the force push

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialTestProcessTests.cs
```diff
@@ -20,21 +21,33 @@ public void InitialTestProcess_ShouldThrowExceptionOnFail()
...
+            testRunnerMock.Setup(x => x.CaptureCoverage(false, false))
                 .Returns(new TestRunResult { Success = false });
-            testRunnerMock.Setup(x => x.DiscoverNumberOfTests()).Returns(0);
+            testRunnerMock.Setup(x => x.DiscoverNumberOfTests()).Returns(1);
 
             var exception = Assert.Throws<StrykerInputException>(() => _target.InitialTest(testRunnerMock.Object));
         }
 
+        [Fact]
+        public void InitialTestProcess_ShouldThrowWhenNoTest()
```
richardwerkman: Why did you add this? I think it is fine if we keep running stryker on a project with 0 tests.

src/Stryker.Core/Stryker.Core/Initialisation/InitialTestProcess.cs
```diff
@@ -31,6 +31,10 @@ public InitialTestProcess()
         public int InitialTest(ITestRunner testRunner)
         {
             var message = testRunner.DiscoverNumberOfTests() is var total && total == -1 ? "Unable to detect" : $"{total}";
+            if (total == 0)
```
Mobrockers: This is already gracefully handled by the mutationtestprocess here

 https://github.com/stryker-mutator/stryker-net/blob/cfc1e1cb3f0f9ef17bfab1d6a0a6e15cc7b01e16/src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs#L178

It will log a warning that none of the mutants were covered by a test and it will create the report with all created mutations. We do not need to throw on no tests found.

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
```diff
@@ -63,12 +67,14 @@ public class MutationTestProcess : IMutationTestProcess
...
         public void Mutate()
         {
+            var preprocessorSymbols = _input.ProjectInfo.ProjectUnderTestAnalyzerResult.DefineConstants;
+
             _logger.LogDebug("Injecting helpers into assembly.");
             var mutatedSyntaxTrees = new List<SyntaxTree>();
-            var cSharpParseOptions = new CSharpParseOptions(_options.LanguageVersion);
-            foreach (var helper in CodeInjection.MutantHelpers)
+            var cSharpParseOptions = new CSharpParseOptions(_options.LanguageVersion, DocumentationMode.None, preprocessorSymbols: preprocessorSymbols);
+            foreach (var (name, code) in CodeInjection.MutantHelpers)
```
richardwerkman: This is awesome! 🚀

dupdob:
During my tests, ‘0 test found’ situation were linked to silent failure from VsTests. 
Maybe there is a better way to report this than my aggressive throw, but we should point this to users

Envoyé de mon iPhone

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
```diff
@@ -79,12 +85,12 @@ public void Mutate()
...
                     options: cSharpParseOptions);
 
-                    _logger.LogDebug($"Mutating {file.Name}");
+                _logger.LogDebug($"Mutating {file.Name}");
                 // Mutate the syntax tree
                 var mutatedSyntaxTree = _orchestrator.Mutate(syntaxTree.GetRoot());
                 // Add the mutated syntax tree for compilation
                 mutatedSyntaxTrees.Add(mutatedSyntaxTree.SyntaxTree);
-
+                _logger.LogTrace($"Mutated {file.Name}:{Environment.NewLine}{mutatedSyntaxTree.ToFullString()}");
```
richardwerkman: Do we need this extra logging? I'm afraid that logfiles from large projects will get even more unclear thanks to these enormous blocks of code. We already log quite some code during rollbacking. Do you think this will make things more clear?

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
```diff
@@ -208,5 +214,18 @@ public StrykerRunResult Test(StrykerOptions options)
 
             return new StrykerRunResult(options, _input.ProjectInfo.ProjectContents.GetMutationScore());
         }
+
+        public TestCoverageInfos GetCoverage()
+        {
+            var (compat_noAppDomain, compat_noPipe) = _input.ProjectInfo.ProjectUnderTestAnalyzerResult.CompatibilityModes;
```
richardwerkman: The variable names `compat_noAppDomain` and `compat_noPipe` are not very clear to me.

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InputFileResolverTests.cs
```diff
@@ -52,6 +53,38 @@ public InputFileResolverTests()
 </Project>";
         }
 
+        [Fact]
+        public void ProjectAnalyzerShouldDecodeFramework()
```
richardwerkman: I think these should be seperate test cases. Maybe xUnit's `[Theory]` could be used in this case.

src/Stryker.Core/Stryker.Core/Initialisation/ProjectInfo.cs
```diff
@@ -50,27 +61,99 @@ public string AssemblyPath
...
+        {
+            get => _defineConstants ?? BuildDefineConstants();
+            set => _defineConstants = value;
+        }
+
+        private IList<string> BuildDefineConstants()
+        {
+            var constants = _analyzerResult?.GetProperty("DefineConstants")?.Split(";")?.ToList() ?? new List<string>();
+
+            var (compat_noAppDomain, compat_noPipe) = CompatibilityModes;
```
richardwerkman: These names don't make any sense to me. What do they contain?

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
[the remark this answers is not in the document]
dupdob: ok. It's just than when the number of parameters increases, interpolated strings keeps it readable.
Can revert

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
dupdob: it's not me, just C# 7 :-)

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
dupdob: I think it would be beneficial be able to access mutated code for mutation error analysis.
On the other hand, when testing, I realised that Trace entries are present when you request debug level. I think it would be better if we have a specific switch for that. Maybe correlating that with Devmode.
What to you think?

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
dupdob: ok

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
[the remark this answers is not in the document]
richardwerkman: We could also use `{errorLocation}` instead of `{0}` if that helps readability?

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
richardwerkman: During development you could also get the mutated code using debugging right? A devmode switch could be better here. If you think it is beneficial we can keep it with a switch :+1:

Mobrockers: When logging to file, all logging is written to the file. This is intended behaviour. If you mean that you see trace logging in the console when requesting debug logging then that is a bug and that should not be happening.

dupdob: Ok, added devmode has a supplemental condition

PR #810, opened 2019-11-15
781 without unsquashed commits

Mobrockers:
@dupdob this is the branch with the history fixed. As far as I can see it contains all the changes you made. Let me know if I missed anything and I'll cherry-pick those to this branch as well. I can force push this history to your branch and PR to fix your history if you want but then you should delete your local branch and check it out from github otherwise you'll merge back in the dirty history when you update your local branch with the upstream changes.

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
+++ b/src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
@@ -126,7 +126,7 @@ namespace Stryker.Core.Initialisation
         {
             string filter = BuildTestProjectFilter(basePath, testProjectNameFilter);
 
-            var projectFiles = _fileSystem.Directory.GetFileSystemEntries(basePath, filter);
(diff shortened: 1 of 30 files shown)


src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -177,7 +178,7 @@ private IEnumerable<string> ExtractProjectFolders(ProjectAnalyzerResult projectA
             var folders = new List<string>();
             var projectDirectory = _fileSystem.Path.GetDirectoryName(projectFilePath);
             folders.Add(projectDirectory);
-            if (!fullFramework)
+           // if (!fullFramework)
```
jnyrup: Intention to leave outcommented code?

Mobrockers: Nope :) Will be removed in #789

PR #830, opened 2019-11-29
feat(Multiple test projects): Add more than one test project support

### LINKED ISSUE #811: Run stryker from project under test folder instead of test folder
How stryker is designed now it is hard to merge two results into one to see the real coverage of a project that has two test projects.

In order to prepare for running stryker from a solution file we want to run on a target project instead of the test project. This makes running agains multiple test projects possible. In order to do this a big refactor is required. Most logic in stryker expects the test project to be in the current directory. 

Steps to work this out:

- [ ] Refactor all usages of `options.BasePath` where it is used as "testproject path"
- [ ] Find out if stryker is running from a test project or target project when it starts
- [ ] Read options from the target project and determine what test projects to use for testing mutations
- [ ] Refactor test project path into a list of test project paths. Pass this list to the testrunners

More steps might become clear once we start.

richardwerkman:
closes #811


Mobrockers:
Warning before merge: breaking change on command line arguments

src/Stryker.Core/Stryker.Core/Exceptions/GeneralStrykerException.cs
```diff
@@ -0,0 +1,47 @@
+using System;
+using System.Text;
+
+namespace Stryker.Core.Exceptions
+{
+    /// <summary>
+    /// Represents errors which are related to user input.
```
Mobrockers: Comments seem to be referring to input exceptions

src/Stryker.Core/Stryker.Core/Initialisation/InitialBuildProcess.cs
```diff
@@ -25,7 +25,7 @@ public InitialBuildProcess(IProcessExecutor processExecutor = null)
 
         public void InitialBuild(bool fullFramework, string projectPath, string solutionPath)
         {
-            _logger.LogInformation("Started initial build using {0}", fullFramework ? "msbuild.exe" : "dotnet build");
+            _logger.LogDebug("Started initial build using {0}", fullFramework ? "msbuild.exe" : "dotnet build");
```
Mobrockers: This means the user will be left wondering if stryker is doing anything, do we want that?

src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs
```diff
@@ -40,7 +43,13 @@ public MutationTestInput Initialize(StrykerOptions options)
             var projectInfo = _inputFileResolver.ResolveInput(options);
 
             // initial build
-            _initialBuildProcess.InitialBuild(projectInfo.ProjectUnderTestAnalyzerResult.TargetFramework == Framework.NetClassic, projectInfo.TestProjectAnalyzerResult.ProjectFilePath, options.SolutionPath);
+            int counter = 1;
+            foreach (var testProject in projectInfo.TestProjectAnalyzerResults)
```
Mobrockers: If you need a counter, I suggest you use a for loop instead of a foreach

Mobrockers:
`Argument 2: cannot convert from 'System.StringSplitOptions' to 'char' (Source code: StringSplitOptions.RemoveEmptyEntries)`

src/Stryker.Core/Stryker.Core/Initialisation/InitialBuildProcess.cs
richardwerkman: No they won't! I added a new log somewhere else. Now the output looks like this:

```
[09:47:03 INF] The project C:\Dev\ExampleProjects\TestApp\TestApp\TestApp.csproj will be mutated
[09:47:05 INF] Building test project C:\Dev\ExampleProjects\TestApp\TestApp.NUnit\TestApp.NUnit.csproj (1/2)
[09:47:09 INF] Building test project C:\Dev\ExampleProjects\TestApp\TestApp.UnitTests\TestApp.UnitTests.csproj (2/2)
[09:47:18 INF] Using testrunner DotnetTest
```

Mobrockers:
This

https://github.com/stryker-mutator/stryker-net/pull/830/files#diff-09b37202cc79fc756b5ad93b0d51a84cR201

Doesn't seem supported in full framework

Mobrockers:
Perhaps we should add multiple test projects to the integration test. We could add it as an extra phase so the total build time does not increase and so we don't lose integration testing on the default stryker use case (from testproject location).

richardwerkman:
We could also choose to run linux or mac with two test projects and windows with 1 test project (as it already is slowest). That way we also test both cases

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -121,13 +141,17 @@ private FolderComposite FindInputFiles(string path, string projectUnderTestDir,
             return folderComposite;
         }
 
-        public string FindProjectFile(string basePath, string testProjectNameFilter)
+        public string FindProjectFile(string path)
```
richardwerkman: It seems I was a little too enthusiastic with removing the testProjectNameFilter here. It is still needed in the old situation. Maybe we can add a overload for the new situation and keep the old one as it is.

PR #845, opened 2019-12-10
docs(docs/Configuration.md) Add example config option

survivorbat:
While working on a project with Stryker, I noticed that the option
ignore-methods was missing from the config, but it did work when I tried
it. So that's why I decided to add it in.


Mobrockers:
👍

PR #859, opened 2019-12-27
build(Nightly run stryker): Update the stryker dashboard every night

### LINKED ISSUE #104: Use stryker on the stryker build

richardwerkman:
closes #104

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -82,10 +82,6 @@ stages:
         custom: tool
         arguments: install dotnet-stryker -g
         includeNuGetOrg: true
-    - task: PowerShell@2
-      inputs:
-        targetType: 'inline'
-        script: 'Get-ChildItem Env:'
     - script: dotnet stryker --reporters "['dashboard', 'dots']" --dashboard-project github.com/stryker-mutator/stryker-net --dashboard-version master --dashboard-module cli
       condition: succeeded()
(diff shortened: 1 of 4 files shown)


azure-pipelines.yml
```diff
@@ -186,6 +186,38 @@ stages:
...
+        arguments: '--reporters "['dashboard', 'dots']" --dashboard-project github.com/stryker-mutator/stryker-net --dashboard-version master --dashboard-module cli'
+        workingDirectory: 'src\Stryker.CLI\Stryker.CLI.UnitTest'
+  - job: StrykerCore
+    displayName: Run Stryker.Core
+    - task: DotNetCoreCLI@2
+      displayName: Run Stryker.NET
+      inputs:
+        command: 'custom'
+        custom: 'stryker'
+        arguments: '--reporters "['dashboard', 'dots']" --dashboard-project github.com/stryker-mutator/stryker-net --dashboard-version master --dashboard-module cli'
```
Mobrockers: Module should be core

richardwerkman:
@Mobrockers I removed the condition because I want it to run in this PR to see if it works. Once it works I'll add the condition again :+1:

Mobrockers:
Perhaps it makes sense to run this with the sonar analysis (but only on timed trigger) and make that a more generic check code quality stage?

PR #898, opened 2020-01-22
feat: use buildalyzer to identify source files

### LINKED ISSUE #897: Feature request: support source file identification with buildalizer
Stryker should support alternate ways to identify the source files of the mutated project, instead of scanning the project folder(s) for .cs files. 

For example, it could use BuildAlyzer results as a reliable source.

dupdob:
Fix #897. No unit tests due to an exploration based approach.
Old scanning logic still present in case buildalyzer completely fails


Mobrockers:
> /azp run (it was worth a try)

Sorry :p You need permission on azure devops to be able to do that.

dupdob:
FYI: I have constant issues with nuget (error nu1403, hash changed), hence I have to juggle with packages.lock.json.
Any pointer on how to fix this will be very appreciated.


```
4>C:\Users\dupdob\source\repos\stryker-net\src\Stryker.Core\Stryker.Core.UnitTest\Stryker.Core.UnitTest.csproj : error NU1403: Package content hash validation failed for System.Reflection.4.3.0. The package is different than the last restore.
4>C:\Users\dupdob\source\repos\stryker-net\src\Stryker.Core\Stryker.Core.UnitTest\Stryker.Core.UnitTest.csproj : error NU1403: Package content hash validation failed for System.Reflection.Emit.4.3.0. The package is different than the last restore.
4>C:\Users\dupdob\source\repos\stryker-net\src\Stryker.Core\Stryker.Core.UnitTest\Stryker.Core.UnitTest.csproj : error NU1403: Package content hash validation failed for System.Reflection.Emit.ILGeneration.4.3.0. The package is different than the last restore.
```

Mobrockers:
@dupdob That issue is due to a bug in previous versions of dotnet core... :(

See: https://github.com/NuGet/Home/issues/7921#issuecomment-478152479

Luckily this is only an issue if you have previously contributed to stryker. Something with the hashes being calculated wrong in something called 'NugetFallbackLocation' which keeps being an issue if you have ever had packages downloaded into there..

These steps should help you fix it for your system:

```
dotnet nuget locals all --clear
git clean -xfd
git rm **/packages.lock.json -f
dotnet restore Stryker.CLI.sln
```

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -73,13 +89,107 @@ public ProjectInfo ResolveInput(StrykerOptions options)
...
                 inputFiles.Add(FindInputFiles(folder, projectUnderTestDir));
             }
-            result.ProjectContents = inputFiles;
 
-            ValidateResult(result, options);
+            return inputFiles;
+        }
 
-            return result;
+        private FolderComposite FindProjectFilesUsingBuildAlyzer(ProjectInfo info)
```
richardwerkman: Why not pass the ProjectUnderTestAnalyzerResult instead of the whole project info?

richardwerkman:
Could you add some unit tests for the InputFileResolver? It contains a lot of new code that seems important 😄

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
dupdob: redesigned to use ...result instead of the all project info struct

dupdob:
Added a unit test

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InputFileResolverTests.cs
```diff
@@ -109,6 +109,45 @@ public void InitializeShouldFindFilesRecursively()
             result.ProjectContents.GetAllFiles().Count().ShouldBe(2);
         }
 
+        [Fact]
+        public void InitializeShouldUseBuildAlyzerResult()
```
richardwerkman: Does this test actually test if the initialize uses the analyzerResult? Both on filesystem and in the analyzerResult are 2 files present. So even if the analyzerResult is not used this test will still pass. Maybe we could have a difference here between filesystem and buildalyzer.

richardwerkman:
The code looks good but the build seems to fail on linux? I'll try running again to see if it was a random encounter

richardwerkman:
I have tested your code on a project and indeed I see buildalyzer provides a set of files. I don't know why it didnt work when I tried a few weeks ago. But that is great news :tada:

dupdob:
it looks a random event, since previous run (with the weak unit test) ran fine.
It looks like failing to resolve dependencies, so there may be an issue somewhere

Mobrockers:
👏

PR #906, opened 2020-01-28
feat(parallelism): Allow more parallel testrunners than logical processors despite performance impact

### LINKED ISSUE #867: Increase the maximum number of parallel testruns
The number of parallel testruns is limited to the number of physical cores on my machine. Even when I knowingly set the number of parallel testruns higher, stryker limits to my physical cores.

I would like to set the number of cores for myself. Even if that means putting a lot of load on my processor.

**Additional context**
On build servers for example I want to push stryker to the limit. On my own machine I like to keep some processing power for working while I'm running stryker.

Mobrockers:
Fix #867


src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
```diff
@@ -273,24 +277,33 @@ private LogEventLevel ValidateLogLevel(string levelText)
             }
         }
 
-        private int ValidateConcurrentTestrunners(int maxConcurrentTestRunners)
+        private int ValidateConcurrentTestrunners(int? maxConcurrentTestRunners)
         {
+            var safeProcessorCount = Math.Max(Environment.ProcessorCount / 2, 1);
+
+            if (maxConcurrentTestRunners.HasValue)
```
richardwerkman: Is this right? If a value has been passed by the user we still use the safe value. I think we should use the safe value if the user did NOT pass a value

Mobrockers: Woops, my bad

src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
```diff
@@ -49,8 +50,10 @@ public class StrykerOptions
 
         private const string ErrorMessage = "The value for one of your settings is not correct. Try correcting or removing them.";
         private readonly IFileSystem _fileSystem;
+        private readonly ILogger _bufferLogger;
```
richardwerkman: Why is the logger called buffelLogger? We just named it logger in all other classes.

src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
```diff
@@ -273,24 +277,33 @@ private LogEventLevel ValidateLogLevel(string levelText)
...
+            if (maxConcurrentTestRunners > safeProcessorCount)
+            {
+                _bufferLogger?.LogWarning("Using {0} testrunners which is more than reccomended {1} for normal system operation. This can have an impact on performance.", maxConcurrentTestRunners, safeProcessorCount);
+            }
 
-            if (maxConcurrentTestRunners <= usableProcessorCount)
+            if (maxConcurrentTestRunners == 1)
             {
-                usableProcessorCount = maxConcurrentTestRunners;
+                _bufferLogger?.LogWarning("Stryker is running in single threaded mode due to max concurrent testrunners being set to 1.", maxConcurrentTestRunners, safeProcessorCount);
```
richardwerkman: Why are these parameters passed? They are not used in the log text

Issue #843, opened 2019-12-09
Stryker.Net does not properly capture coverage for some xUnit's theories


dupdob:
**Describe the bug**
Stryker.Net VsTest plug-in (DataCollector) reports only coverage info for the first case of some/all xUnit theories, leading to false positive: mutants flagged as non covered or survivors.

**Workaround**
One can disable coverage analysis to get reliable results if the issue appears, using the following command line option:
```
-ca off
```

**Expected behavior**
Stryker should be able to properly capture coverage of any test.

**Additional context**
problem identified through issue #842

dupdob:
Preliminary analysis hints that Stryker.DataCollector receives test event only one for the all theory. Not sure if it encompasses the whole theory or only the first test case.

richardwerkman:
Do you think this can only occur in xUnit or in other test frameworks as well?

dupdob:
I definitely need further analysis to reach a conclusion. As of now, I think this is very specific as:
- I am pretty sure NFluent uses some TestCase (NUnit feature similar to XUnit's Theory) and I do not have any coverage issue with NFluent
- many projects use XUnit Theory and it is the first time this issue is reported.

What raised suspicions regarding coverage with issue #842 is the fact that the test project provides TheoryData factories, which is pretty specific.

dupdob:
Analysis update: turning on 'perTestInIsolation' coverage analysis mode, revealed that Stryker launch two test sessions, but 3 and 9 results instead of the expected one!.

dupdob:
Btw, it looks a lot like it is related to this:
https://github.com/xunit/xunit/issues/1669

dupdob:
Log analysis confirm first impression:
the xUnit visual studio runner treat (some?) theories as a single test (from vstest perspective).
It returns results for all test cases, but notifications to datacollector only happens on the first testcase.
This  kinda sucks, but should not come as no surprise when you see the comments on the xUnit issue I pointed to earlier.
I will look into workarounds/options to control that.

dupdob:
# Analysis
I reviewed VsTest and xUnit code. I am afraid I was not able to fully reverse engineer xUnit behaviour (due to design complexity), but here is my understanding of the situation:
1. When xUnit (testrunner) **discovers** the tests, there is a special logic for **Theories**:
    1. If all Theory data can be discovered at analysis time, xUnit report one test per set of data
[...]

Issue #814, opened 2019-11-18
System.Linq.Expressions.Expression mutations result in untestable mutations


rouke-broersma:
## Issue: Significant discrepancy in mutation testing result when disabling coverage analysis on [Unmockable](https://github.com/riezebosch/Unmockable)
All mutants (37 at time of testing) are killed when turning coverage analysis off whereas 6 are uncovered and 3 survived with coverage analysis on.

## Cause: Unexpected impact of mutant control logic
**[Unmockable](https://github.com/riezebosch/Unmockable)** extensively uses the [System.Linq.Expressions.Expression](https://docs.microsoft.com/en-us/dotnet/api/system.linq.expressions.expression?view=netframework-4.8) class in method signature. **StrykerMutator.Net** relies on an in-place mutation for expressions, using ternary operators to enable them when needed.
Therefore some mutated expressions are **not evaluated** at runtime, but simply passed as an expression tree to **Unmockable** code. 
As a result, mutant in **non evaluated expressions**:
- are invisible during coverage analysis  ⇒  flagged as not covered
- can't be selectively activated  ⇒  can't be tested 
- result mutated code that is active during every test sessions  ⇒ Stryker.Mutator results are wrong with or without coverage analysis.

To confirm that last statement, one simply has to manually introduce a mutant and run the code. For example, this mutant breaks no test and should be marked as survived 
```csharp
namespace Unmockable.Matchers
{
  internal static class ArgMatcherFactory
  {
    public static IArgumentMatcher? Create(Expression arg) => arg switch
      {
          MethodCallExpression call when call.Method.DeclaringType == typeof(Arg) => call.Method.Name switch
          {
              nameof(Arg.Ignore) => (IArgumentMatcher)new IgnoreArgument(),
              // here we have replaced Arguments.Single() with Arguments.SingleOrDefault()
              nameof(Arg.Where) => new LambdaArgument(call.Arguments.SingleOrDefault()),
              _ => throw new InvalidOperationException()
          },
          _ => null
      };
   }
}
```
But it is marked as *'not covered'* with coverage analysis and as *'killed'* when coverage analysis is off. Please note that having it marked as *'killed'* means it is very likely that other mutants are flagged as *'killed'* while they are actual survivors, invalidating the test session's results.

## Workaround
The only workaround is to use the ignore methods option (see #646) to disable mutation for each call to method expecting System.Linq.Expressions.Expression as parameter(s).

## Fix
[...]

Issue #519, opened 2019-05-17
Progress bar appending instead of replacing values inline (Macos 10.14.5)


Kevintjeb:
**Describe the bug**
Currently when running dotnet stryker, the reporter does not replace values inline in the terminal. It appends to it. This makes it look like the following:

**Logs**
[10:54:19 INF] Started initial build using dotnet build
[10:54:24 INF] Initial build successful
[10:54:24 INF] Using testrunner DotnetTest
[10:54:24 INF] Initial testrun started
[10:54:29 INF] Total number of tests found in initial test run: 56
[10:54:29 INF] Initial testrun successful in 4518 ms
[10:54:29 INF] Using 36777 ms as testrun timeout
[10:54:32 INF] 53 mutants ready for test

53 mutants have been created. Each mutant will now be tested, this could take a while.

Tests progress | ---------- | 0 / 53 | 0 % | NA |
Killed : 0
Survived: 0
Timeout : 0

Tests progress | ---------- | 1 / 53 | 1 % | ~5m 58s |

Killed : 1

Tests progress | ---------- | 2 / 53 | 3 % | ~2m 56s |

Killed : 2

Tests progress | ---------- | 3 / 53 | 5 % | ~1m 55s |

Survived: 1

Tests progress | ---------- | 4 / 53 | 7 % | ~1m 25s |

**Expected behavior**
I expect the output to be replaced, and thus having an animating progress bar.

**Desktop:**
 - OS: Macos
-  Terminal: Iterm2, and integrated terminal. Using the ZSH framework.
 - Type of project: core
 - Framework Version: core 2.2
 - Stryker Version:  beta 0.11.0

Mobrockers:
When we cannot determine if we are dealing with an interactive console, we disable replacing the progress bar inline. See: 
https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/Reporters/Progress/ConsoleOneLineLogger.cs#L27

And

https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/Reporters/Progress/ConsoleOneLineLogger.cs#L62

Our check is pretty crude so it could very well be something else is crashing on MacOS which causes us to treat your terminal as non-interactive.

richardwerkman:
We could look into https://www.nuget.org/packages/ShellProgressBar/# as an alternative for our own progress bar. ShellProgressBar claims to be .net core ready and to run on macos.

[...]

PR #936, opened 2020-02-11
feat(coverage analysis): Use coverage analysis to determine mutations that can be active at the same time

dupdob:
This PR support testing multiple mutants in a single test run.
Still work in progress at this stage, this draft PR is open so people can appreciate the impact it will induce


dupdob:
Just for information. This is not fully working as coverage capture appears broken. I will obviously post updates when I make progress

dupdob:
Issue has been fixed, I am testing against NFluent. If it works, next step will be reviewing the design.

dupdob:
Test against NFluent gave me weird results: the progress reporter told me that all tested mutants were killed, but 9 were not tested.
But the dashboard reported an abysmal 68% result. I assume I broke something in the reporting logic I guess, but it is weird that different reporters give different results.
I'll keep you posted

Mobrockers:
That is weird...

dupdob:
I found the issue: there was a inversion between failed tests and ran tests for real time reporting, leading to a certain 100% kill ratio.
I have fixed it, and indeed, I have many undue survivors. Proving there is something fishy somewhere.

dupdob:
I had an epiphany last night: the problem results from the fact that now the active mutant can be changed before each test is run. But the current control logic cache the environment variable value on first access. 


I will remove caching and confirm this fix this problem

dupdob:
Issue is fixed when removing the cache. When my run on NFluent is finished with correct result, I will update this PR so anybody can test it.
Next steps are:
  - fix any remaining issue
  - update this PR
 - review the design and refactor where relevant
 - design a new caching strategy for the active mutant to remove the performance hit due to constantly reading the environment variables.

dupdob:
Still WIP. This update is to show the progress of this effort
Main changes:
- got rid of environment variables for active mutant control. Use reflection instead. This logic can (and will) be extended to coverage result as well
- fixed some issues regarding handing of 'static' mutants. some issues are remaining
The biggest remaining issue is that it appears that only ~90% of the mutants are tested. Probably some logic flow when batching mutants in groups.

Overall: getting rid of the environment variable will give a significant performance boost. Multiple mutants testing results in cutting the test run time in half for NFluent (50 minutes down from 2 hours)

dupdob:
The problem of 90% tested mutants was a reporting issue (i.e. mutants were tested but not reported when the test sessions timed out).

One remaining issue: on NFluent, I have three undue survivors (i.e. they are killed when using the latest Stryker version)

dupdob:
The three survivors comes from partial failure from coverage analysis. They relate to lazy initialisation of a static cache => once the correct value is loaded in the cache, mutation has no effect.
Need to think if there is a work around this, but I am afraid detection is getting harder and harder

dupdob:
Behaviour appears ok at this stage. sill some redesign/refactoring to do, mostly removing the usage of environment variables for coverage capture as well as mutant control.
I also need to do some regression tests with dotnet nest to ensure I have not broken anything there (it is likely)

src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
```diff
@@ -137,6 +137,16 @@ public static class CLIOptions
             JsonKey = "abort-test-on-fail"
         };
 
+        public static readonly CLIOption<bool> TestMultipleMutants = new CLIOption<bool>
+        {
+            ArgumentName = "--test-multiple-mutants",
```
richardwerkman: Maybe the name `--overlap-mutants` describes this feature better. As we already test multiple mutants at the same time. But this PR adds the ability to overlap mutants in the same testrun.

dupdob: I don't like the name _test-multiple-mutants_ either. But _overlap-mutants_ sounds strange to me.
Thinking about it, what about using reverse naming, such as _test-each-mutant-in-isolation_, or something similar.
Ultimately, we should test mutants in batches for performance reason, and this feature comes with no anticipated limitation (vs testing mutant one after the other).

Mobrockers: I think we should treat it as a temporary inverse feature flag and to give the feature a name we can live with like mix-mutations. So the flag would become `--disable-mix-mutations` or whatever other name we come up with. If we start naming features like these in such a way then we can also start removing them easily later. There shouldn't be any reason to turn this off unless it doesn't work.

I don't think the name is that important since I don't think we should plan to have this as an option for very long. As long as we have a good description people will find the flag if they need it.

dupdob: I propose `--disable-testing-mix-mutations`

dupdob:
Wel
I think this PR is ready for review. Thanks in advance for your time doing this!

riezebosch:
[Archive.zip](https://github.com/stryker-mutator/stryker-net/files/4275522/Archive.zip)
Here are results from my PR #964 and with your PR merged. The results look promising, the percentages are the same and the runtime reduced from `00:04:29.2866302` to `00:02:30.5456240`. One difference is though that on master my excluded mutants are reported as `no coverage` and on your branch as `survived`.

Had to resolve some merge conflicts also.

riezebosch:
If I run it on https://github.com/riezebosch/Unmockable/tree/master/Unmockable.Intercept.Tests with your branch and `-ca off` I hit a `NullReferenceException`:

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Stryker.Core.TestRunners.VsTest.VsTestRunner.TestMultipleMutants(Nullable`1 timeoutMs, IReadOnlyList`1 mutants, TestUpdateHandler update) in /Users/manuelriezebosch/git/stryker-net/src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs:line 90
   at Stryker.Core.TestRunners.VsTest.VsTestRunner.RunAll(Nullable`1 timeoutMs, Mutant mutant, TestUpdateHandler update) in /Users/manuelriezebosch/git/stryker-net/src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs:line 83
   at Stryker.Core.TestRunners.VsTest.VsTestRunnerPool.TestMultipleMutants(Nullable`1 timeoutMs, IReadOnlyList`1 mutants, TestUpdateHandler update) in /Users/manuelriezebosch/git/stryker-net/src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunnerPool.cs:line 51
   at Stryker.Core.TestRunners.VsTest.VsTestRunnerPool.RunAll(Nullable`1 timeoutMs, Mutant mutant, TestUpdateHandler update) in /Users/manuelriezebosch/git/stryker-net/src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunnerPool.cs:line 61
   at Stryker.Core.Initialisation.InitialTestProcess.InitialTest(ITestRunner testRunner) in /Users/manuelriezebosch/git/stryker-net/src/Stryker.Core/Stryker.Core/Initialisation/InitialTestProcess.cs:line 43
   at Stryker.Core.Initialisation.InitialisationProcess.InitialTest(StrykerOptions options, Int32& nbTests) in /Users/manuelriezebosch/git/stryker-net/src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs:line 74
   at Stryker.Core.StrykerRunner.RunMutationTest(StrykerOptions options, IEnumerable`1 initialLogMessages) in /Users/manuelriezebosch/git/stryker-net/src/Stryker.Core/Stryker.Core/StrykerRunner.cs:line 82
```

dupdob:
@riezebosch : thanks for the feedback. I pushed a commit which fixes both issues:
- NRE was due to a last (post testing) ill advised refactoring
- loss of `no coverage`: the feature has been lost during redesign.

the NRE is definitely fixed, but I would love confirmation for the `no coverage` issue (note to self: add a unit test for this)

Note that `coverage analysis` is required for this feature to work.
The issue you encounter with `Unmockable` comes from an issue in the xUnit test runner and runtime generated test cases.

dupdob:
Last commit improves on the rollback process by triggering safe mode only when Stryker did not remove any mutations for the source files, in case the error message was a side effect of another error. As an attempt to keep as much mutations as possible

dupdob:
@riezebosch : It turns out that I lost the workaround for xUnit runtime test cases during refactoring. I restored it and pushed a new commit. You should get correct results on `Unmockable`. Note that Stryker will not be able to use multiple mutation tests due to the dynamic tests cases. You should get a warning stating that when you run this branch

riezebosch:
@dupdob I'm not entirely sure if the problems with Unmockable are related to the dynamic test cases problem present in the xUnit runtime since I don't have any `[Theory]` used.

dupdob:
Sorry, it turns out I was thinking about a different project.
Please take a look at issue #843 and see if it relates to your problems, and consider opening an issue if it does not. Starting with this PR, any `Failed to retrieve coverage info` warning relates to the test engine failing to report test case(s) appropriately

riezebosch:
The issue with Unmockable was identified before https://github.com/stryker-mutator/stryker-net/issues/781 and current problems are captured in this issue https://github.com/stryker-mutator/stryker-net/issues/814. So you've indeed seen it before 😎 .

dupdob:
Thanks for the reminder. So, yes, not much we can do for this problem then

src/Stryker.CLI/Stryker.CLI/OptionsBuilder.cs
```diff
@@ -37,6 +37,7 @@ public OptionsBuilder(ILogger logger)
             CommandOption coverageAnalysis,
             CommandOption abortTestOnFail,
             CommandOption configFilePath,
+            CommandOption runMultipleMutations,
```
Mobrockers:
```suggestion
            CommandOption disableSimultaneousTesting,
```

src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
```diff
@@ -51,6 +51,7 @@ public int Run(string[] args)
             var devMode = CreateOption(app, CLIOptions.DevMode);
             var coverageAnalysis = CreateOption(app, CLIOptions.CoverageAnalysis);
             var abortTestOnFailParam = CreateOption(app, CLIOptions.AbortTestOnFail);
+            var testMultipleMutations = CreateOption(app, CLIOptions.DisableTestingMix);
```
Mobrockers:
```suggestion
            var disableSimultaneousTesting = CreateOption(app, CLIOptions.DisableTestingMix);
```

src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
```diff
@@ -92,6 +93,7 @@ public int Run(string[] args)
                     coverageAnalysis: coverageAnalysis,
                     abortTestOnFail: abortTestOnFailParam,
                     configFilePath: configFilePathParam,
+                    runMultipleMutations: testMultipleMutations,
```
Mobrockers:
```suggestion
                    disableSimultaneousTesting: disableSimultaneousTesting,
```

src/Stryker.CLI/Stryker.CLI/OptionsBuilder.cs
dupdob: thanks for the fixes. it seems I forgot some along the way, indeed

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
```diff
@@ -230,7 +258,7 @@ private void ParseResultsForCoverage(IEnumerable<TestResult> testResults)
                     var propertyPairValue = (value as string);
                     if (string.IsNullOrWhiteSpace(propertyPairValue))
                     {
-                        CoverageMutants.DeclareMappingForATest(testResult.TestCase, new List<int>(), new List<int>());
+                        //
```
Mobrockers: Is there supposed to be nothing here>

src/Stryker.Core/Stryker.Core/Stryker.Core.csproj
```diff
@@ -72,4 +72,9 @@
   <ItemGroup>
     <ProjectReference Include="..\..\Stryker.DataCollector\Stryker.DataCollector\Stryker.DataCollector.csproj" />
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Reporters\TestStatisticsReporter\" />
```
Mobrockers: This shouldn't be neccesary on dotnet core, all source files are automatically included in the project

src/Stryker.Core/Stryker.Core/Reporters/Progress/ConsoleOneLineLogger.cs
```diff
@@ -33,23 +33,31 @@ public void StartLog(string text, params object[] args)
...
                 ClearLine();
-
                 Console.SetCursorPosition(0, _cursorTop);
             }
 
             _logger.LogInformation(text, args);
 
             if (_hasConsole)
             {
+                if (Console.BufferWidth > 0)
```
Mobrockers: Is this a fix for #519 ?

src/Stryker.Core/Stryker.Core/Reporters/TestStatisticsReporter/TestStatisticsReporter.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Linq;
+using Stryker.Core.Mutants;
+using Stryker.Core.Options;
+using Stryker.Core.ProjectComponents;
+using Stryker.Core.Reporters.Json;
+using Stryker.Core.Testing;
+
+namespace Stryker.Core.Reporters
+{
+    public class TestStatisticsReporter: IReporter
```
Mobrockers: Is this reporter registered as a reporter and added to the list of reporter options?

src/Stryker.Core/Stryker.Core/Initialisation/ProjectInfo.cs
```diff
@@ -21,6 +21,8 @@ public class ProjectInfo
         /// </summary>
         public FolderComposite ProjectContents { get; set; }
 
+        public string MutantControlNameSpace { get; set; }
```
Mobrockers: The MutantControlNameSpace is not part of the project the user is running stryker on so ProjectInfo does not seem like the correct place for this property.

src/Stryker.CLI/Stryker.CLI/OptionsBuilder.cs
```diff
@@ -78,6 +79,7 @@ public OptionsBuilder(ILogger logger)
                 maxConcurrentTestRunners: GetOption(maxConcurrentTestRunners.Value(), CLIOptions.MaxConcurrentTestRunners),
                 coverageAnalysis: GetOption(coverageAnalysis.Value(), CLIOptions.CoverageAnalysis),
                 abortTestOnFail: GetOption(abortTestOnFail.HasValue(), CLIOptions.AbortTestOnFail),
+                disableSimultaneousTesting: GetOption(runMultipleMutations.HasValue(), CLIOptions.DisableTestingMix),
```
Mobrockers:
```suggestion
                disableSimultaneousTesting: GetOption(disableSimultaneousTesting.HasValue(), CLIOptions.DisableTestingMix),
```

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
```diff
@@ -107,17 +107,20 @@ public RollbackProcessResult Start(CSharpCompilation compiler, ImmutableArray<Di
 
         private static SyntaxNode FindEnclosingMember(SyntaxNode node)
         {
-            if (node.Kind() == SyntaxKind.MethodDeclaration || node.Kind() == SyntaxKind.GetAccessorDeclaration || node.Kind() == SyntaxKind.SetAccessorDeclaration)
+            for(;;)
```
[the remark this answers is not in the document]
dupdob: none but habit/cosmetic reason. I find `for(;;)`to be more elegant than `while(true)`

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
```diff
@@ -107,17 +107,20 @@ public RollbackProcessResult Start(CSharpCompilation compiler, ImmutableArray<Di
...
-            {
-                return null;
-            }
+                if (node.Parent == null)
+                {
+                    return null;
+                }
 
-            return FindEnclosingMember(node.Parent);
+                node = node.Parent;
```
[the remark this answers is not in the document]
dupdob: if you are more confortable me using an intermediate variables for readability reason, no problem.
Here, I am merely doing tail recursion optimisation which leads to some performance increase as we get rid of stack allocation and parameter passing

src/Stryker.CLI/ToRemove.cs
```diff
@@ -0,0 +1,858 @@
+﻿//----------------------
```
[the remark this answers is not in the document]
dupdob: I have no recollection of why this file is here.
Will remove

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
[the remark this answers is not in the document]
dupdob: I have no recollection of why this file is here.
Will remove

src/Stryker.Core/Stryker.Core/Reporters/Progress/ConsoleOneLineLogger.cs
dupdob: No. it turns out that Console.BufferWidth returns 0 on Linux to mark it as not supported.
The overall logic is to fix a race condition leading to a crash when you resize the console window

src/Stryker.Core/Stryker.Core/Initialisation/ProjectInfo.cs
dupdob: ProjectInfo is the only structure that is used all along the workflow. So the alternative is to create another structure to store some variables that relates to the current mutation session.
Note that this could be a project dependant variables, since its role is to avoid collision with project existing namespaces.

src/Stryker.Core/Stryker.Core/Initialisation/EmbeddedResourcesGenerator.cs
```diff
@@ -59,8 +59,9 @@ private static ModuleDefinition LoadModule(string assemblyPath, ILogger logger)
 
         private static IEnumerable<ResourceDescription> ReadResourceDescriptionsFromModule(ModuleDefinition module)
         {
-            foreach (EmbeddedResource moduleResource in module.Resources.Where(r => r.ResourceType == ResourceType.Embedded))
+            foreach (var resource in module.Resources.Where(r => r.ResourceType == ResourceType.Embedded))
```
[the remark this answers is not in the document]
dupdob: reverted

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
dupdob: Added log to clarify what is (not) happening here

src/Stryker.Core/Stryker.Core/Stryker.Core.csproj
dupdob: it was added by VStudio, does not know why...

dupdob: removed

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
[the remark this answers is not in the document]
dupdob: Refactored to a more readable implementation

src/Stryker.Core/Stryker.Core/Reporters/TestStatisticsReporter/TestStatisticsReporter.cs
dupdob: At this state, it is experimental. I used it to assess the interest of batching mutants. It is a useful tool that can lead to many interesting findings. But not for now. 
It should not have been part of this PR.
Removed

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
```diff
@@ -153,18 +156,38 @@ private void DumpBuildErrors(KeyValuePair<SyntaxTree, ICollection<Diagnostic>> s
...
-                if (mutationIf == null)
+                if (mutationIf != null)
+                {
+                    if (brokenMutations.Contains(mutationIf))
+                    {
+                        continue;
+                    }
+
+                    brokenMutations.Add(mutationIf);
+                    phaseOneFixFound = true;
```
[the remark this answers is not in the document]
dupdob: renamed to `rollbackedAtLeastOneMutation`

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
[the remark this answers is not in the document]
Mobrockers: I like the refactor much better, and also addresses the empty for :)

src/Stryker.Core/Stryker.Core/Initialisation/ProjectInfo.cs
Mobrockers: Mhm I agree that it seems like the only suitable place because it is the only thing move all along the workflow, but it is distinctly not something part of the original project and rather something that we generated to be able to execute the mutation test. Will need to think about this one with @richardwerkman 

Introducing a new structure just for this one property also seems unfortunate..

Mobrockers:
I'm getting weird results testing this branch on full framework:

```
[08:07:33 INF] Initializing test runners (VsTest)
**[08:07:35 ERR] Runner 0: Test discovery has been aborted!**
[08:07:36 INF] Test runners are ready
**[08:07:36 INF] Total number of tests found: 0**
[08:07:36 INF] Initial testrun started
[08:07:36 INF] Using 5157 ms as testrun timeout
[08:07:36 INF] Generating mutants.
[08:07:43 INF] 4 mutants got status CompileError. Reason: Could not compile
[08:07:43 INF] 178 mutants got status Ignored. Reason: Removed by exclude from code coverage filter
[08:07:43 INF] **725 mutants ready for test**
[08:07:43 INF] Capture mutant coverage using 'perTest' mode.
[08:07:43 INF] Analyze coverage info to test multiple mutants per session.
[08:07:43 WRN] Unable to test more than one mutation per test run.

Testing mutant | ██████████ | **13 / 13** | 100 % | ~0m 00s |
Killed : 0
Survived: 13
Timeout : 0
```

dupdob:
@Mobrockers : 
short answer: it is likely there is a bug with full framework, probably linked to reflection failing somewhere or an appdomain related issue. I intended to perform some exploratory testing but I kina forgot. Was hoping integration tests were a sufficient safety net.

Visibly not. Will look into it.

dupdob:
It looks more trivial than that ` Microsoft.VisualStudio.TestPlatform.ObjectModel.SettingsException: Paramètres non valides 'RunConfiguration'. Valeur non valide '.NETFramework,Version= v4.6.1' spécifiée pour 'TargetFrameworkVersion'.`
It looks '.NETFramework,Version= v4.6.1' is not a valid value for `TargetFrameworkVersion'

It turns out that MS documentation conflicts on that topic:
Visual Studio 2019 doc refers to [well-known names](https://docs.microsoft.com/en-us/visualstudio/test/configure-unit-tests-by-using-a-dot-runsettings-file?view=vs-2019).
VsTest doc refers to a [structured naming logic](https://github.com/microsoft/vstest-docs/blob/master/docs/configure.md)

Do we have any good reason to explicit the framework version, as VsTest is supposed to detect it automatically?

dupdob:
It turns out the fix was simple: just use `Major.Minor` format for the version string. If you `patch` it fails.

dupdob:
I pushed the fix

src/Stryker.Core/Stryker.Core/Initialisation/ProjectInfo.cs
dupdob: What if generating the random namespace was part of the 'project analysis' phase?

Mobrockers:
> Do we have any good reason to explicit the framework version, as VsTest is supposed to detect it automatically?

Without specifying framework version VsTest did not work in the past, that's why we explicitly set it.

Mobrockers:
The reporting in the console is still weird:

```
[14:32:13 INF] Initializing test runners (VsTest)
[14:32:18 INF] Test runners are ready
[14:32:18 INF] Total number of tests found: 179
[14:32:18 INF] Initial testrun started
[14:32:23 INF] Using 11987 ms as testrun timeout
[14:32:23 INF] Generating mutants.
[14:32:29 INF] 4 mutants got status CompileError. Reason: Could not compile
[14:32:29 INF] 178 mutants got status Ignored. Reason: Removed by exclude from code coverage filter
[14:32:29 INF] 725 mutants ready for test
[14:32:29 INF] Capture mutant coverage using 'perTest' mode.
[14:32:35 INF] Analyze coverage info to test multiple mutants per session.
[14:32:35 WRN] Unable to test more than one mutation per test run.

Testing mutant | ██████████ | 13 / 13 | 100 % | ~0m 00s |
Killed : 12
Survived: 1
Timeout : 0
```

A lot of mutants are getting reported as no-coverage when previously they were covered. It also looks like there is something wrong with the text spans in the mutants because the html report is displaying mutations on the wrong lines, but I don't know if that's introduced by this PR or by another change since latest release. I'll get back to you on the reporting by comparing to latest non-released master.

Old:
image


New:
image

dupdob:
Indeed. Not normal. 
Can you share the log ?

Envoyé de mon iPhone

dupdob:
I reproduce the issue with NFluent on Net Full. Not analysed yet.

dupdob:
update: as anticipated, there is an issue when using reflection. Digging into this...

src/Stryker.Core/Stryker.Core/Initialisation/ProjectInfo.cs
Mobrockers: It would change the scope of ProjectInfo to something beyond just info about the project we'll be testing, but perhaps it's the best option

Mobrockers:
Do you still need the log from my project? I can provide it in private if neccesary.

dupdob:
no, let me fix this and we'll see if you still have the issue

dupdob:
I pushed a workaround (disabling AppDomain when used with Full framework).
Proper fix requires having a DataCollector version dedicated to Full framework and having the required cross Appdomain logic; which requires accessing the Runtime host of the process among other niceties.

Simply disabling AppDomain sounds smarter as it has only limited drawbacks:
- test process will be slightly more fragile as appdomains provide some extra protection for some low level exceptions (mostly around assembly loading). In practice, it means that some nasty mutation may lead to hard crash of the vstest process in rare occurence.
- some ill designed test suites may assume running in a dedicated appdomain and having an extra assembly may trigger some bugs.

I propose to keep the workaround for this release and reevaluate the situation if someone run into some issues.

dupdob:
@Mobrockers : please test again, I expect your problem to be fixed.

dupdob:
I made a run on NFluent for Net 4.5 and got the expected results. Problem is fixed

Mobrockers:
Detection is now correct, but the result is unexpected. Only 13 mutants are killed:

```
[07:05:13 INF] Initializing test runners (VsTest)
[07:05:21 INF] Test runners are ready
[07:05:21 INF] Total number of tests found: 179
[07:05:21 INF] Initial testrun started
[07:05:25 INF] Using 11255 ms as testrun timeout
[07:05:25 INF] Generating mutants.
[07:05:31 INF] 4 mutants got status CompileError. Reason: Could not compile
[07:05:31 INF] 178 mutants got status Ignored. Reason: Removed by exclude from code coverage filter
[07:05:31 INF] 725 mutants ready for test
[07:05:31 INF] Capture mutant coverage using 'perTest' mode.
[07:05:35 INF] Analyze coverage info to test multiple mutants per session.
[07:05:35 INF] Mutations will be tested in 65 test runs, instead of 458.

Testing mutant | ██████████ | 458 / 458 | 100 % | ~0m 00s |
Killed : 13
Survived: 445
Timeout : 0
```

I checked the 13 mutants and they are not in a static context.

dupdob:
Issues with NetClassic and MsTest/ASP.Net have been fixed. 
Some work remains to restore partial support for net core 1.x. Partial as in no support for coverage analysis. 

This is due to the full redesign of StrykerDataCollector that brings improved performance and more reliable results but it relies on the AppDomain class

src/Stryker.Core/Stryker.Core.UnitTest/Compiling/RollbackProcessTests.cs
```diff
@@ -328,6 +326,11 @@ public string AddTwoStrings(string first, string second, out string third)
 
                 var rollbackedResult = fixedCompilation.Compilation.Emit(ms);
 
+                rollbackedResult.Success.ShouldBeFalse();
+                rollbackedResult.Diagnostics.ShouldHaveSingleItem();
+
+                fixedCompilation = target.Start(fixedCompilation.Compilation, rollbackedResult.Diagnostics, false,false);
+                rollbackedResult = fixedCompilation.Compilation.Emit(ms);
```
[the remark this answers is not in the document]
dupdob: yes. that is exactly the purpose of this test: verify that safe mode kicks in when no mutation has been rolled back

src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/CoverageCaptureTests.cs
```diff
@@ -28,74 +29,28 @@ private static bool WaitFor(object lck, Func<bool> predicate, int timeout)
...
-                    WaitFor(lck, () => !string.IsNullOrEmpty(message), 1000).ShouldBeTrue("Client message received event was not raised");
-                    message.ShouldBe("it works");
-                }
-            }
+            var referenceConf="<Parameters><Environment name=\"ActiveMutant\" value=\"1\"/></Parameters>";
+            var node = new XmlDocument();
+
+            node.LoadXml(referenceConf);
+
+            node.ChildNodes.Count.ShouldBe(1);
```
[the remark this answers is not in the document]
dupdob: ShouldHaveSingleItem is not available for this type.

src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/CoverageCaptureTests.cs
```diff
@@ -28,74 +29,28 @@ private static bool WaitFor(object lck, Func<bool> predicate, int timeout)
...
+            var coolChild = node.GetElementsByTagName("Parameters");
+            coolChild[0].Name.ShouldBe("Parameters");
+            var envVars = node.GetElementsByTagName("Environment");
+
+            envVars.Count.ShouldBe(1);
+
+//            coolChild.ChildNodes.Count.ShouldBe(1);
```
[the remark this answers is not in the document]
dupdob: done

src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/CoverageCaptureTests.cs
```diff
@@ -28,74 +29,28 @@ private static bool WaitFor(object lck, Func<bool> predicate, int timeout)
...
-                    message.ShouldBe("it works");
-                }
-            }
+            var referenceConf="<Parameters><Environment name=\"ActiveMutant\" value=\"1\"/></Parameters>";
+            var node = new XmlDocument();
+
+            node.LoadXml(referenceConf);
+
+            node.ChildNodes.Count.ShouldBe(1);
+            var coolChild = node.GetElementsByTagName("Parameters");
```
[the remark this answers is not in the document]
dupdob: yep, guilty of senseless copy paste :-)

src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/CoverageCaptureTests.cs
```diff
@@ -28,74 +29,28 @@ private static bool WaitFor(object lck, Func<bool> predicate, int timeout)
...
+            coolChild[0].Name.ShouldBe("Parameters");
+            var envVars = node.GetElementsByTagName("Environment");
+
+            envVars.Count.ShouldBe(1);
```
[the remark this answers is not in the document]
dupdob: does not work

src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/CoverageCollectorShould.cs
```diff
@@ -0,0 +1,64 @@
...
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.DataCollection;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.DataCollector.InProcDataCollector;
+using Moq;
+using Shouldly;
+using Stryker.DataCollector;
+using Xunit;
+
+namespace Stryker.Core.UnitTest.TestRunners
+{
+    public static class MutantControl
```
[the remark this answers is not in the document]
dupdob: added a descriptive comment. it is used to perform unit testing around the control logic implemented in DataCollector class.
`mock for the actual MutantControl class injected in the mutated assembly.`

src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/DotnetTestRunnerTests.cs
```diff
@@ -32,12 +32,12 @@ public void RunAllExitCode1SuccessShouldBeFalse()
...
-            string path = FilePathUtils.NormalizePathSeparators("c://test");
-            var target = new DotnetTestRunner(path, processMock.Object, OptimizationFlags.NoOptimization, new string[] { "C://test//mytest.dll" });
+            var path = FilePathUtils.NormalizePathSeparators("c://test");
+            var target = new DotnetTestRunner(path, processMock.Object, OptimizationFlags.NoOptimization, new[] { "C://test//mytest.dll" });
 
-            var result = target.RunAll(null, null);
+            var result = target.RunAll(null, null, null);
 
-            Assert.False(result.Success);
+            Assert.Equal(1, result.FailingTests.Count);
```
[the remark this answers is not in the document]
dupdob: again, not available. used Shouldly instead of xUnit assert

src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/DotnetTestRunnerTests.cs
```diff
@@ -48,26 +48,26 @@ public void RunAllExceptionStatusCodeSuccessShouldBeFalse()
...
 
             string path = FilePathUtils.NormalizePathSeparators("c://test");
-            var target = new DotnetTestRunner(path, processMock.Object, OptimizationFlags.NoOptimization, new string[] { "C://test//mytest.dll" });
+            var target = new DotnetTestRunner(path, processMock.Object, OptimizationFlags.NoOptimization, new[] { "C://test//mytest.dll" });
 
-            var result = target.RunAll(null, null);
+            var result = target.RunAll(null, null, null);
 
-            Assert.False(result.Success);
+            Assert.Equal(1, result.FailingTests.Count);
```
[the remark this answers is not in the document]
dupdob: still not available

src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/DotnetTestRunnerTests.cs
```diff
@@ -48,26 +48,26 @@ public void RunAllExceptionStatusCodeSuccessShouldBeFalse()
...
 
             string path = FilePathUtils.NormalizePathSeparators("c://test");
-            var target = new DotnetTestRunner(path, processMock.Object, OptimizationFlags.NoOptimization, new string[] { "C://test//mytest.dll" });
+            var target = new DotnetTestRunner(path, processMock.Object, OptimizationFlags.NoOptimization, new[] { "C://test//mytest.dll" });
 
-            var result = target.RunAll(null, new Mutant(){Id =  1});
+            var result = target.RunAll(null, new Mutant(){Id =  1}, null);
 
-            Assert.False(result.Success);
+            Assert.Equal(0, result.FailingTests.Count);
```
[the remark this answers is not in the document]
dupdob: nope

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
```diff
@@ -153,18 +151,35 @@ private void DumpBuildErrors(KeyValuePair<SyntaxTree, ICollection<Diagnostic>> s
...
         }
 
-        private SyntaxTree RemoveMutantIfStatements(SyntaxTree originalTree, ICollection<Diagnostic> diagnosticInfo,
+        private SyntaxTree RemoveMutantIfStatements(SyntaxTree originalTree, IEnumerable<Diagnostic> diagnosticInfo,
             bool lastAttempt, bool devMode)
         {
             var rollbackRoot = originalTree.GetRoot();
             // find all if statements to remove
             var brokenMutations = new Collection<SyntaxNode>();
+            var rolledBackSomeMutation = false;
```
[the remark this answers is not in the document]
dupdob: you are right. removed and changed using brokenMutations

src/Stryker.Core/Stryker.Core/Mutants/TestDescription.cs
```diff
@@ -21,23 +48,99 @@ public override bool Equals(object obj)
...
         }
 
         public override int GetHashCode()
         {
             return (Guid != null ? Guid.GetHashCode() : 0);
         }
+    }
 
-        public TestDescription(string guid, string name)
+    public class TestListDescription
```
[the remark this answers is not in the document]
dupdob: done

src/Stryker.Core/Stryker.Core/Mutants/TestDescription.cs
```diff
@@ -21,23 +48,99 @@ public override bool Equals(object obj)
...
+                return;
+            foreach (var testDescription in otherFailingTests._tests)
+            {
+                Add(testDescription);
+            }
+        }
+
+        public bool ContainsAny(TestListDescription other)
+        {
+            return _tests?.Any(other.Contains) == true;
```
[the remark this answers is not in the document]
dupdob: I can't even understand how it happened...

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestExecutor.cs
```diff
@@ -13,45 +13,97 @@ namespace Stryker.Core.MutationTest
...
+                    if (!result.SessionTimedOut)
+                    {
+                        // something bad happened.
+                        Logger.LogError($"Stryker failed to test {remainingMutants.Count} mutant(s).");
+                        return;
+                    }
+
+                    // test session's results have been corrupted by the time out
+                    // we retry and run tests one by one, if necessary
+                    if (remainingMutants.Count == 1)
```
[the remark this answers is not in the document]
dupdob: actually that's the logic: we removed every mutation for which the status has been determined.
plus, if there was a test active when the timeout triggers, we mark the associated mutation as `TimeOut` 
But my testing revealed that sometimes we get a timeout without having any results or any active test.
Probably because the endless loop somehow prevented vstest to report progress.

When this happen, here the logic is to fallback to a cautious approach by running each mutation on its own. I chosen this strategy to avoid having multiple timeouts by re running the same configuration. This is quite rare and random.

src/Stryker.CLI/ToRemove.cs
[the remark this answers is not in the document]
dupdob: already removed.

richardwerkman:
I've done some testing and here are my findings:

- `--coverage-analysis "perTest"`
  - Results in the same mutation score as master and is very fast! (nice job!)
- `--coverage-analysis "perTestInIsolation"`
  - Marks more mutants as "not covered". My mutation score is way lower than on master.
- `--coverage-analysis "off"`
  - No mutants were killed. All were marked as survived. (no errors or warnings on screen)
- `--coverage-analysis "all"`
  - Stryker crashed... See:
```
[09:48:45 ERR] An error occurred during the mutation test run
System.AggregateException: One or more errors occurred. (Multiple mutant test requires coverage analysis) (Multiple mutant test requires coverage analysis) (Multiple mutant test requires coverage analysis) ---> Stryker.Core.Exceptions.StrykerInputException: Multiple mutant test requires coverage analysis
   at Stryker.Core.TestRunners.VsTest.VsTestRunner.TestMultipleMutants(Nullable`1 timeoutMs, IReadOnlyList`1 mutants, TestUpdateHandler update) in C:\Dev\Repos\styker-net-forks\dubdop\stryker-net\src\Stryker.Core\Stryker.Core\TestRunners\VsTest\VsTestRunner.cs:line 101
   at Stryker.Core.TestRunners.VsTest.VsTestRunnerPool.TestMultipleMutants(Nullable`1 timeoutMs, IReadOnlyList`1 mutants, TestUpdateHandler update) in C:\Dev\Repos\styker-net-forks\dubdop\stryker-net\src\Stryker.Core\Stryker.Core\TestRunners\VsTest\VsTestRunnerPool.cs:line 51
   at Stryker.Core.MutationTest.MutationTestExecutor.RunTestSession(IList`1 mutantsToTest, Int32 timeoutMs, TestUpdateHandler updateHandler, Boolean forceSingle) in C:\Dev\Repos\styker-net-forks\dubdop\stryker-net\src\Stryker.Core\Stryker.Core\MutationTest\MutationTestExecutor.cs:line 80
   at Stryker.Core.MutationTest.MutationTestExecutor.Test(IList`1 mutantsToTest, Int32 timeoutMs, TestUpdateHandler updateHandler) in C:\Dev\Repos\styker-net-forks\dubdop\stryker-net\src\Stryker.Core\Stryker.Core\MutationTest\MutationTestExecutor.cs:line 35
   at Stryker.Core.MutationTest.MutationTestProcess.<>c__DisplayClass11_0.<Test>b__4(List`1 mutants) in C:\Dev\Repos\styker-net-forks\dubdop\stryker-net\src\Stryker.Core\Stryker.Core\MutationTest\MutationTestProcess.cs:line 215
   at System.Threading.Tasks.Parallel.<>c__DisplayClass33_0`2.<ForEachWorker>b__0(Int32 i)
   at System.Threading.Tasks.Parallel.<>c__DisplayClass19_0`1.<ForWorker>b__1(RangeWorker& currentWorker, Int32 timeout, Boolean& replicationDelegateYieldedBeforeCompletion)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.Tasks.Parallel.<>c__DisplayClass19_0`1.<ForWorker>b__1(RangeWorker& currentWorker, Int32 timeout, Boolean& replicationDelegateYieldedBeforeCompletion)
   at System.Threading.Tasks.TaskReplicator.Replica`1.ExecuteAction(Boolean& yieldedBeforeCompletion)
   at System.Threading.Tasks.TaskReplicator.Replica.Execute()
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.TaskReplicator.Run[TState](ReplicatableUserAction`1 action, ParallelOptions options, Boolean stopOnFirstFailure)
   at System.Threading.Tasks.Parallel.ForWorker[TLocal](Int32 fromInclusive, Int32 toExclusive, ParallelOptions parallelOptions, Action`1 body, Action`2 bodyWithState, Func`4 bodyWithLocal, Func`1 localInit, Action`1 localFinally)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.Tasks.Parallel.ThrowSingleCancellationExceptionOrOtherException(ICollection exceptions, CancellationToken cancelToken, Exception otherException)
   at System.Threading.Tasks.Parallel.ForWorker[TLocal](Int32 fromInclusive, Int32 toExclusive, ParallelOptions parallelOptions, Action`1 body, Action`2 bodyWithState, Func`4 bodyWithLocal, Func`1 localInit, Action`1 localFinally)
   at System.Threading.Tasks.Parallel.ForEachWorker[TSource,TLocal](IList`1 list, ParallelOptions parallelOptions, Action`1 body, Action`2 bodyWithState, Action`3 bodyWithStateAndIndex, Func`4 bodyWithStateAndLocal, Func`5 bodyWithEverything, Func`1 localInit, Action`1 localFinally)
   at System.Threading.Tasks.Parallel.ForEachWorker[TSource,TLocal](IEnumerable`1 source, ParallelOptions parallelOptions, Action`1 body, Action`2 bodyWithState, Action`3 bodyWithStateAndIndex, Func`4 bodyWithStateAndLocal, Func`5 bodyWithEverything, Func`1 localInit, Action`1 localFinally)
   at System.Threading.Tasks.Parallel.ForEach[TSource](IEnumerable`1 source, ParallelOptions parallelOptions, Action`1 body)
   at Stryker.Core.MutationTest.MutationTestProcess.Test(StrykerOptions options) in C:\Dev\Repos\styker-net-forks\dubdop\stryker-net\src\Stryker.Core\Stryker.Core\MutationTest\MutationTestProcess.cs:line 209
   at Stryker.Core.StrykerRunner.RunMutationTest(StrykerOptions options, IEnumerable`1 initialLogMessages) in C:\Dev\Repos\styker-net-forks\dubdop\stryker-net\src\Stryker.Core\Stryker.Core\StrykerRunner.cs:line 95
---> (Inner Exception #0)
```

I think the PR is definitely not ready for merge... When releasing a big feature like this users should be able to fall back to other coverage analyse strategies.

dupdob:
working on it

dupdob:
`--coverage-analysis "all"`: fixed. problem was triggered by the logic that verified the consistance of settings, which should have been removed when we opted to make multi testing default

dupdob:
I found the main issue: the StrykerDataCollector only supports 'perTest' mode. 🙄, me stupid
Fixing this.

dupdob:
I fixed all issues with alternative coverage modes.

dupdob:
Just added the final touch: now coverage-analysis 'all' mode ensures each (covered) mutations is run against all tests (vs against tests covering the mutation).

In short: each coverage analysis mode now works as documented when using VsTest.

Mobrockers:
@dupdob we're almost ready to merge this, I'm just gonna do one final check running all the options on my full framework project and then we should be good to go

dupdob:
cool 🤞

dupdob:
having issues is not surprising due to the scale of change I made to the code.
I added some unit tests to help secure results. But it is clear that it would make sense to have test for the various coverage analysis mode

Mobrockers:
Ignore the stupid amount of killed mutants in CA off and CA all, this is due to shared state in database drive tests throwing off the killed percentage.

CA off:

image

CA perTestInIsolation:

image

CA all:

image (1)

CA perTest:

image (2)


I'm done testing, I accept :) This is awesome

wmeints:
wowsers, that's quite an improvement!

PR #961, opened 2020-02-27
feat(mutation exclusion): allow property members to be excluded by name

riezebosch:
I've some readonly properties that only provide metadata and would like to ignore:

```
        string IRule.Description => "some description that is pushed to the frontend";
        string IRule.Link => "https://google.com";
```

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/IgnoredMethodMutantFilterTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/IgnoredMethodMutantFilterTests.cs
@@ -211,5 +211,67 @@ public class IgnoredMethodMutantFilter_NestedMethodCalls
             // Assert
             filteredMutants.ShouldContain(mutant);
         }
+        
+        [Fact]
+        public void MutantFilter_ShouldIgnoreProperty()
+        {
+            // Arrange
+            var source = @"
+public class IgnoredMethodMutantFilter_Properties
+{
+    public string Description { get; } = ""some input"";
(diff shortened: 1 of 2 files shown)


riezebosch:
Auto exclude by honoring  the `ExcludeFromCodeCoverage` attribute would even beter fit my purpose: #964 But I guess it still makes sense to expand the filter method to properties.

richardwerkman:
This will cause the `--ignore-methods` option to be quite expanded. Could you please update the documentation? 

Also: this doesn't only target your test cases but all properties. Excluding them seems rigorous as properties are often named the same across a lot of classes. You could add an extra test case for:
`public int Age { get; set; }`

But we can live with this fact so we would like to embrace this feature! But please create an issue first next time so we can discuss a new feature first 🚀

riezebosch:
> This will cause the `--ignore-methods` option to be quite expanded. Could you please update the documentation?
> 
> Also: this doesn't only target your test cases but all properties. Excluding them seems rigorous as properties are often named the same across a lot of classes. You could add an extra test case for:
> `public int Age { get; set; }`
> 
> But we can live with this fact so we would like to embrace this feature! But please create an issue first next time so we can discuss a new feature first 🚀

I'm not sure if I still want this myself since I already dislike the ambiguity of given filter. The use of excluding method-calls seems appropriate since one can't really test the mutations on `ConfigureAwait` for example but excluding method-bodies (like it already does) is a whole different beast.

Since you already accepted my other PR where I can opt-out of testing with the `ExcludeCoverage` attribute you maybe want to restrict this feature even more.

PR #964, opened 2020-02-28
feat(mutation exclusion): mutants with ExcludeFromCodeCoverage attribute

riezebosch:
If you choose to exclude something from coverage I think it makes sense to also exclude it from mutation testing.

### DIFF
--- a/src/Stryker.Core/Stryker.Core/MutantFilters/ExcludeFromCodeCoverageFilter.cs
+++ b/src/Stryker.Core/Stryker.Core/MutantFilters/ExcludeFromCodeCoverageFilter.cs
@@ -1,0 +1,45 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Stryker.Core.Mutants;
+using Stryker.Core.Options;
+using Stryker.Core.ProjectComponents;
+
+namespace Stryker.Core.MutantFilters
+{
+    public class ExcludeFromCodeCoverageFilter : IMutantFilter
+    {
+        string IMutantFilter.DisplayName { get; } = "exclude from code coverage filter";
+
+        public IEnumerable<Mutant> FilterMutants(IEnumerable<Mutant> mutants, FileLeaf file, StrykerOptions options)
+        {
+            return mutants.Where(m => !Exclude(m.Mutation.OriginalNode));
+        }
+
+        private static bool Exclude(SyntaxNode node)
+        {
+            return node != null && (HasExcludeFromCodeCoverageAttribute(node) || Exclude(node.Parent));
+        }
+
+        private static bool HasExcludeFromCodeCoverageAttribute(SyntaxNode node)
+        {
+            switch (node)
+            {
+                case MemberDeclarationSyntax m:
+                    return HasExcludeFromCodeCoverageAttribute(m);
+                default:
+                    return false;
+            }
+        }
+
+        private static bool HasExcludeFromCodeCoverageAttribute(MemberDeclarationSyntax m)
+        {
+            return m.AttributeLists
+                .SelectMany(attr => attr.Attributes)
+                .Any(attr => Regex.IsMatch(attr.ToString(),
+                    @"^(?:System\.Diagnostics\.CodeAnalysis\.)?ExcludeFromCodeCoverage(?:Attribute)?$"));
+        }
+    }
+}
--- a/src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/ExcludeFromCodeCoverageFilterTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/ExcludeFromCodeCoverageFilterTests.cs
@@ -1,0 +1,146 @@
+using System;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Text;
+using Shouldly;
+using Stryker.Core.MutantFilters;
+using Stryker.Core.Mutants;
+using Stryker.Core.Options;
+using Xunit;
+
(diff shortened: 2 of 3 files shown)


src/Stryker.Core/Stryker.Core/MutantFilters/ExcludeFromCodeCoverageFilter.cs
```diff
@@ -0,0 +1,38 @@
...
+                case MemberDeclarationSyntax m:
+                    return HasExcludeFromCodeCoverageAttribute(m);
+                default:
+                    return false;
+            }
+        }
+
+        private static bool HasExcludeFromCodeCoverageAttribute(MemberDeclarationSyntax m) =>
+            m.AttributeLists.Any(attr =>
+                attr.Attributes.Any(a => Regex.IsMatch(a.ToFullString(), @"^(?:System\.Diagnostics\.CodeAnalysis\.)?ExcludeFromCodeCoverage(?:Attribute)?$")));
```
riezebosch: Not sure if there's a better way to validate attributes without to much context.

richardwerkman: I don't think there is. We won't calculate the semantic model for each file because of the performance costs. So we have no way of getting the full name of the attribute for now.

riezebosch: By that I changed it to `ToString`

riezebosch:
Have ran it on my own realistic project and the recursive traversal of the tree handles this for sure but I agree on improving the mutants used in test.

riezebosch:
Nice: 
<pre><code>
[11:41:16 INF] Using testrunner VsTest
[11:41:16 INF] Total number of tests found: 230
[11:41:16 INF] Initial testrun started
[11:43:26 INF] Using 199299 ms as testrun timeout
[11:43:30 INF] <b>69 mutants got status Skipped. Reason: Removed by exclude from code coverage filter</b>
[11:43:30 INF] 78 mutants got status Skipped. Reason: Removed by method filter
[11:43:30 INF] 477 mutants ready for test
</code></pre>

src/Stryker.Core/Stryker.Core/MutantFilters/ExcludeFromCodeCoverageFilter.cs
```diff
@@ -0,0 +1,38 @@
...
+using Stryker.Core.Options;
+using Stryker.Core.ProjectComponents;
+
+namespace Stryker.Core.MutantFilters
+{
+    public class ExcludeFromCodeCoverageFilter : IMutantFilter
+    {
+        string IMutantFilter.DisplayName { get; } = "exclude from code coverage filter";
+
+        IEnumerable<Mutant> IMutantFilter.FilterMutants(IEnumerable<Mutant> mutants, FileLeaf file, StrykerOptions options) =>
```
richardwerkman:
```suggestion
        IEnumerable<Mutant> FilterMutants(IEnumerable<Mutant> mutants, FileLeaf file, StrykerOptions options) =>
```

src/Stryker.Core/Stryker.Core/MutantFilters/ExcludeFromCodeCoverageFilter.cs
```diff
@@ -0,0 +1,38 @@
...
+            mutants.Where(m => !Exclude(m.Mutation.OriginalNode));
+
+        private static bool Exclude(SyntaxNode node) =>
```
richardwerkman: Could you refactor these methods to regular methods? We don't use arrow methods like this anywhere else in the codebase.

src/Stryker.Core/Stryker.Core/MutantFilters/ExcludeFromCodeCoverageFilter.cs
```diff
@@ -0,0 +1,38 @@
...
+                    return HasExcludeFromCodeCoverageAttribute(m);
+                default:
+                    return false;
+            }
+        }
+
+        private static bool HasExcludeFromCodeCoverageAttribute(MemberDeclarationSyntax m) =>
+            m.AttributeLists
+                .SelectMany(attr => attr.Attributes)
+                .Any(attr => Regex.IsMatch(attr.ToString(), @"^(?:System\.Diagnostics\.CodeAnalysis\.)?ExcludeFromCodeCoverage(?:Attribute)?$"));
```
richardwerkman: I just learned that Namespace naming is enclosed in `QualifiedNameSyntax` nodes. So you could also get the `IdentifierNameSyntax` from the attribute DescendingNodesAndSelf(). We could get rid of the regex that way.

src/Stryker.Core/Stryker.Core/MutantFilters/ExcludeFromCodeCoverageFilter.cs
riezebosch: Sure thing, but not entirely true: https://github.com/stryker-mutator/stryker-net/blob/cee97b5113e4896df1443fadf8a31bd7514ab08d/src/Stryker.Core/Stryker.Core/MutantFilters/ExcludeMutationMutantFilter.cs#L17

src/Stryker.Core/Stryker.Core/MutantFilters/ExcludeFromCodeCoverageFilter.cs
riezebosch: Not sure how to fix this in the tests since I suspect it to not be available with the currently used code snippets.

src/Stryker.Core/Stryker.Core/MutantFilters/ExcludeFromCodeCoverageFilter.cs
richardwerkman: Yeah true. But I think that is different from a whole expression that has to fit on another line 😉

PR #965, opened 2020-03-04
feat(mutation placing): Return default(T) for async Task<T> methods

richardwerkman:
Try and fix the return default bug when the return value is a task

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
+++ b/src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
@@ -329,6 +329,7 @@ namespace Stryker.Core.Mutants
             {
                 TypeSyntax returnType = methodNode.ReturnType;
 
+                // the GenericNameSyntax node can be encapsulated by QualifiedNameSyntax nodes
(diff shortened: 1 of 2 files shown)


richardwerkman:
I'm in doubt if checking for the name "Task" is the best solution to this. Maybe I can check if the method is async as well.

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -319,13 +319,32 @@ private IEnumerable<Mutant> ApplyMutator(SyntaxNode syntaxNode, IMutator mutator
...
 
+        /// <summary>
+        /// Add return default to the end of the method to prevent "not all code paths return a value" error as a result of mutations
+        /// </summary>
         private SyntaxNode AddReturnDefault(SyntaxNode currentNode)
         {
-            // add return default to the end of the method to prevent "not all code paths return a value" error as a result of mutations
-            if (currentNode is MethodDeclarationSyntax methodNode && 
-                methodNode.ReturnType.ToString() != "void" && methodNode.Body != null)
+            if (currentNode is MethodDeclarationSyntax methodNode && methodNode.Body != null &&
```
Mobrockers: This would be much more readable if it was not one big if

PR #993, opened 2020-03-20
fix(cleartext reporter): Re-add mutation score after a mutation testrun

### LINKED ISSUE #986: Console report misses mutation score
**Describe the bug**
The console report no longer shows the complete mutation score. Only of files.

image

**Expected behavior**
I expect the folder structure to be shown and the mutation score to be shown for the whole project and per directory.

 - OS: windows
 - Type of project: core
 - Stryker Version 0.16.1 (master)

**Additional context**

Mobrockers:
Fixes #986


src/Stryker.Core/Stryker.Core/ProjectComponents/ProjectComponent.cs
```diff
@@ -44,37 +44,34 @@ public abstract class ProjectComponent : IReadOnlyInputComponent
...
-                    return Health.Danger;
+                var mutationScorePercentage = mutationScore * 100;
+
+                switch (mutationScorePercentage)
+                {
+                    case var score when score > threshold.High:
+                        return Health.Good;
+                    case var score when score <= threshold.High && score > threshold.Low:
+                        return Health.Warning;
+                    case var score when score <= threshold.Low:
```
richardwerkman: Why don't we make this the default? If the health is not good or warning it should be bad.

PR #1000, opened 2020-03-22
feat: add statistics message

### LINKED ISSUE #824: Coverage analysis statistics (x out of y mutants will be skipped due to not being covered) are no longer reported
**Describe the bug**
As seen in issue #821, Stryker outputs invalid optimisation statistics when the number of tests cannot be established.
This is a minor issue as the behaviour is not impacted.
**Logs**
> [15:26:34 INF] Coverage analysis eliminated 100.00% of tests (i.e. -132 tests out of -132).

**Expected behavior**
Either provide accurate statistics or remove the message altogether.

dupdob:
Restore coverage analysis statistics (fix #824)

### DIFF
--- a/src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
+++ b/src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
@@ -13,6 +13,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.IO.Abstractions;
 using System.Linq;
+using System.Security.Principal;
 using System.Threading.Tasks;
 
 namespace Stryker.Core.MutationTest
@@ -167,6 +168,7 @@ namespace Stryker.Core.MutationTest
 
         public StrykerRunResult Test(StrykerOptions options)
         {
+            var viableMutantsCount = _input.ProjectInfo.ProjectContents.Mutants.Count(x => x.CountForStats);
             var mutantsNotRun = _input.ProjectInfo.ProjectContents.Mutants.Where(x => x.ResultStatus == MutantStatus.NotRun).ToList();
 
             if (!mutantsNotRun.Any())
@@ -186,7 +188,27 @@ namespace Stryker.Core.MutationTest
                 }
             }
 
-            var mutantsToTest = mutantsNotRun.Where(x => x.ResultStatus != MutantStatus.Ignored && x.ResultStatus != MutantStatus.NoCoverage);
+            var mutantsToTest = mutantsNotRun;
+            if (_options.Optimizations.HasFlag(OptimizationFlags.CoverageBasedTest))
+            {
+                var testCount = _mutationTestExecutor.TestRunner.DiscoverNumberOfTests();
+                var toTest = @mutantsNotRun.Sum(x => x.MustRunAgainstAllTests ? testCount : x.CoveringTests.Count);
+                var total = testCount * viableMutantsCount;
+                if (total > 0 && total != toTest)
+                {
+                    _logger.LogInformation($"Coverage analysis eliminated {(total-toTest)/(double)total:P1} useless tests.");
+                }
+            }
+            else if (_options.Optimizations.HasFlag(OptimizationFlags.SkipUncoveredMutants))
+            {
+                var total = viableMutantsCount;
+                var toTest = mutantsToTest.Count();
+                if (total > 0  && total!= toTest)
+                {
+                    _logger.LogInformation($"Coverage analysis eliminated {(total-toTest)/(double)total:P1} useless tests.");
+                }
+            }
+            
             if (mutantsToTest.Any())
             {
                 var mutantGroups = BuildMutantGroupsForTest(mutantsNotRun);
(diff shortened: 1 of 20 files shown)


dupdob:
cool. I got PR #1000!!! without even trying

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
```diff
@@ -186,7 +188,27 @@ public StrykerRunResult Test(StrykerOptions options)
...
-            var mutantsToTest = mutantsNotRun.Where(x => x.ResultStatus != MutantStatus.Ignored && x.ResultStatus != MutantStatus.NoCoverage);
+            var mutantsToTest = mutantsNotRun;
+            if (_options.Optimizations.HasFlag(OptimizationFlags.CoverageBasedTest))
+            {
+                var testCount = _mutationTestExecutor.TestRunner.DiscoverNumberOfTests();
+                var toTest = @mutantsNotRun.Sum(x => x.MustRunAgainstAllTests ? testCount : x.CoveringTests.Count);
+                var total = testCount * viableMutantsCount;
+                if (total > 0 && total != toTest)
+                {
+                    _logger.LogInformation($"Coverage analysis eliminated {(total-toTest)/(double)total:P1} useless tests.");
```
Mobrockers:
```suggestion
                    _logger.LogInformation($"Coverage analysis eliminated {(total-toTest)/(double)total:P1} of tests because they do not cover a mutant.");
```

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
```diff
@@ -186,7 +188,27 @@ public StrykerRunResult Test(StrykerOptions options)
...
+                }
+            }
+            else if (_options.Optimizations.HasFlag(OptimizationFlags.SkipUncoveredMutants))
+            {
+                var total = viableMutantsCount;
+                var toTest = mutantsToTest.Count();
+                if (total > 0  && total!= toTest)
+                {
+                    _logger.LogInformation($"Coverage analysis eliminated {(total-toTest)/(double)total:P1} useless tests.");
```
Mobrockers:
```suggestion
                    _logger.LogInformation($"Coverage analysis eliminated {(total-toTest)/(double)total:P1} of tests because they do not cover a mutant.");
```

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
dupdob: Actually, this is not about eliminating tests that do not cover mutants. Here it is about keeping only the tests that covers the mutant you are testing, so not sure how to translate that

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
dupdob: Here, the strategy is simply to removed non covered mutants so we can skip testing the associated tests. so the proposed message is not a true descritption

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
Mobrockers: How do you feel about my update?

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
Mobrockers: I'm confused, if a mutant is not covered, how can it have associated tests?

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
dupdob: It is better, but I think it is still misguiding
What I attempted here is to convey an overall impression of how much time the coverage analysis  feature helped saved. I kept both messages similar for consistency. They are both true, but do not describe precisely what is happening.

In both cases, the algorithm is compute the percentage of tests that Stryker will not run (thanks to coverage analysis). The initial number of needed tests is:
number of tests in the assembly(ies) * number of mutations, quite a large number for a significant project (roughly 4 millions for NFluent as an example).
Bear in mind this figure is provided for the whole test phase, not per mutant.

With `perTest` coverage analysis, each mutation will only be tested with the test covering it. Other tests will not be run, except when we cannot reliably determine the needed tests, the so called _static mutations_. 

Here we skip all tests that provide no information on the active mutation. In practice, I simply count how many tests will be run.

With 'all' coverage analysis, we will run every tests against every test mutations, assuming the mutation is tested by at least one test. Since the 'all' mode mostly make sense for `dotnet test`, where the number of tests is not available, I use the number of test sessions to compute the percentage. And it stil provides an accurate result, since each test session will run all existing tests.

**Ultimately**: what message do you want to convey with this information?

Mobrockers: My main issue with the original message you proposed is that 'useless' has a negative value connotation. It could appear to the user that their unit tests are useless but that is not the case.

Furthermore it is not explaining why the tests are eliminated, so in that case the message is quite useless to them.

I would like the message to convey to the user 'hey you have this feature turned on, it makes the mutation testrun this much faster because we don't have to run 98% of tests that would not have impact the result, but would have been run anyway if this feature was not turned on'

dupdob: I like this phrasing, I will make a proposal

dupdob:
Please review the latest message, I hope you will like them

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
dupdob: check the latest comit

Mobrockers:
Perfect :)

PR #1004, opened 2020-03-24
fix(threshold break): Threshold break calculation should use percentage

### LINKED ISSUE #1003: Run fails due to threshold break limit but mutation score is above break limit
Running the tool on version 0.17 the CLI build fails with the below message.

The mutation score is lower than the configured break threshold of 80%.
 If you're running in a CI environment, this means your pipeline will now fail.

My mutation score is 100%

[12:23:22 INF] Time Elapsed 00:01:49.1238882
[12:23:22 INF] The final mutation score is 100.00 %
[12:23:22 WRN] Final mutation score is below threshold break. Crashing...

Config file as below

{
  "stryker-config": {
    "log-level": "info",
    "test-runner": "DotnetTest",
    "timeout-ms": 100000,
    "threshold-high": 100,
    "threshold-low": 90,
    "threshold-break": 80,
    "excluded-mutations": [
      "string",
    ],
    "mutate": [
      "!**/Startup.cs",
      "!**/IConfigurationExtensions.cs"
    ],
    "reporters": "['html']"
  }
}

 - OS: Windows
 - Type of project: core
 - Framework Version: netcoreapp2.1
 - Stryker Version: 0.17

Mobrockers:
Fixes #1003

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/StrykerRunResultTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/StrykerRunResultTests.cs
@@ -20,7 +20,7 @@ namespace Stryker.Core.UnitTest
             var scoreIsLowerThanThresholdBreak = runResult.ScoreIsLowerThanThresholdBreak();
 
             // Assert
-            scoreIsLowerThanThresholdBreak.ShouldBeFalse("because the mutation score is higher than the threshold break");
+            scoreIsLowerThanThresholdBreak.ShouldBeFalse("because the mutation score is higher than or equal to the threshold break");
         }
 
         [Theory]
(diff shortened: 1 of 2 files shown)


src/Stryker.Core/Stryker.Core.UnitTest/StrykerRunResultTests.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            // Arrange
+            var options = new StrykerOptions(thresholdHigh: 100, thresholdLow: 100, thresholdBreak: thresholdBreak);
+            var runResult = new StrykerRunResult(options, mutationScore);
+
+            // Act
+            var scoreIsLowerThanThresholdBreak = runResult.ScoreIsLowerThanThresholdBreak();
+
+            // Assert
+            scoreIsLowerThanThresholdBreak.ShouldBeFalse("because the mutation score is higher than the threshold break");
```
richardwerkman:
```suggestion
            scoreIsLowerThanThresholdBreak.ShouldBeFalse("because the mutation score is higher than or equal to the threshold break");
```

PR #1017, opened 2020-04-02
fix: avoid mutating enum member declarations (fixes #1016)

### LINKED ISSUE #1016: Compile error for mutated enum member declarations
**Describe the bug**
When using enums that have calculations in it, Stryker.NET prints a warning message:

`Stryker.NET encountered an compile error in [...] with message: The expression being assigned to 'Fruit.Banana' must be constant (Source code: ((StrykerXGJbRBlHxqRdD9O.MutantControl.IsActive(0)?Apple + 1:Apple - 1)))`

The code used:
 ```csharp
public enum Fruit
{
    Apple = 1,
    Banana = (Apple - 1)
}
```

**Expected behavior**
No warning should be printed.

 - OS: Windows
 - Type of project: .Net Framework Library
 - Framework Version: 4.7.2
 - Stryker Version: 0.17.1 (beta)

**Additional context**
Enum member declarations must be constants. When doing a substraction from a constant like above, the result is still a constant and can be compiled. However, when the right-hand-side of the declaration is changed to `StrykerXGJbRBlHxqRdD9O.MutantControl.IsActive(0) ? Apple + 1 : Apple - 1`, this is no longer a constant and compilation fails.

A similar problem would occur for all constants. This is currently solved by skipping constant declarations and not mutating them at all. Therefore I fixed it by skipping enum member declarations as well and notmutating them. **I created a unit test and a fix for the bug already and will attach a pull request to this issue in a minute.**

In the long run, I think a concept should be developed to also mutate enum member declarations and constants, since changing a `-` to a `+` here would be a valid mutation that breaks unit tests. It therefore would provide a benefit to the user of Stryker.NET.

### PR DESCRIPTION
**Pull request by Danghor**
This excludes EnumMemberDeclaration nodes from the nodes to be mutated, which will avoid compiler errors due to the mutation engine generating non-constant code at a place where constants must be used.

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/MutantOrchestratorTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/MutantOrchestratorTests.cs
@@ -205,6 +205,11 @@ var (one, two) = ((StrykerNamespace.MutantControl.IsActive(0)?1 - 1:1 + 1), (Str
             ShouldMutateSourceToExpected(source, expected);
         }
 
+        /// <summary>
+        /// Verifies that <code>EnumMemberDeclarationSyntax</code> nodes are not mutated.
+        /// Mutating would introduce code like <code>StrykerXGJbRBlHxqRdD9O.MutantControl.IsActive(0) ? One + 1 : One - 1</code>
+        /// Since enum members need to be constants, this mutated code would not compile and print a warning.
+        /// </summary>
         [Fact]
         public void ShouldNotMutateEnum()
         {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/Stryker.Core/Stryker.Core.UnitTest/Mutants/MutantOrchestratorTests.cs [hunk 1]
```diff
@@ -205,6 +205,15 @@ public void ShouldNotMutateConst()
             ShouldMutateSourceToExpected(source, expected);
         }
 
+        [Fact]
+        public void ShouldNotMutateEnum()
```
richardwerkman, line 209: Please add a message as to why enums should not be mutated. We could forget why this is necessary.

PR #1023, opened 2020-04-05
feat(Collection initializer mutator): Add mutator for collections and arrays

richardwerkman:
This PR adds mutators for collections and arrays. 

I skipped the implicid array for now is it results in compile errors when mutated. We could mutate it by resolving its type and converting it to a typed array in the mutation. I'll create an issue for that.

### DIFF
--- a/docs/Mutators.md
+++ b/docs/Mutators.md
@@ -78,7 +78,7 @@ Stryker supports a variety of mutators, which are listed below. Do you have a su
 |`int[] numbers = { 1, 2 };`               | `int[] numbers = { };`               |
 |`new List<int> { 1, 2 };`                 | `new List<int> { };`                 |
 |`new Collection<int> { 1, 2 };`           | `new Collection<int> { };`           |
-|`new Dictionary<int, int> { { 1, 1 } };`  | `new Dictionary<int, int> { { } };`  |
+|`new Dictionary<int, int> { { 1, 1 } };`  | `new Dictionary<int, int> { };`      |
 
 ## Unary Operators
 |    Original   |   Mutated  | 
--- a/docs/Mutators.md
+++ b/docs/Mutators.md
@@ -76,7 +76,6 @@ Stryker supports a variety of mutators, which are listed below. Do you have a su
 | ---------------------------------------- | ------------------------------------ |
 |`new int[] { 1, 2 };`                     | `new int[] { };`                     |
 |`int[] numbers = { 1, 2 };`               | `int[] numbers = { };`               |
-|`new [] { 1, 2 };`                        | `new [] { };`                        |
 |`new List<int> { 1, 2 };`                 | `new List<int> { };`                 |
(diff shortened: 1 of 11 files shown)


jnyrup:
Can this also handle C#6 dictionary initializer like
```c#
new Dictionary<int, int>
{
    [0] = 1
};
```
Or should that be another issue?

richardwerkman:
@jnyrup I've checked and no, the current code doesn't mutate the new initializer. Apparently it's not of type `CollectionInitializerSyntax` but of type `ObjectInitializerSyntax`. 

If I add a mutator for that type it will also mutate object creations like:
``` cs
var person = new Person()
{
    Age = 1
};
```
We could check in the mutator if the classname of the created object == "Dictionary" but that seems a bit tricky. Or we could just also mutate all object initializers as well. I could see the value in that. Maybe it's best to create a seperate issue indeed so we can discuss this over there.

src/Stryker.Core/Stryker.Core/Mutators/ObjectCreationMutator.cs
```diff
@@ -0,0 +1,26 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Stryker.Core.Mutants;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Stryker.Core.Mutators
+{
+    public class ObjectCreationMutator : MutatorBase<ObjectCreationExpressionSyntax>, IMutator
```
Mobrockers, line 10: Since you did add this, it would probably be a good idea to describe this in the docs.

docs/Mutators.md
```diff
@@ -70,6 +71,15 @@ Stryker supports a variety of mutators, which are listed below. Do you have a su
 |`^= ` | `\|= `|
 |`^= ` | `&= ` |
 
+## Collection initialization
```
Mobrockers, line 74: Could probably add a note that this works for everything that uses object initializer syntax, maybe with a link to:https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/object-and-collection-initializers

richardwerkman: But it doesn't. The whole point (for now) is to only mutate collections. I check if the kind is "Colleciton initializer" and not "Object initializer". We could add the object initializer as well but I'd do that in a seperate PR.

src/Stryker.Core/Stryker.Core/Mutators/ObjectCreationMutator.cs
richardwerkman: It is described 👀

docs/Mutators.md
Mobrockers: In that case ObjectCreationMutator can use a comment to explain that difference as I hadn't caught it from the kind name alone

PR #1039, opened 2020-04-17
feat(Multi project runs): Multi project runs using solution file

### LINKED ISSUE #459: Multi project runs: Mutate more projects at once

### LINKED ISSUE #1076: DefineConstants are not parsed correctly
Stryker throws a compile error during the initial test run on vstest.
No special configurations, just specified the path for my *.csproj file on cmd. 
Attached the file with logs on the cmd during test run.
[StrykerLogs.txt](https://github.com/stryker-mutator/stryker/files/4628241/StrykerLogs.txt)
 
Please let me know if you need additional information.

richardwerkman:
blocked by #1136

closes #459 
closes #1076

Also utilizes the new interfaces from buildalyzer to make testing possible for the parts that use buildalyzer. I'll wait to add more tests after merging this PR as this one is already quite large.

Needs some performance tweaks:
- The initial build is called for each test project, it could run for the whole solution once
- Initial testing could also be done in one run??
- Buildalyzer analyses the projects twice now. Once in the projectOrchestrator and once in the InputFileResolver. It would be better to pass the buildalyzer results to reuse. 

Open questions:
 - Should the thresholds count for the whole solution or for each project seperate? For example:
    - If one project scores 50% and the other 100% with equal mutants, would the run succeed with threshold break: 60?


src/Stryker.Core/Stryker.Core/Reporters/JsonReporter/JsonReportFileComponent.cs
```diff
@@ -22,7 +22,7 @@ private JsonReportFileComponent(string language, string source, ISet<JsonMutant>
 
         public JsonReportFileComponent(FileLeaf file, ILogger logger = null)
         {
-            var log = logger ?? ApplicationLogging.LoggerFactory.CreateLogger<JsonReportFileComponent>();
+            logger = logger ?? ApplicationLogging.LoggerFactory.CreateLogger<JsonReportFileComponent>();
```
Mobrockers, line 25: Did sonarqube not mention that you should not overwrite a method parameter value?

richardwerkman: No

Mobrockers: Mhmm 🤔

PR #1055, opened 2020-04-24
refactor(MutantFilters): Add MutantFilterFactory and BroadcastMutantFilter

### LINKED ISSUE #1054: MutantFilterFactory and BroadcastMutantFilter
Right now the MutantFilters have the responsibility of checking whether they should be applied or not. 

I would like to move the handling of mutants to a `BroadcastMutantFilter` using the same design pattern as the `BroadcastReporter`. We could create a Factory that checks the `StrykerOptions` whether or not a `MutantFilter` has to be applied. If it's the case we add an instance of that filter to the `BroadCastMutantFilter`. This would also make filtering in `MutationTestProcess` more streamlined.

Stijn-Rutten:
closes #1054 

After merging this PR the following things will be refactored:

* Resposibility of creating `MutantFilters` is moved to a seperate `MutantFilterFactory`.
* Looping over the `MutantFitlers` no longer happens in `MutationTestProcess` but is done using a `BroadcastMutantFilter`.
* The `DiffMutantFilter` no longer is responsible for checking whether the `Diff` feature is enabled.

### DIFF
--- a/src/Stryker.Core/Stryker.Core/MutantFilters/MutantFilterFactory.cs
+++ b/src/Stryker.Core/Stryker.Core/MutantFilters/MutantFilterFactory.cs
@@ -14,7 +14,7 @@ namespace Stryker.Core.MutantFilters
         {
             if(options == null)
             {
-                throw new ArgumentNullException($"{nameof(options)} cannot be null");
+                throw new ArgumentNullException(nameof(options));
             }
 
             return new BroadcastMutantFilter(DetermineEnabledMutantFilters(options));
--- a/src/Stryker.Core/Stryker.Core/MutantFilters/MutantFilterFactory.cs
+++ b/src/Stryker.Core/Stryker.Core/MutantFilters/MutantFilterFactory.cs
@@ -10,13 +10,18 @@ namespace Stryker.Core.MutantFilters
     {
         public static IMutantFilter Create(StrykerOptions options)
         {
+            if(options == null)
+            {
+                throw new ArgumentNullException($"{nameof(options)} cannot be null");
+            }
+
             return new BroadcastMutantFilter(DetermineEnabledMutantFilters(options));
         }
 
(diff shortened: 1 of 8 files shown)


src/Stryker.Core/Stryker.Core/MutantFilters/MutantFilterFactory.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Stryker.Core.MutantFilters
+{
+    public static class MutantFilterFactory
+    {
+        public static IMutantFilter Create(StrykerOptions options)
+        {
+            if(options == null)
+            {
+                throw new ArgumentNullException($"{nameof(options)} cannot be null");
```
jnyrup, line 15: Contrary to most other exception constructors, the first parameter of `ArgumentNullException` is not `message` but `paramName`.

src/Stryker.Core/Stryker.Core/MutantFilters/MutantFilterFactory.cs
```diff
@@ -0,0 +1,47 @@
...
+                enabledFilters.Add(new DiffMutantFilter(DiffProvider ?? new GitDiffProvider(options)));
+            }
+            
+
+            return enabledFilters;
+
+           
+        }
+
+        public static void SetDiffProvider(IDiffProvider diffProvider)
```
Mobrockers, line 42: What's the intended use of this method?

Stijn-Rutten: I used this method for mocking of the `diffProvider` in de `DiffMutantFilter`. Now that I am rethinking this it could also be a nullable parameter of the `Create` method.

 The `diffProvider` has to be mocked because the constructor of `DiffMutantFilter` calls the `diffProvider` which in turn calls `Gitlib2sharp`. This results in an error in pipelines because these builds do a shallow fetch where branches don't exist.

PR #1101, opened 2020-06-12
feat(dashboard compare): test the mutants which are covered by unit tests in changed test files

### LINKED ISSUE #1068: Only test mutants that are covered by changed test files
Right now the diff feature resets all the mutants when a test is changed.

The current coverage analysis makes it possible to check which tests cover which mutants. I would extend this information by adding a property `TestFilename` to `TestListDescription` . This way we can check if the `TestFilename` equals one of the changed testfiles we using the `GitDiffProvider`. Whenever the tests are changed, the result status can than be reset to `NotRun`.

Ideally I would do like to check for a modified test which mutants it covers. This is however not a possibility because of the limitations of libgit2sharp. libgit2/libgit2sharp#1790 should fix this limitation. In the meanwhile however this is a good alternative

**Additional context**
This feature relies on #1026

Stijn-Rutten:
Extends de Diff feature to, making it possible to only test mutants on changed tests files.
Also adds the option to choose a baseline provider 

closes #1068

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -380,7 +380,7 @@ When enabling the --dashboard-compare feature you can provide a fallback version
 dotnet stryker --dashboard-compare --dashboard-fallback-version master
 dotnet stryker -compare -fallback-version master
 ```
-Default: `null`
+Default: value provided to --git-source or null
 
 ## Baseline Storage location
 Sets the storage location for the baseline used by --dashboard-compare. By default this is set to disk, when the dashboard reporter is enabled this is automatically set to Stryker Dashboard.
@@ -431,7 +431,7 @@ When configuring the --dashboard-compare feature on pull requests please provide
 4. Set --git-source to the name of the target branch of your pull request.
 
 ```
-dotnet stryker --dashboard-compare --dashboard-fallback-version master -git-source master --dashboard-version development
-dotenet stryker -compare -fallback-version master -source master -version development
+dotnet stryker --dashboard-compare --git-source master --dashboard-version development
+dotnet stryker -compare -source master -version development
 
 ```
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -389,9 +389,9 @@ Supported storage locations are:
 
 | Storage location | Option | Description |
 |------------------|--------|-------------|
-| Disk             | disk   | Saves the baseline to the `StrykerOutput` folder|
-| Stryker Dashboard| dashboard | Saves the baseline to Stryker Dashboard |
-| Azure File Storage | azure | Saves the baseline to Azure File Storage |
+| Disk             | Disk   | Saves the baseline to the `StrykerOutput` folder|
+| Stryker Dashboard| Dashboard | Saves the baseline to Stryker Dashboard |
+| Azure File Storage | AzureFileStorage | Saves the baseline to Azure File Storage |
 
 ```
 dotnet stryker --dashboard-compare --baseline-storage-location disk
@@ -414,11 +414,11 @@ Providing a subfolder is optional, we storage the baseline in a `StrykerOutput` 
 For authentication we support Shared Access Signatures. For more information on how to configure a SAS check the [Azure documentation](https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview).
 
 ```
-dotnet stryker --dashboard-compare --baseline-storage-location azure --azure-file-storage-url https://STORAGE_NAME.file.core.windows.net/FILE_SHARE/(optional)SUBFOLDER --azure-sas SAS_KEY
+dotnet stryker --dashboard-compare --baseline-storage-location AzureFileStorage --azure-storage-url https://STORAGE_NAME.file.core.windows.net/FILE_SHARE/(optional)SUBFOLDER --azure-storage-sas STORAGE_SAS
 
 or
 
-dotnet stryker -compare -bsl azure -storage-url --azure-file-storage-url https://STORAGE_NAME.file.core.windows.net/FILE_SHARE/(optional)SUBFOLDER -sas SAS_KEY
+dotnet stryker -compare -bsl AzureFileStorage -storage-url https://STORAGE_NAME.file.core.windows.net/FILE_SHARE/(optional)SUBFOLDER -sas STORAGE_SAS
 ```
 
 
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -364,7 +364,7 @@ dotnet stryker -gs "development"
 
 Default: `master`
 
-## Dashboard Compare
+## EXPERIMENTAL: Dashboard Compare
 Enabling the dashboard compare feature saves reports and re-uses the result when a mutant or it's tests are unchanged.
 
 ```
@@ -375,6 +375,8 @@ dotnet stryker -compare
 Default `"off"`
 
 This feature automatically enables the --diff feature.
+
+This feature is experimental. Results can contain slight false postives and false negatives.
 
 ## Fallback version
 When enabling the --dashboard-compare feature you can provide a fallback version. This version will be used to pull a baseline when we cannot find a baseline for your current branch. When we are still unable to provide a baseline we will start a complete testrun to create a complete baseline.
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -30,7 +30,6 @@ Example `stryker-config.json` file:
 {
     "stryker-config":
     {
-        "test-runner": "vstest",
         "reporters": [
             "progress",
             "html"
@@ -56,7 +55,11 @@ Example `stryker-config.json` file:
             "*Log*",
             "ToString",
             "*HashCode*"
-        ]
+        ],
+        "dashboard-compare": true,
+        "baseline-storage-location": "AzureFileStorage",
+        "azure-storage-url": "https://storageaccount.file.core.windows.net/sharename",
+        "azure-storage-sas": "<SAS>"
     }
 }
 ```
@@ -408,11 +411,22 @@ This is the url to your Azure File Storage. The URL should look something like t
 ```
 https://STORAGE_NAME.file.core.windows.net/FILE_SHARE/(optional)SUBFOLDER
 ```
-Providing a subfolder is optional, we storage the baseline in a `StrykerOutput` subfolder.
+Providing a subfolder is optional, we store the baseline in a `StrykerOutput` subfolder.
+
+```
+-storage url https://STORAGE_NAME.file.core.windows.net/FILE_SHARE/(optional)SUBFOLDER
+--azure-storage-url https://STORAGE_NAME.file.core.windows.net/FILE_SHARE/(optional)SUBFOLDER
+```
 
 ### Shared Access Signature (SAS)
 For authentication we support Shared Access Signatures. For more information on how to configure a SAS check the [Azure documentation](https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview).
 
+```
+-storage-sas <STORAGE_SAS>
+--azure-storage-sas <STORAGE_SAS>
+```
+
+The complete configuration would look like this:
 ```
 dotnet stryker --dashboard-compare --baseline-storage-location AzureFileStorage --azure-storage-url https://STORAGE_NAME.file.core.windows.net/FILE_SHARE/(optional)SUBFOLDER --azure-storage-sas STORAGE_SAS
 
--- a/src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
+++ b/src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
@@ -1,5 +1,6 @@
 ﻿using McMaster.Extensions.CommandLineUtils;
 using Microsoft.CodeAnalysis.CSharp;
+using Stryker.Core.Baseline;
 using Stryker.Core.Options;
 using Stryker.Core.Reporters;
 using Stryker.Core.TestRunners;
@@ -120,7 +121,8 @@ namespace Stryker.CLI
         {
             ArgumentName = "--baseline-storage-location",
             ArgumentShortName = "-bsl <storageLocation>",
-            ArgumentDescription = $@"Allows to choose a storage location, this can be either azure, disk or dashboard. When the dashboard reporter is enabled, the dashboard will be the default storage location. Otherwise disk is the default storage location. Selecting azure allows you to store baselines using your own instance of azure file storage.",
+            ArgumentDescription = $@"Allows to choose a storage location | Options[{FormatOptionsString(_defaultOptions.BaselineProvider, (IEnumerable<BaselineProvider>)Enum.GetValues(_defaultOptions.BaselineProvider.GetType())) }]
+                                     When using the azure file storage, make sure to configure the -sas and -storage-url options.",
             ValueType = CommandOptionType.SingleValue,
             DefaultValue = _defaultOptions.BaselineProvider.ToString(),
             JsonKey = "baseline-storage-location"
@@ -340,7 +342,7 @@ For example: Your project might be called 'consumer-loans' and it might contains
         public static readonly CLIOption<string> AzureFileStorageUrl = new CLIOption<string>
         {
             ArgumentName = "--azure-storage-url",
-            ArgumentShortName = "-storage-url",
+            ArgumentShortName = "-storage-url <url>",
             ArgumentDescription = @"The url for the Azure File Storage, only needed when the azure baseline provider is selected. 
                                     The url should look something like this: 
 
--- a/src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
+++ b/src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
@@ -124,7 +124,6 @@ namespace Stryker.CLI
             ValueType = CommandOptionType.SingleValue,
             DefaultValue = _defaultOptions.BaselineProvider.ToString(),
             JsonKey = "baseline-storage-location"
-
         };
 
         public static readonly CLIOption<string> GitSource = new CLIOption<string>
--- a/src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
+++ b/src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
@@ -111,7 +111,7 @@ namespace Stryker.CLI
             ArgumentShortName = "-compare",
             ArgumentDescription = $@"Enables comparing to results stored in Stryker Dashboard. This feature is only available in combination with {Diff.ArgumentName}",
             ValueType = CommandOptionType.NoValue,
-            JsonKey = "dashboard-diff-compare"
+            JsonKey = "dashboard-compare"
 
         };
 
@@ -120,7 +120,7 @@ namespace Stryker.CLI
         {
             ArgumentName = "--baseline-storage-location",
             ArgumentShortName = "-bsl <storageLocation>",
-            ArgumentDescription = $@"Allows to choose a storage location, this can be either disk or dashboard. When the dashboard reporter is enabled, the dashboard will be the default storage location. Otherwise disk is the default storage location. ",
+            ArgumentDescription = $@"Allows to choose a storage location, this can be either azure, disk or dashboard. When the dashboard reporter is enabled, the dashboard will be the default storage location. Otherwise disk is the default storage location. Selecting azure allows you to store baselines using your own instance of azure file storage.",
             ValueType = CommandOptionType.SingleValue,
             DefaultValue = _defaultOptions.BaselineProvider.ToString(),
             JsonKey = "baseline-storage-location"
@@ -330,16 +330,16 @@ For example: Your project might be called 'consumer-loans' and it might contains
 
         public static readonly CLIOption<string> AzureSAS = new CLIOption<string>
         {
-            ArgumentName = "--azure-sas-key",
+            ArgumentName = "--azure-storage-sas",
             ArgumentShortName = "-sas <azure-sas-key>",
             ArgumentDescription = $"The Shared Access Signature for Azure File Storage, only needed when the azure baseline provider is selected. For more information: https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview",
             DefaultValue = null,
-            JsonKey = "azure-sas-key"
+            JsonKey = "azure-storage-sas"
         };
 
         public static readonly CLIOption<string> AzureFileStorageUrl = new CLIOption<string>
         {
-            ArgumentName = "--azure-file-storage-url",
+            ArgumentName = "--azure-storage-url",
             ArgumentShortName = "-storage-url",
             ArgumentDescription = @"The url for the Azure File Storage, only needed when the azure baseline provider is selected. 
                                     The url should look something like this: 
@@ -348,7 +348,7 @@ For example: Your project might be called 'consumer-loans' and it might contains
 
                                     Note, the url might be different depending of where your file storage is hosted.",
             DefaultValue = null,
-            JsonKey = "azure-file-storage-url"
+            JsonKey = "azure-storage-url"
         };
 
         private static string FormatOptionsString<T, Y>(T @default, IEnumerable<Y> options)
--- a/src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
+++ b/src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
@@ -110,7 +110,7 @@ namespace Stryker.CLI
         {
             ArgumentName = "--dashboard-compare",
             ArgumentShortName = "-compare",
-            ArgumentDescription = $@"Enables comparing to results stored in Stryker Dashboard. This feature is only available in combination with {Diff.ArgumentName}",
+            ArgumentDescription = $@"EXPERIMENTAL: Enables comparing to results stored in Stryker Dashboard. This feature is only available in combination with {Diff.ArgumentName}",
             ValueType = CommandOptionType.NoValue,
             JsonKey = "dashboard-compare"
 
--- a/src/Stryker.Core/Stryker.Core/Baseline/AzureFileShareBaselineProvider.cs
+++ b/src/Stryker.Core/Stryker.Core/Baseline/AzureFileShareBaselineProvider.cs
@@ -4,6 +4,7 @@ using Stryker.Core.Logging;
 using Stryker.Core.Options;
 using Stryker.Core.Reporters.Json;
 using System;
+using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Text;
@@ -16,7 +17,7 @@ namespace Stryker.Core.Baseline
         private readonly StrykerOptions _options;
         private readonly HttpClient _httpClient;
         private readonly ILogger<AzureFileShareBaselineProvider> _logger;
-        private const string _outputPath = "StrykerOutput/Baselines/";
+        private const string _outputPath = "StrykerOutput/Baselines";
 
         public AzureFileShareBaselineProvider(StrykerOptions options, HttpClient httpClient = null)
         {
@@ -92,12 +93,15 @@ namespace Stryker.Core.Baseline
         {
             _logger.LogDebug("Creating directories for file {0}", fileUrl);
 
-            var pathSegments = fileUrl.Split('/');
-            var currentDirectory = pathSegments[0];
+            var uriParts = fileUrl.Split(_outputPath);
+            var currentDirectory = $"{uriParts[0]}";
 
-            for (var i = 1; i < pathSegments.Length; i++)
+            var storagePathSegments = _outputPath.Split('/').Concat(uriParts[1].Split('/')).Where(s => !string.IsNullOrEmpty(s) && !s.Contains("."));
+
+            foreach (var segment in storagePathSegments)
             {
-                if (!await CreateDirectory($"{currentDirectory}/{pathSegments[i]}"))
+                currentDirectory += $"{segment}/";
+                if (!await CreateDirectory(currentDirectory))
                 {
                     return false;
                 }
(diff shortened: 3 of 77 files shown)


Stijn-Rutten:
The bug marked by SonarCloud is a false positive in my opinion. However I can't mark it as one.

src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
```diff
@@ -115,6 +115,18 @@ public static class CLIOptions
...
+
+        public static readonly CLIOption<string> BaselineStorageLocation = new CLIOption<string>
+        {
+            ArgumentName = "--baseline-storage-location",
+            ArgumentShortName = "-bsl <storageLocation>",
+            ArgumentDescription = $@"Allows to choose a storage location, this can be either disk or dashboard. When the dashboard reporter is enabled, the dashboard will be the default storage location. Otherwise disk is the default storage location. ",
+            ValueType = CommandOptionType.SingleValue,
+            DefaultValue = _defaultOptions.BaselineProvider.ToString(),
+            JsonKey = "baseline-storage-location"
+
```
Mobrockers, line 127: Extra whitespace mag weg

src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
```diff
@@ -111,10 +111,21 @@ public static class CLIOptions
...
+            JsonKey = "dashboard-compare"
 
         };
 
+
+        public static readonly CLIOption<string> BaselineStorageLocation = new CLIOption<string>
+        {
+            ArgumentName = "--baseline-storage-location",
+            ArgumentShortName = "-bsl <storageLocation>",
+            ArgumentDescription = $@"Allows to choose a storage location, this can be either azure, disk or dashboard. When the dashboard reporter is enabled, the dashboard will be the default storage location. Otherwise disk is the default storage location. Selecting azure allows you to store baselines using your own instance of azure file storage.",
```
Mobrockers, line 123: The azure file storage option is actually called azurefilestorage, not azure. Perhaps it would be better to use an Enum instead of hard coded strings to avoid mistakes like these.

docs/Configuration.md
```diff
@@ -375,29 +374,64 @@ Default `"off"`
...
+## Baseline Storage location
+Sets the storage location for the baseline used by --dashboard-compare. By default this is set to disk, when the dashboard reporter is enabled this is automatically set to Stryker Dashboard.
+
+Supported storage locations are:
+
+| Storage location | Option | Description |
+|------------------|--------|-------------|
+| Disk             | disk   | Saves the baseline to the `StrykerOutput` folder|
+| Stryker Dashboard| dashboard | Saves the baseline to Stryker Dashboard |
+| Azure File Storage | azure | Saves the baseline to Azure File Storage |
```
Mobrockers, line 394: This option is not the option name used in StrykerOptions

docs/Configuration.md
```diff
@@ -375,29 +374,64 @@ Default `"off"`
...
+
+### Azure File Storage URL
+This is the url to your Azure File Storage. The URL should look something like this:
+
+```
+https://STORAGE_NAME.file.core.windows.net/FILE_SHARE/(optional)SUBFOLDER
+```
+Providing a subfolder is optional, we storage the baseline in a `StrykerOutput` subfolder.
+
+### Shared Access Signature (SAS)
```
Mobrockers, line 413: This example needs to be updated with the new option flags

Mobrockers:
@Stijn-Rutten Do you think you'll have time to finish this today?

src/Stryker.Core/Stryker.Core/Reporters/ReporterFactory.cs
```diff
@@ -12,12 +12,12 @@ namespace Stryker.Core.Reporters
 {
     public static class ReporterFactory
     {
-        public static IReporter Create(StrykerOptions options, IGitInfoProvider branchProvider = null)
+        public static IReporter Create(StrykerOptions options, IGitInfoProvider gitInfoProvider = null, IBaselineProvider baselineProvider = null)
```
Mobrockers, line 15: When are these not required?

src/Stryker.Core/Stryker.Core/Reporters/ReporterFactory.cs
```diff
@@ -12,12 +12,12 @@ namespace Stryker.Core.Reporters
...
         {
-            return new BroadcastReporter(DetermineEnabledReporters(options.Reporters.ToList(), CreateReporters(options, branchProvider)));
+            return new BroadcastReporter(DetermineEnabledReporters(options.Reporters.ToList(), CreateReporters(options, gitInfoProvider, baselineProvider: baselineProvider)));
         }
 
-        private static IDictionary<Reporter, IReporter> CreateReporters(StrykerOptions options, IGitInfoProvider branchProvider = null)
+        private static IDictionary<Reporter, IReporter> CreateReporters(StrykerOptions options, IGitInfoProvider gitInfoProvider = null, IBaselineProvider baselineProvider = null)
```
Mobrockers, line 20: When are these not required?

src/Stryker.Core/Stryker.Core/Baseline/DiskBaselineProvider.cs
```diff
@@ -0,0 +1,52 @@
...
+            _fileSystem = fileSystem ?? new FileSystem();
+            _logger = ApplicationLogging.LoggerFactory.CreateLogger<DiskBaselineProvider>();
+        }
+
+
+        public async Task<JsonReport> Load(string version)
+        {
+            var reportPath = Path.Combine(_options.BaselineOutputPath, version, "stryker-report.json");
+
+            using StreamReader inputReader = _fileSystem.File.OpenText(reportPath);
```
Mobrockers, line 30: What happens if the baseline does not exist yet?

src/Stryker.Core/Stryker.Core/Baseline/AzureBaselineProvider.cs
```diff
@@ -0,0 +1,179 @@
...
+using Stryker.Core.Reporters.Json;
+using System;
+using System.Net;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Stryker.Core.Baseline
+{
+    public class AzureBaselineProvider : IBaselineProvider
```
Mobrockers, line 14: Looking at this again it should probably be renamed to AzureFileShareBaselineProvider to avoid confusion.

src/Stryker.Core/Stryker.Core/Baseline/AzureBaselineProvider.cs
```diff
@@ -0,0 +1,179 @@
...
+                var successfullyAllocated = await AllocateFileLocation(byteSize, version);
+
+                if (!successfullyAllocated)
+                {
+                    return;
+                }
+            }
+            else
+            {
+                byteSize = Encoding.UTF8.GetByteCount(existingReport.ToJson());
```
Mobrockers, line 74: If a report already exists, you set the byteSize for the upload to the size of the old report. What happens if the new report is larger? Do the extra bytes just get chopped off resulting in a malformed report?

docs/Configuration.md
```diff
@@ -375,29 +374,64 @@ Default `"off"`
...
+```
+dotnet stryker --dashboard-compare --baseline-storage-location disk
+dotnet stryker -compare -bsl disk
+```
+Defaut `"disk"`
+
+## Configuring Azure File Storage
+When using Azure File Storage as baseline storage location you are required to provide the following values.
+
+### Azure File Storage URL
```
Mobrockers, line 405: This is missing the azure file storage url parameters

Stijn-Rutten: Please explain further. What url parameters?

src/Stryker.Core/Stryker.Core/Baseline/AzureBaselineProvider.cs
Stijn-Rutten: This is now fixed by doing an extra http call to azure

src/Stryker.Core/Stryker.Core/Baseline/AzureBaselineProvider.cs
Stijn-Rutten: Applied.

src/Stryker.Core/Stryker.Core/Baseline/DiskBaselineProvider.cs
Stijn-Rutten: Now handling FileNotFoundExceptions

src/Stryker.Core/Stryker.Core/Baseline/DiskBaselineProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+            var reportPath = Path.Combine(_options.BaselineOutputPath, version, "stryker-report.json");
+            try
+            {
+                using StreamReader inputReader = _fileSystem.File.OpenText(reportPath);
+
+                var reportJson = await inputReader.ReadToEndAsync();
+
+                return JsonConvert.DeserializeObject<JsonReport>(reportJson);
+            } 
+            catch(FileNotFoundException)
```
Mobrockers, line 37: The baseline not existing is not an unexpected failure situation, it probably happens quite often. Use File.Exists instead of a try-catch

docs/Configuration.md
Mobrockers: What commandline flag or json key should I set for the fileshare url? The docs only show an example of what the url looks like, not how to set it.

docs/Configuration.md
```diff
@@ -375,29 +374,64 @@ Default `"off"`
...
-3. Set --git-source to the name of the target branch of your pull request.
-4. (Optionally) Set a --dashboard-fallback-version. When you do not set a fallback version we will use --git-source as fallback. Since the source branch should be based on the target branche, this fallback baseline should be fairly up to date.
+3. Set --dashboard-fallback-version to the name of the target branch for your pull request.
+4. Set --git-source to the name of the target branch of your pull request.
 
 ```
-dotnet stryker --dashboard-compare --git-source master --dashboard-version development
-dotnet stryker -compare -source master -version development
+dotnet stryker --dashboard-compare --dashboard-fallback-version master -git-source master --dashboard-version development
+dotenet stryker -compare -fallback-version master -source master -version development
```
Mobrockers, line 435:
```suggestion
dotnet stryker -compare -source master -version development
```

docs/Configuration.md
```diff
@@ -375,29 +374,64 @@ Default `"off"`
...
 2. Set --dashboard-version to the name of the source branch for your pull request.
-3. Set --git-source to the name of the target branch of your pull request.
-4. (Optionally) Set a --dashboard-fallback-version. When you do not set a fallback version we will use --git-source as fallback. Since the source branch should be based on the target branche, this fallback baseline should be fairly up to date.
+3. Set --dashboard-fallback-version to the name of the target branch for your pull request.
+4. Set --git-source to the name of the target branch of your pull request.
 
 ```
-dotnet stryker --dashboard-compare --git-source master --dashboard-version development
-dotnet stryker -compare -source master -version development
+dotnet stryker --dashboard-compare --dashboard-fallback-version master -git-source master --dashboard-version development
```
Mobrockers, line 434:
```suggestion
dotnet stryker --dashboard-compare --git-source master --dashboard-version development
```

docs/Configuration.md
```diff
@@ -375,29 +374,64 @@ Default `"off"`
...
-You start a pull request to merge your finished feature back into development. Since this is a new feature and a new pull request, you will not yet have a report for your feature branch and the fallback version is used instead.
-Since your feature branch is branched off of development, a report for development will be fairly up-to-date with your feature branch, and you will only have to test the new mutations added in your feature branch. All other mutants can be skipped, because we have an up-to-date report for the development branch.
+When enabling the --dashboard-compare feature you can provide a fallback version. This version will be used to pull a baseline when we cannot find a baseline for your current branch. When we are still unable to provide a baseline we will start a complete testrun to create a complete baseline.
 
 ```
 dotnet stryker --dashboard-compare --dashboard-fallback-version master
 dotnet stryker -compare -fallback-version master
 ```
-Default: `--git-source`
+Default: `null`
```
Mobrockers, line 383:
```suggestion
Default: value provided to --git-source or null
```

Mobrockers:
https://github.com/stryker-mutator/stryker-net/blob/7e3c51ba186129e9f22093dc14c04e381048bd08/src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs#L234

The option in the docs and the commandline flag are called dashboard-version not project version. Internally we call it the project versiom but this error message is not clear to users.

Mobrockers:
`[18:43:21 INF] We could not locate a baseline for version null, now trying fallback version master` 

Since this is informational logging that the user always gets to see we should log useful values instead of 'null'.

Mobrockers:
A failure to create the directory on the fileshare logs the complete url including SAS, this means a failure can leak the SAS to for example build logs:

```
[18:45:36 ERR] Creating directory failed with status Forbidden and message Method: PUT, RequestUri: 'https://xxxxxx.file.core.windows.net/reports/strykerOutput/baselines/master?restype=directory&sv=<SAS_HERE>', Version: 1.1, Content: <null>, Headers:
{
  x-ms-file-permission: inherit
  x-ms-file-attributes: None
  x-ms-file-creation-time: now
  x-ms-file-last-write-time: now
}
```

It's probably a good idea to remove the SAS from the logstring before writing the log line.

src/Stryker.Core/Stryker.Core/Baseline/AzureFileShareBaselineProvider.cs
```diff
@@ -0,0 +1,212 @@
...
+                }
+            }
+
+
+            await UploadFile(reportJson, byteSize, version);
+        }
+
+        public async Task<bool> CreateDirectory(string version)
+        {
+            var subdir = $"strykerOutput/baselines/{version}";
```
Mobrockers, line 87: When trying out the feature I get a notfound. When trying to create the directories strykerOutput/baselines in the portal it says I can't have a / in the name so it looks like you have to create the sub-directories one-by-one.

src/Stryker.Core/Stryker.Core/Baseline/DiskBaselineProvider.cs
Mobrockers: Also, the first time you run dashboard compare the directory Baselines\dashboard-compare\<version> itself does not exist yet, so a catch on FileNotFoundException is not sufficient.

src/Stryker.Core/Stryker.Core/MutantFilters/DiffMutantFilter.cs
```diff
@@ -20,232 +20,247 @@ namespace Stryker.Core.MutantFilters
...
+            return mutants.Where(x =>
+                x.Mutation.OriginalNode.ToString() == baselineMutantSourceCode &&
+                x.Mutation.DisplayName == baselineMutant.MutatorName);
+        }
+
+        private async Task<JsonReport> GetBaselineAsync()
+        {
+            var branchName = _gitInfoProvider.GetCurrentBranchName();
 
             var baselineLocation = $"dashboard-compare/{branchName}";
```
Mobrockers, line 175: Use Path.Combine to avoid wrong paths on different platforms.

src/Stryker.Core/Stryker.Core/Reporters/GitBaselineReporter.cs
```diff
@@ -0,0 +1,46 @@
...
+        public GitBaselineReporter(StrykerOptions options, IBaselineProvider baselineProvider = null, IGitInfoProvider gitInfoProvider = null)
+        {
+            _options = options;
+            _baselineProvider = baselineProvider ?? BaselineProviderFactory.Create(options);
+            _gitInfoProvider = gitInfoProvider ?? new GitInfoProvider(options);
+        }
+        public void OnAllMutantsTested(IReadOnlyInputComponent reportComponent)
+        {
+            var mutationReport = JsonReport.Build(_options, reportComponent);
+            var projectVersion = _gitInfoProvider.GetCurrentBranchName() ?? _options.FallbackVersion;
```
Mobrockers, line 26: Baselines are loaded using this location `$"dashboard-compare/{branchName}"`

This means in the best case scenario you can the fallback version every test run and in the worst case scenario (no fallback) this feature never actually works.

Mobrockers:
With the disk baseline reporter, reports are showing up in git results.

This is because currently we don't place a gitignore in the StrykerOutputs folder but rather in the individual testrun folders. This is imo kinda inefficient as we only really need to place one gitignore in the root stryker output folder. Changing the folder where we place the gitignore to the StrykerOutput folder also fixes the baseline files showing up as changed files.

Creating the gitignore is currently done here:
It is currently done here: https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/StrykerRunner.cs#L54

We should move it to here:
https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs#L234

src/Stryker.Core/Stryker.Core/Baseline/AzureFileShareBaselineProvider.cs
```diff
@@ -0,0 +1,212 @@
...
+            _logger.LogInformation("Creating directory {0}", subdir);
```
Mobrockers, line 88: This doesn't need to be informational logging

src/Stryker.Core/Stryker.Core/Baseline/AzureFileShareBaselineProvider.cs
```diff
@@ -0,0 +1,212 @@
...
+            }
+
+            _logger.LogError("Creating directory failed with status {0} and message {1}", response.StatusCode.ToString(), response.RequestMessage);
+            return false;
+
+        }
+
+        private async Task<bool> AllocateFileLocation(int byteSize, string version)
+        {
+            _logger.LogInformation("Allocating storage for file");
```
Mobrockers, line 116: This doesn't need to be informational logging

src/Stryker.Core/Stryker.Core/Baseline/AzureFileShareBaselineProvider.cs
```diff
@@ -0,0 +1,212 @@
...
+            else
+            {
+                _logger.LogError("Azure File Storage upload failed with statuscode {0} and message: {1}", response.StatusCode.ToString(), await response.Content.ReadAsStringAsync());
+                return false;
+            }
+        }
+
+        private async Task UploadFile(string report, int byteSize, string version)
+        {
+            _logger.LogInformation("Uploading file to azure file storage");
```
Mobrockers, line 146:
```suggestion
            _logger.LogInformation("Uploading baseline report to azure file storage");
```

Mobrockers, line 146: This doesn't need to be logged here and not as information. Place it as debug logging before the SendAsync and add the url of the file share without sas.

src/Stryker.Core/Stryker.Core/Baseline/AzureFileShareBaselineProvider.cs
```diff
@@ -0,0 +1,212 @@
...
+
+            var response = await _httpClient.SendAsync(requestMessage);
+
+            if (response.StatusCode != HttpStatusCode.Created)
+            {
+                _logger.LogError("Azure File Storage upload failed with statuscode {0} and message: {1}", response.StatusCode.ToString(), await response.Content.ReadAsStringAsync());
+            }
+            else
+            {
+                _logger.LogInformation("Report uploaded");
```
Mobrockers, line 171:
```suggestion
                _logger.LogInformation("Uploaded baseline report to azure file share");
```

src/Stryker.Core/Stryker.Core/Baseline/AzureFileShareBaselineProvider.cs
```diff
@@ -0,0 +1,212 @@
...
+
+            var response = await _httpClient.SendAsync(requestMessage);
+
+            if (response.StatusCode != HttpStatusCode.OK)
+            {
+                _logger.LogError("Azure File Storage file resizing failed with statuscode {0} and message: {1}", response.StatusCode.ToString(), await response.Content.ReadAsStringAsync());
+                return false;
+            } else
+            {
+                _logger.LogInformation("File resized");
```
Mobrockers, line 198: This doesn't need to be informational logging

src/Stryker.Core/Stryker.Core/Baseline/AzureFileShareBaselineProvider.cs
```diff
@@ -0,0 +1,212 @@
...
+            }
+            else
+            {
+                _logger.LogInformation("Report uploaded");
+            }
+        }
+
+        private async Task<bool> ResizeFile(int byteSize, string version)
+        {
+            _logger.LogInformation("Updating file size to the size of the current report");
```
Mobrockers, line 177: This doesn't need to be informational logging

src/Stryker.Core/Stryker.Core/Baseline/DiskBaselineProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            var reportJson = report.ToJson();
+
+            _fileSystem.Directory.CreateDirectory(reportPath);
+
+            using StreamWriter outputWriter = _fileSystem.File.CreateText(Path.Combine(reportPath, $"stryker-report.json"));
+
+            await outputWriter.WriteAsync(reportJson);
+
+            _logger.LogDebug($"Baseline report has been writter to {Path.Combine(reportPath, $"stryker-report.json")}");
```
Mobrockers, line 57: Also add an informational variant of this logging that simply says the baseline has been saved. Or remove that same logging from the azure file share provider so they are consistent in their informational logging.

src/Stryker.Core/Stryker.Core/Reporters/GitBaselineReporter.cs
Mobrockers, line 26: Shouldn't this use the dashboard version and not the fallback version if the current branch name can't be determined?

src/Stryker.Core/Stryker.Core/Reporters/ReporterFactory.cs
Mobrockers: As discussed before, these null defaults can be removed

src/Stryker.Core/Stryker.Core/Reporters/ReporterFactory.cs
Mobrockers: As discussed before, these null defaults can be removed

Mobrockers:
When updating the baseline report in azure file storage I get the following error:

Mobrockers:
Is it expected that I get the below results in parts?

```
[22:42:39 INF] 37 mutants got status Killed. Reason: Result based on previous run.
[22:42:39 INF] 5 mutants got status Killed. Reason: Result based on previous run.
[22:42:39 INF] 13 mutants got status Killed. Reason: Result based on previous run.
[22:42:39 INF] 3 mutants got status NoCoverage. Reason: Result based on previous run.
[22:42:39 INF] 4 mutants got status NoCoverage. Reason: Result based on previous run.
[22:42:40 INF] 8 mutants got status NotRun. Reason: Could not determine the correct mutant status
[22:42:40 INF] 81 mutants got status Survived. Reason: Result based on previous run.
[22:42:40 INF] 14 mutants got status NotRun. Reason: Could not determine the correct mutant status
[22:42:40 INF] 34 mutants got status Killed. Reason: Result based on previous run.
[22:42:40 INF] 29 mutants got status NotRun. Reason: Could not determine the correct mutant status
[22:42:40 INF] 91 mutants got status NoCoverage. Reason: Result based on previous run.
[22:42:40 INF] 32 mutants got status NotRun. Reason: Could not determine the correct mutant status
[22:42:40 INF] 42 mutants got status Killed. Reason: Result based on previous run.
[22:42:40 INF] 2 mutants got status NotRun. Reason: Could not determine the correct mutant status
[22:42:40 INF] 9 mutants got status NoCoverage. Reason: Result based on previous run.
[22:42:40 INF] 20 mutants got status Survived. Reason: Result based on previous run.
[22:42:40 INF] 339 mutants got status Killed. Reason: Result based on previous run.
```

Mobrockers:
There are a lot of inconsistencies in naming for the 'location' or 'version' of the baseline reports being saved.

The dashboard baseline provider uses just the `<branchName>` to save and load the baseline.
The azure file share baseline provider uses `strykerOutput/baselines/<branchName>` to save the report and `strykerOutput/baselines/dashboard-compare/<branchName>` to load the baseline.
The disk baseline provider uses `StrykerOutput/Baselines/<branchName>` to save the report and `StrykerOutput/Baselines/dashboard-compare<branchName>` to load the report.

These could probably do with all being the same format and the same capitalization for consistency sake. Perhaps put the `dashboard-compare-report-path` in a calculated stryker option or something.

Mobrockers:
Something is very wrong with the mutant got status logging.

This is the logging in 0.18:
```
[14:18:00 INF] 10 mutants got status CompileError. Reason: Could not compile
[14:18:00 INF] 11 mutants got status Ignored. Reason: Removed by exclude from code coverage filter
```

This is the logging from this pr when no baseline exists
```
[14:09:57 INF] 31 mutants got status NoCoverage. Reason: null
[14:09:57 INF] 10 mutants got status NoCoverage. Reason: Could not compile
```

This is the logging when a baseline is used to turn off mutants we don't need to test:

```
[22:42:39 INF] 37 mutants got status Killed. Reason: Result based on previous run.
[22:42:39 INF] 5 mutants got status Killed. Reason: Result based on previous run.
[22:42:39 INF] 13 mutants got status Killed. Reason: Result based on previous run.
[22:42:39 INF] 3 mutants got status NoCoverage. Reason: Result based on previous run.
[22:42:39 INF] 4 mutants got status NoCoverage. Reason: Result based on previous run.
[22:42:40 INF] 8 mutants got status NotRun. Reason: Could not determine the correct mutant status
[22:42:40 INF] 81 mutants got status Survived. Reason: Result based on previous run.
[22:42:40 INF] 14 mutants got status NotRun. Reason: Could not determine the correct mutant status
[22:42:40 INF] 34 mutants got status Killed. Reason: Result based on previous run.
[22:42:40 INF] 29 mutants got status NotRun. Reason: Could not determine the correct mutant status
[22:42:40 INF] 91 mutants got status NoCoverage. Reason: Result based on previous run.
[22:42:40 INF] 32 mutants got status NotRun. Reason: Could not determine the correct mutant status
[22:42:40 INF] 42 mutants got status Killed. Reason: Result based on previous run.
[22:42:40 INF] 2 mutants got status NotRun. Reason: Could not determine the correct mutant status
[22:42:40 INF] 9 mutants got status NoCoverage. Reason: Result based on previous run.
[22:42:40 INF] 20 mutants got status Survived. Reason: Result based on previous run.
[22:42:40 INF] 339 mutants got status Killed. Reason: Result based on previous run.
```

src/Stryker.Core/Stryker.Core/MutantFilters/DiffMutantFilter.cs
```diff
@@ -20,232 +20,247 @@ namespace Stryker.Core.MutantFilters
...
+                .OrderBy(x => x.Key);
 
-        private IEnumerable<Mutant> SetMutantStatusForFileChanged(IEnumerable<Mutant> mutants)
-        {
-            foreach (var mutant in mutants)
+            foreach (var skippedMutantGroup in mutantGroups)
             {
-                mutant.ResultStatus = MutantStatus.NotRun;
-                mutant.ResultStatusReason = "File changed since last commit.";
+                _logger.LogInformation("{0} mutants got status {1}. Reason: {2}", skippedMutantGroup.Count(),
```
Mobrockers, line 127: What is the reason for logging this here as well as in MutationTestProcess?

Mobrockers:
There are some discrepancies between the pre- and post-baseline reports:

Pre:
image

Post:
image

Mobrockers:
@dupdob If you have time could you try out this new feature?

richardwerkman:
I think we should still test all mutants if a "non source" file is changed. This file could be used by a test. For example, a test could load a json file and use its contents. Thus the test has changed, even though git won't say it has changed. 

Just to be sure we should still test all mutants if another file than a `.cs` file is changed.

Stijn-Rutten:
@richardwerkman makes sense. Maybe we could look into ways to determine which tests would be affected by these changes. For example, if there is a `assets/mockdata.json` file that has changed  it would be nice to check for tests that load this file. However I can imagine that this would be something quite complex. For now let's implement testing all mutants when a none `.cs` file is changed. I will create an issue and look into this when I can find some time.

Mobrockers:
I just noticed that the mutation coverage is 0% when turning off coverage analysis on this branch. This does not seem correct to me. @Stijn-Rutten Do you have any idea why this would happen?

Stijn-Rutten:
@richardwerkman 
The changes in 1a7bbdf check if the file extension does not equal `.cs`.

src/Stryker.Core/Stryker.Core/MutantFilters/DiffMutantFilter.cs
```diff
@@ -203,6 +210,17 @@ private IEnumerable<Mutant> SetMutantStatusForFileChanged(IEnumerable<Mutant> mu
             return mutants;
         }
 
+        private IEnumerable<Mutant> SetMutantStatusForNonCSharpFileChanged(IEnumerable<Mutant> mutants)
+        {
+            foreach (var mutant in mutants)
+            {
+                mutant.ResultStatus = MutantStatus.NotRun;
+                mutant.ResultStatusReason = "A file with a different extension than .cs was changed. Cannot determine influence of file change on this mutant";
```
Mobrockers, line 218: This reason is a bit long. Since it will be shown to the user in the console I would prefer to keep it short.

```suggestion
                mutant.ResultStatusReason = "Non-code test project files were changed.";
```

Mobrockers:
Work left to do:

 - [ ] Check why the unit test fails on *nix systems
 - [ ] Check results vs master with coverage analysis off
 - [ ] Check results vs master with compare off
 - [ ] Decide whether the results are acceptable enough to release this as experimental feature
 - [ ] Mark feature as experimental in docs and warn that results can differ

PR #1135, opened 2020-07-09
chore: Fix exclude files example in docs

richardwerkman:


Mobrockers:
Ouch 👍

richardwerkman:
Nobody ever noticed so I guess the feature is not so popular 🤔

PR #1151, opened 2020-07-24
fix(Multi test projects): Fix broken multi test project feature

richardwerkman:
The multi test project feature was broken since V0.18.0. This was probably done during a refactor. This change fixes this issue.

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
+++ b/src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
@@ -76,7 +76,14 @@ namespace Stryker.Core.Initialisation
             projectInfo.TestProjectAnalyzerResults = testProjectAnalyzerResults;
 
             // Determine project under test
-            projectUnderTest = FindProjectUnderTest(projectInfo.TestProjectAnalyzerResults, options.ProjectUnderTestNameFilter);
+            if (options.TestProjects != null && options.TestProjects.Any())
+            {
+                projectUnderTest = FindProjectFile(options.BasePath);
+            }
+            else
+            {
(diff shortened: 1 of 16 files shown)


src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -76,7 +76,14 @@ public ProjectInfo ResolveInput(StrykerOptions options)
             projectInfo.TestProjectAnalyzerResults = testProjectAnalyzerResults;
 
             // Determine project under test
-            projectUnderTest = FindProjectUnderTest(projectInfo.TestProjectAnalyzerResults, options.ProjectUnderTestNameFilter);
+            if (options.TestProjects != null && options.TestProjects.Any())
+            {
+                projectUnderTest = FindProjectFile(options.BasePath);
```
Mobrockers, line 81: Using disk scanning to find the project file when one or more test projects have been configured in settings means that we implicitly expect the user to run this configuration from the project under test. However, we never test if this actually happens, and we log no warning or error to the user when they apply this wrong configuration.

The result is that the mutation testrun succeeds, but the user will have tested their test project instead of their project under test.

We should either make it explicit that this feature only works from the project under test (and throw an exception if we detect we're being run from a test project), or fix the previous behavior to correctly find the project under test based on the test project references.

In fact we should probably investigate why FindProjectUnderTest using test project references is not working correctly anyway, because this bug might also affect people that do not use the test project option.

richardwerkman: I'll look into why the FindProjectUnderTest method is not working as expected. If I dont find why I will add a exception if the feature is not used right to avoid confusion

richardwerkman:
I can't seem to reproduce the bug anymore. So I'll close this PR until I need it again.

Issue #868, opened 2020-01-10
Reorganize mutation placing


Mobrockers:
**Is your feature request related to a problem? Please describe.**
Currently mutation placing decision logic is contained in https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs

This is a huge class which decides for each piece of syntax in what way a mutation should be placed. 

**Describe the solution you'd like**
If we reorganize this into smaller components in the same way as the mutators are organized, we could improve readability and make it easier to introduce components for new syntax kinds that need to be treated in a specific way. For example: the contents of switch statements cannot currently be mutated because the cases need to be compile-time static. We can solve this by placing the whole switch statement in an if-else structure. This would be easier to implement if you would only need to add a component that deals with switch constructions. Currently you would have to dig through the complete mutation placing logic to identify the correct place to implement the feature.

richardwerkman:
We have to keep in mind that this would introduce a lot of mutators. For example, the switch statement should be handeled differently than the switch expression. But what I like about this is that the mutant orchestrator doesn't need to know how to place mutations for switch statements or switch expressions. It just calls a mutator for that type and places the outcome in the syntax tree. 

We could add a new Dictionary in the MutantOrchestrator that has node types as keys. Then all the mutant orechestrator has to do is check the dictionary for each node. If the dictionary contains the key, a mutator for that type is available. It calls the mutator with the node and places the mutated node and stores the returned mutants. 

This way we can slowly shrink the Orchestrator by introducing new mutators and removing specific logic from the Orchestrator.

dupdob:
Sounds like a good idea. 

Just a quick note: I think current logic can be trimmed down.
One thing that annoys me is that Stryker is not able to automatically use an if construction when it can't use a conditional operator. This is really tricky to implement, but the main reason we need statement specific logic is that we need to use if patterns for some language constructs.

dupdob:
just a quick note to say that I start to work on this

dupdob:
I think this issue can be safely closed now.

Mobrockers:
Yep :) Thank you!

PR #1166, opened 2020-08-07
Refactor mutation logic

dupdob:
Refactor the design of mutation logic as per  #868. Added per SyntaxConstruction mutation logic to streamline the MutationOrchestrator logic.

Here is the new mutation logic design in a nutshell:
- the `MutantOrchestrator` class is responsible for:
  - referencing all (configured) strategies for mutation (mutator classes)
  - referencing SyntaxNode specific strategies for mutation orchestration (`NodeSpecificOrchestrator` implementations)
  - provide public entry point for the mutant generation process
- the `MutationContext` is responsible for:
  - storing the (current) state of the mutation when iterating through the syntax tree (e.g. knowing when inside a `static` element).
  - provide a mutation entry point for `NodeSpecificOrchestrator`: `MutateNodeAndChildren(SyntaxNode node, bool statementLevelControlled)` which mutates the given node and its children and return the mutated node. `statementLevelControlled` allow to force the node's mutations to be controlled at the statement (or block) level instead of the expression level (i.e. using `if` instead of a ternary operator)
 - the various `NodeSpecificOrchestrator` implementations are responsible for handling specific syntax node types (e.g: `for` statements or static constructors). Their duties are to update the mutation context (if relevant) and decide how mutations are controlled (if relevant).
- `Mutators` implementations are responsible to generate the various mutations of the syntax constructs they are in charge of


Mobrockers:
I'll take some hours tomorrow to review this 👍

dupdob:
Close this PR to reopen one without the last commit, as per Slack discussion

Issue #1038, opened 2020-04-17
linq mutators for Union, Intersect, Concat & Except


jnyrup:
I was wondering how to mutate code that involved the following LINQ methods

* `Union()`
* `Intersect()`
* `Concat()`
* `Except()`

One idea was this mapping

Original | Mutated
-- | --
`Union()` | `Intersect()`
`Intersect()` | `Union()`
`Concat()` | `Except()`
`Except()` | `Concat()`

Alternatively would it be within the scope of Stryker or possible at all, to mutate the `IEnumerable<T>` parameter with `Enumerable.Empty<T>()`?

richardwerkman:
I Like these additions to the linq mutator. I'm not sure what your plan is with the `Enumerable.Empty<T>()`. Should it replace the whole linq query?

If you'd like to help us more with our mutators you could check out this issue: https://github.com/stryker-mutator/stryker-handbook/issues/33

jnyrup:
@richardwerkman 
My idea with `Enumerable.Empty<T>()` was to introduce two mutants of e.g. `a.Concat(b)`:
* `Enumerable.Empty<T>().Concat(b)`
* `a.Concat(Enumerable.Empty<T>())`

I had a look at the other issue, but I don't really have any opinions on mutant levels, except that it sounds useful.

richardwerkman:
That could prove hard as these methods are not always directly called on the list itself. They are often called like:

`a.ToList().Concat(b);`

which would become: 

`a.Enumerable.Empty<T>().Concat(b)`

That would not compile of couse. I think swapping the methods with their "opposite" method does the job as well and is a much easier mutation. So I'd rather go for your first suggestion.

And thanks for taking a look at the other issue, good to know it sounds useful.

PR #1182, opened 2020-08-31
feat(Added linq mutator): Added linq Union and intersect mutator.

Myoxocephalus:
Added linq mutators as proposed in https://github.com/stryker-mutator/stryker-net/issues/1038


richardwerkman:
Thank you for your contribution! I do have one question. Why did you only add these mutations and not all suggested mutations in the issue? If you feel like it you could add the other two suggested mutations as well.

Myoxocephalus:
Hi!
Thanks for the comment!
I think the other mutations makes sense as well. The reason i scoped this PR was mainly because im a new contributor to this project and wanted to start small and get some feedback.
I could create an additional PR for that other mutation.

richardwerkman:
Thanks for your contribution! If you like to add the other mutations as well, please open another PR 😃

PR #1183, opened 2020-09-01
fix: dont mutate const local declaration

### LINKED ISSUE #1181: methods containing const string compile error
Methods containing const string variables gets compile exception. The error detail shown below.

The stryker version is  "**version": "0.18.0**".
.net core version =  **2.2**

**Actual code is:**

public void TestMethod(SomeClass instance)
{
            const string query = "SELECT COL1, COL2 FROM TABLE (NOLOCK)" +
                "WHERE COL1 = @COL_NAME";
}

[15:27:16 WRN] Stryker.NET encountered an compile error in C:\Git\someCode.cs (at 31:33) with message: The expression being assigned to 'query' must be constant (Source code: (Strykerl9bPlZSUPUE24og.MutantControl.IsActive(209)?"":"XXXX " )+
(Strykerl9bPlZSUPUE24og.MutantControl.IsActive(210)?"":                "YYYYYY" )+
(Strykerl9bPlZSUPUE24og.MutantControl.IsActive(211)?"":                "ZZZZ" )+
(Strykerl9bPlZSUPUE24og.MutantControl.IsActive(212)?"":                "AAAAA"))
[15:27:16 WRN] Safe Mode! Stryker will try to continue by rolling back all mutations in method. This should not happen, please report this as an issue on github with the previous error message.

dupdob:
Closes #1181

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/MutantOrchestratorTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/MutantOrchestratorTests.cs
@@ -688,6 +688,16 @@ namespace TestApp
         }
 
         [Fact]
+        public void ShouldNotMutateConstDeclaration()
+        {
+            var source = @"void Test(){
+const string text = ""a""+""b""+""c"";}";
+            var expected = @"void Test(){
+const string text = ""a""+""b""+""c"";}";
+            ShouldMutateSourceToExpected(source, expected);
+        }
+
+        [Fact]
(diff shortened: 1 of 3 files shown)


richardwerkman:
@dupdob thanks for fixing this so quickly!

dupdob:
@richardwerkman : don't mention. It was a great opportunity to reap rewards from the new design 😄.

Note: I am afraid there may still be issues with case labels, but only for weird construct, such as `case 2+2:` (yes, this is legal, I just checked it).

PR #1192, opened 2020-09-07
feat(Added linq mutator): Added linq Concat and Except mutator.

Myoxocephalus:
Second PR with suggestions from #1038


Mobrockers:
Awesome :)

PR #1199, opened 2020-09-11
fix(git diff): exclude stryker generated files from git diff

g7ed6e:


Mobrockers:
Hi!

Thank you for this pull request, however I don't think this is the best solution to this problem. Stryker generated files should all be hidden behind a gitignore file, however libgit2sharp seems to not respect this gitignore file. The best solution IMO would be for libgit2sharp to respect the gitignore file. Do you think perhaps you could look into that instead of this fix?

Mobrockers:
I will allow this as a quick fix but we need a proper fix for this issue.

PR #1206, opened 2020-09-14
feat(Dashboard Compare): Add non-source file pattern exclusion

### LINKED ISSUE #1165: Determine affected mutants by changes in "non source" file
Depends on #1101 

When using the dashboard compare feature and a none source file is changed there is no way of knowing which tests are affected by this. Thus all mutants have to be tested.

I have not looked into any concrete solutions but we should look into ways to determine which parts of the source code or the tests have been affected by these changes. That way we only have to enable the mutants in these parts of the source code.

Stijn-Rutten:
Enables users using the dashboard compare feature to specify a list of excluded non-c# files. Whenever these files are changed we will not mark all mutants for testing. Instead we will continue as if no 'non-source files' were changed.


Mobrockers:
Let's not add the span feature, I'm not dealing wth that can of worms for non source files.

Stijn-Rutten:
@Mobrockers Please elaborate what you would like to see instead.

src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
```diff
@@ -116,6 +116,22 @@ public static class CLIOptions
 
         };
 
+        public static readonly CLIOption<string[]> DashboardCompareFileExcludePatterns = new CLIOption<string[]>
+        {
+            ArgumentName = "--dashboard-compare-excluded-non-source-files", // TODO: Really needs a better name
```
Stijn-Rutten, line 121: Any ideas for a better argument name?

src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
```diff
@@ -53,6 +53,8 @@ public class StrykerOptions
         public string ModuleName { get; }
         public string ProjectVersion { get; }
 
+        public IEnumerable<FilePattern> DashboardCompareFileExcludePatterns { get; } // TODO: Should use a better name
```
Stijn-Rutten, line 56: Also could use a shorter name.

Mobrockers:
You can exclude files and folders, but not contents of files. Either the file has impact on tests or it does not.

Stijn-Rutten:
Would you rather see me change the use of `FilePattern` to `Dotnet.Globbing.Glob`?

Mobrockers:
> Would you rather see me change the use of `FilePattern` to `Dotnet.Globbing.Glob`?

I'm talking about this part in the docs:  Use '{< start > .. < end >}' at the end of a pattern to specify spans of text in files to in- or exclude.

src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
Mobrockers: For starters this feature shouldn't only work for the dashboard compare but also the regular diff feature.

How about `--diff-ignore-files`
And allow this to be used to also ignore changes in source files.

src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
```diff
@@ -213,15 +216,25 @@ private string ValidateGitSource(string gitSource)
             return gitSource;
         }
 
-        private (string ProjectVersion, string FallbackVersion, string GitSource) ValidateCompareToDashboard(string projectVersion, string fallbackVersion, string gitSource)
+        private (string ProjectVersion, string FallbackVersion, string GitSource, IEnumerable<FilePattern> DashboardCompareFileExcludePatterns) ValidateCompareToDashboard(string projectVersion, string fallbackVersion, string gitSource, IEnumerable<string> dashboardCompareFileExcludePatterns)
```
Mobrockers, line 219: I don't think the exclude from diff result is exclusive to dashboard compare so it should have it's own validator

src/Stryker.Core/Stryker.Core/MutantFilters/DiffMutantFilter.cs
```diff
@@ -78,23 +78,33 @@ public IEnumerable<Mutant> FilterMutants(IEnumerable<Mutant> mutants, FileLeaf f
                 UpdateMutantsWithBaselineStatus(mutants, file);
             }
 
-            // A non-cSharp file is flagged by the diff result as modified. We cannot determine which mutants will be affected by this, thus all mutants have to be tested.
-            if (_diffResult.TestFilesChanged is { } && _diffResult.TestFilesChanged.Any(x => !x.EndsWith(".cs")))
+
+
+            // A non-cSharp file is flagged by the diff result as modified. We cannot determine which mutants will be affected by this, thus all mutants have to be tested, except when the changed files are excluded.
+            if (_diffResult.TestFilesChanged is { })
```
Mobrockers, line 84: I don't think this option should be exclusive to the diff mutant filter, it might be better to do the filtering in diffProvider.ScanDiff() and simply not return any files that match the pattern (test AND source files).

src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
```diff
@@ -116,9 +116,9 @@ public static class CLIOptions
 
         };
 
-        public static readonly CLIOption<string[]> DashboardCompareFileExcludePatterns = new CLIOption<string[]>
+        public static readonly CLIOption<string[]> DiffIgnoreFiles = new CLIOption<string[]>
         {
-            ArgumentName = "--dashboard-compare-excluded-non-source-files", // TODO: Really needs a better name
+            ArgumentName = "--diff-ignore-files",
             ArgumentShortName = "-compare-excluded",
```
Mobrockers, line 122: You also need to update the shortname

src/Stryker.CLI/Stryker.CLI/CLIOptions.cs
```diff
@@ -116,9 +116,9 @@ public static class CLIOptions
...
             ArgumentDescription = @"Allows to specify an array of non-C# files that you want to exclude from testing.
```
Mobrockers, line 123:
```suggestion
            ArgumentDescription = @"Allows to specify an array of C# files which should be ignored if present in the diff.
```

Issue #1170, opened 2020-08-18
Stryker generates duplicate mutations


dupdob:
**Describe the bug**
The 'NegateConditionMutator' and the `BinaryExpressionMutator' often result in duplicate mutation,e.g:
original code:
```csharp
if (x == 0)
```
NegateConditionMutator gives
```csharp
if (!(x == 0))
```
BinaryExpressionMutator gives
```chsarp
if (x!=0)
```
Both mutations are equivalent and mutant outcomes are the same. This result is biased statistics and supplemental test time.

**Expected behavior**
Stryker should not create equivalent mutations.
One possible solutions would be to:
- modify 'NegateConditionMutator' so it properly negates binary expression
- add duplicate detection logic so that Stryker no longer injects (obvious) duplicate.

This approach is implemented in PR #1166

Mobrockers:
Did you mean PR #1166  ?

Mobrockers:
Are both approaches implemented in #1166 or just the first?

For the second approach I was thinking we could write a very crude mutation filter that goes over all mutations and finds mutations within the same file/span and checks them against known equivalent mutations.

dupdob:
> Did you mean PR #1166 ?
Yes, I have no idea I ended up using the other PR's id instead. Thanks

dupdob:
I implemented both in PR #1166. What you suggest is exactly what I have done: MutationOrchestration node apply this crude filter when generating mutation (and logs when discarding a duplicate).
And I modified `NegateConditionMutator` so that mutations could match the outcome of `BinaryExpressionMutator`

dupdob:
z

Mobrockers:
> I implemented both in PR #1166. What you suggest is exactly what I have done: MutationOrchestration node apply this crude filter when generating mutation (and logs when discarding a duplicate).
> And I modified `NegateConditionMutator` so that mutations could match the outcome of `BinaryExpressionMutator`

Ah cool! I wonder if that's the best place for this filtering but we can always move it if we notice we can identify a lot of different equivalent mutations later on 👍

dupdob:
As per discussion on PR #1166, if we cannot change the outcome of mutators, two solutions are still possible:
1) prevent known duplicates to happen in the first place. For example, limit the NegateConditionMutator so that it does not mutate binary expressions
2) improve the mutant detection logic so it can recognise (some) known duplicates, such as a != b vs !(a==b).

[...]

Issue #1171, opened 2020-08-19
Stryker does not mark expression bodied static constructor for coverage capture


dupdob:
**Describe the bug**
Stryker injects marking logic in static constructors (and properties) to identify mutant coverage. Indeed, so called 'static mutants' need to be tested against all tests as one cannot identify dependences through coverage data.
But as of V0.18 (and previous), Stryker does not mark expression bodied properties. This results in potentially incorrect coverage data hence incorrect test results. 

**Expected behavior**
Stryker should be able to mark every static constructors/properties, disregarding its form. Note this implies converting expression bodied construct to full body variant.

Mobrockers:
You say as of v0.18. Is this because we introduced a breaking change there or because the feature didn't exist before v0.18?

dupdob:
No regression here. I updated my comment. Sorry for the ambiguity.

PR #1236, opened 2020-10-05
Improve mutation orchestration and rollback logic

dupdob:
Redesigned the rollback as well as mutation orchestration.
As a bonus, fixes issue #1170 and #1171 .


src/Stryker.Core/Stryker.Core/Helpers/TypeBasedStrategy.cs
```diff
@@ -33,23 +33,27 @@ public THandler FindHandler(T item)
 
         private THandler FindHandler(T item, Type type)
         {
-            if (item == null || type == null)
+            for (;;)
```
Mobrockers, line 36:
```suggestion
            while(true)
```

src/Stryker.Core/Stryker.Core/Helpers/TypeBasedStrategy.cs
```diff
@@ -33,23 +33,27 @@ public THandler FindHandler(T item)
 
         private THandler FindHandler(T item, Type type)
```
Mobrockers, line 34: To me the previous style of recursion is more readable. @richardwerkman what do you think

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
```diff
@@ -286,7 +286,7 @@ public void FilterMutants()
             }
 
             var skippedMutants = _input.ProjectInfo.ProjectContents.ReadOnlyMutants.Where(m => m.ResultStatus != MutantStatus.NotRun);
-            var skippedMutantGroups = skippedMutants.GroupBy(x => new { x.ResultStatus, x.ResultStatusReason }).OrderBy(x => x.Key.ResultStatusReason);
+            var skippedMutantGroups = skippedMutants.GroupBy(x => new { x.ResultStatus, x.ResultStatusReason }).OrderBy(x => x.Key.ResultStatus);
```
Mobrockers, line 289: This change means that skipped mutants that have the same result status reason (eg when using dashboard compare to reuse previous results) but different result statuses will no longer be ordered together.

@richardwerkman would you prefer

```
[21:34:40 INF] 11    mutants got status Ignored.      Reason: Removed by exclude from code coverage filter
[21:34:40 INF] 25    mutants got status Ignored.      Reason: Result based on previous run
[21:34:40 INF] 15    mutants got status NoCoverage.   Reason: Mutant has no coverage
[21:34:40 INF] 16    mutants got status NoCoverage.   Reason: Result based on previous run
[21:34:40 INF] 201   mutants got status Killed.       Reason: Result based on previous run
[21:34:40 INF] 100   mutants got status Survived.     Reason: Result based on previous run
[21:34:40 INF] 10    mutants got status CompileError. Reason: Result based on previous run
[21:34:40 INF] 3     mutants got status Timeout.      Reason: Result based on previous run
```

Or:

```
[21:34:40 INF] 11    mutants got status Ignored.      Reason: Removed by exclude from code coverage filter
[21:34:40 INF] 15    mutants got status NoCoverage.   Reason: Mutant has no coverage
[21:34:40 INF] 201   mutants got status Killed.       Reason: Result based on previous run
[21:34:40 INF] 16    mutants got status NoCoverage.   Reason: Result based on previous run
[21:34:40 INF] 25    mutants got status Ignored.      Reason: Result based on previous run
[21:34:40 INF] 100   mutants got status Survived.     Reason: Result based on previous run
[21:34:40 INF] 10    mutants got status CompileError. Reason: Result based on previous run
[21:34:40 INF] 3     mutants got status Timeout.      Reason: Result based on previous run
```

src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
```diff
@@ -0,0 +1,46 @@
...
+        // gets the return type of a method (incl. constructor, destructor...)
+        public static TypeSyntax ReturnType(this BaseMethodDeclarationSyntax baseMethod)
+        {
+            return baseMethod switch
+            {
+                ConstructorDeclarationSyntax _ => null,
+                ConversionOperatorDeclarationSyntax _ => null,
+                DestructorDeclarationSyntax _ => null,
+                MethodDeclarationSyntax method => method.ReturnType,
+                OperatorDeclarationSyntax oper => oper.ReturnType,
```
Mobrockers, line 20:
```suggestion
                OperatorDeclarationSyntax operator => operator.ReturnType,
```

It's not that long ;)

src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
```diff
@@ -0,0 +1,46 @@
...
+                ConversionOperatorDeclarationSyntax _=> true,
+                _ => false
+            };
+        }
+
+        public static T InjectMutation<T>(this T original, Mutation mutation) where T:SyntaxNode
+        {
+            if (!original.Contains(mutation.OriginalNode))
+            {
+                throw new InvalidOperationException($"Cannot inject mutation '{mutation.ReplacementNode}' in {original}.");
```
Mobrockers, line 41:
```suggestion
                throw new InvalidOperationException($"Cannot inject mutation '{mutation.ReplacementNode}' in {original} because this node and mutation original node are not the same.");
```

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -103,7 +107,19 @@ public IReadOnlyCollection<Mutant> GetLatestMutantBatch()
         /// <returns>Mutated node</returns>
         public SyntaxNode Mutate(SyntaxNode currentNode)
         {
-            return Mutate(currentNode, new MutationContext(this));
+            var mutationContext = new MutationContext(this);
+            var mutation = Mutate(currentNode, mutationContext);
+            if (mutationContext.HasStatementLevelMutant && _options?.DevMode == true)
+            {
+                Log.Error($"Several mutants were not injected in the project : {mutationContext.BlockLevelControlledMutations.Count+mutationContext.StatementLevelControlledMutations.Count}");
```
Mobrockers, line 114: This uses serilog directly instead of through Microsoft.Extensions.Logging.ILogger, skipping our configured logger.

src/Stryker.Core/Stryker.Core/Mutants/MutantOrchestrator.cs
```diff
@@ -103,7 +107,19 @@ public IReadOnlyCollection<Mutant> GetLatestMutantBatch()
         /// <returns>Mutated node</returns>
         public SyntaxNode Mutate(SyntaxNode currentNode)
         {
-            return Mutate(currentNode, new MutationContext(this));
+            var mutationContext = new MutationContext(this);
+            var mutation = Mutate(currentNode, mutationContext);
+            if (mutationContext.HasStatementLevelMutant && _options?.DevMode == true)
```
Mobrockers, line 112: Could you add some comment about why this is bad but is only logged if it's a devmode run?

src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs
```diff
@@ -5,75 +5,111 @@
...
+        {
+            var engine = nodeToRemove.GetAnnotatedNodes(Injector).FirstOrDefault()?.GetAnnotations(Injector).First().Data;
+            if (!string.IsNullOrEmpty(engine))
+            {
+                return InstrumentEngines[engine].RemoveInstrumentation(nodeToRemove);
+            }
 
-        private static SyntaxNode RemoveByConditionalExpression(ParenthesizedExpressionSyntax parenthesized) => 
-            parenthesized.Expression is ConditionalExpressionSyntax conditional ? conditional.WhenFalse : null;
+            throw new InvalidOperationException($"Unable to remove any injection from this node: {nodeToRemove}");
```
Mobrockers, line 94: Could you add the reason why it wasn't possible to remove the injected mutant?

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
richardwerkman: I think it's both a bit much. We could also go for something like:

```
[21:34:40 INF] Based on your previous run 336 mutants go a status without testing (201 Killed, 100 Survived, 11 Ignored, 16 NoCoverage, 10 CompileError, 3 Timeout)
```
Also, is the CompileError really based on the previous run? I think we get the errors again before we use the baseline so we could skip them in printing here?

Mobrockers: That means specifically looking for the string 'Result based on previous run' and changing the output based on it. Currently this whole reporting piece is generic for all filters.

It only shows up if you use dashboard compare, and I would think a user would want clear information about what exactly was picked up by the comparison, especially since we've seen some small errors with this feature.

src/Stryker.Core/Stryker.Core/Helpers/TypeBasedStrategy.cs
richardwerkman: Yeah to me "traditional" recursion is more readable. I'm not sure about possible performance gains this way?

dupdob: I did not make any benchmarck to confirm my intuition, but I am concerned since this method is called for every syntax node of the project, and the SyntaxNode inheritance tree is 6 or 7 level deep. Transforming tail recursion to loop is a classic optimisation which typically bring a twofold performance increase for simple function, which is pretty much the case here as we simply try again with the parent class when failing to find a specialised handler.
Nonetheless, it is clear that this loop version was crude, so I implemented a proper for loop that looks far better.

If you still prefer a recursive one, I will revert to it, but feedback on the new version will be welcomed

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
dupdob: I see, the grouping appeared a bit strange to me. I can revert, but the problem is that the algorithm now reports non injected mutants as CompileError, so the sorting by 'Reason' first resulted in having them separated from actual compileError.

I am not sure what the proper design is, but it sounds to me that the 'reason' field has many different usage then

src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
dupdob: actually, you cannot use 'operator', as it is a keyword. I ended up using operatorSyntax (could have used @operator, but I think it could have been misleading). Anyway, this is really a trivial helper

src/Stryker.Core/Stryker.Core/Helpers/TypeBasedStrategy.cs
Mobrockers: Makes sense, we don't want any stackoverflows.

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
Mobrockers: Can you give an example of the output and expected output?

src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
Mobrockers: ah right

Mobrockers (approved):
Overall looks very good! :) 

Other than the minor changes already requested I think this can be merged as soon as I have done a comparison testrun unless @richardwerkman sees anything else worth mentioning

src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs
dupdob: As hinted earlier, I am not sure what the output should look like. For context: I have never tested the diff mode, so I never encountered the 'Result based on previous run' reason. Also, I was under the impression that 'StatusReason' purpose was to describe how Stryker reached to that status.
I think this deserve more discussion before reaching a conclusion, so I restored the order as it was, and I will open an issue for discussion.

Here is a simulated output of what can reported now, and I think this looks messy as 'CompileError' status is all over the place.

```[10:55:29 INF] Capture mutant coverage using 'perTest' mode.
[10:55:41 INF] 32    mutants got status CompileError. Reason: Mutant caused compile errors
[10:55:41 INF] 2     mutants got status NoCoverage.   Reason: Mutant has no test coverage
[10:55:41 INF] 11    mutants got status Ignored.      Reason: Removed by exclude from code coverage filter
[10:55:41 INF] 10    mutants got status CompileError. Reason: Result based on previous run
[10:55:41 INF] 201   mutants got status Killed.       Reason: Result based on previous run
[10:55:41 INF] 16    mutants got status NoCoverage.   Reason: Result based on previous run
[10:55:41 INF] 25    mutants got status Ignored.      Reason: Result based on previous run
[10:55:41 INF] 100   mutants got status Survived.     Reason: Result based on previous run
[10:55:41 INF] 3     mutants got status CompileError. Reason: Stryker was not able to inject mutation in code.
[10:55:41 INF] xxx   total mutants are skipped for the above mentioned reasons
[10:55:41 INF] yyyy  total mutants will be tested```

Mobrockers: I agree, we should think about what we're gonna do with this 👍

PR #1256, opened 2020-10-14
chore: Add issue template configuration

Mobrockers:
Gives us something like this:

image


Mobrockers:
@richardwerkman what do ya think

richardwerkman:
It looks good but didn't we want to direct questions towards Github Discussions? If so, we don't want to send people to slack also.

Mobrockers:
Do we want to completely move away from slack? I deliberately expressed it as talking (chatting) with the team on slack and asking questions and such to github discussions.

richardwerkman:
Sorry I oversaw the `Ask a Question` part. Looks good to me then

Issue #255, opened 2018-11-29
Colorful.Console instead of static Chalk


richardwerkman:
Right now we use a static Chalk class to write in colors to the console. We also use this for testing.

Using the library [ColorfulConsole](http://colorfulconsole.com/) we could create a cleaner code style. It also has some awesome features like printing in gradient. This could be an cool addition to the ascii logo.

My question is, would this be an improvement? Or does anyone have experience with the library?

Mobrockers:
If someone really wants to add this, sure. Does not have any prio for me because who looks at the cmd output anyway? CI + json report is what is going to be used imo.

richardwerkman:
True, of course a json report has more prio but I thought this would be a fun addition.

mcliment:
I've been taking a look at this issue as it seemed kind of fun and I have some experience with Colorful.Console but the main problem is that `IChalk` is used in many tests for mocks that later check that the messages have been properly printed.

So wrapping `Colorful.Console` inside `IChalk` does not look interesting at all and capturing all things written to console with `Console.SetOut` and then checking stuff does not look very appealing either.

I ran out of ideas.

eNeRGy164:
I'm investigating this, using Crayon.

![image](https://user-images.githubusercontent.com/10671831/96516599-c3275b80-1267-11eb-893c-f268c57773c6.png)

Mobrockers:
Cool!

eNeRGy164:
Chalk has been replaced now, should this issue be closed?

Maybe create a new one for a fancier banner?

Mobrockers:
Yep! Thank you for your contributions!

PR #1266, opened 2020-10-20
Replace Chalk with Crayon

eNeRGy164:
+ Add some familiar colors to the ASCII art

Open for feedback ⬛🖍️

Implements #255


Mobrockers:
Would you say using Crayon is an improvement? :)

eNeRGy164:
If you look at it as only the replacement, I think it's a small improvement on not having custom code (Chalk.cs) with a mix of console code and a lot of linefeeds in the strings that you also should test.

I think it is a bit more future-proof. If you now want to use more/different colors, you don't have to add any code to your project for that, as it's already available.

src/Stryker.Core/Stryker.Core/Reporters/HtmlReporter/HtmlReporter.cs
```diff
@@ -14,13 +15,13 @@ public class HtmlReporter : IReporter
     {
         private readonly StrykerOptions _options;
         private readonly IFileSystem _fileSystem;
-        private readonly IChalk _chalk;
+        private readonly TextWriter _output;
```
richardwerkman, line 18: I think the name _output is a bit misleading. IMO `_consoleWriter` would be a better fit. The same goes for the other classes that use this.

src/Stryker.Core/Stryker.Core.UnitTest/StringWriterExtenstions.cs
```diff
@@ -0,0 +1,82 @@
...
+        public static string RemoveAnsi(this TextWriter stringWriter)
+        {
+            return Ansi.Replace(stringWriter.ToString(), string.Empty);
+        }
+
+        /// <summary>
+        /// Counts the number of green spans in the <paramref name="stringWriter"/>.
+        /// </summary>
+        /// <returns>The number of green spans.</returns>
+        public static int GreenSpanCount(this TextWriter stringWriter)
```
richardwerkman, line 27: Very nice! Looks much better in our tests :)

richardwerkman (commented):
Awesome! I think this is indeed an improvement over Chalk.cs

I do have some doubts about the new red "logo" on startup. It could be a bit too much red in your face haha. I think a gradient colour would be nicer.

eNeRGy164:
> I do have some doubts about the new red "logo" on startup. It could be a bit too much red in your face haha. I think a gradient colour would be nicer.

Ever visited https://stryker-mutator.io/ ? 😛

src/Stryker.Core/Stryker.Core/Reporters/HtmlReporter/HtmlReporter.cs
eNeRGy164: I thought to keep it a little bit more abstract, as you're not always writing to a console (f.e. in the unit tests).
But I will change this

src/Stryker.Core/Stryker.Core/Reporters/DashboardReporter.cs
```diff
@@ -16,14 +17,14 @@ public partial class DashboardReporter : IReporter
...
 
-        public DashboardReporter(StrykerOptions options, IDashboardClient dashboardClient = null, ILogger<DashboardReporter> logger = null, IChalk chalk = null)
+        public DashboardReporter(StrykerOptions options, IDashboardClient dashboardClient = null, ILogger<DashboardReporter> logger = null, TextWriter output = null)
         {
             _options = options;
             _dashboardClient = dashboardClient ?? new DashboardClient(options);
             _logger = logger ?? ApplicationLogging.LoggerFactory.CreateLogger<DashboardReporter>();
-            _chalk = chalk ?? new Chalk();
+            _output = output ?? Console.Out;
             _logger = ApplicationLogging.LoggerFactory.CreateLogger<DashboardReporter>();
```
eNeRGy164, line 28: I will remove the double initialization of _logger here, If you would pass it it along now, it will always get overwritten by the Factory instance.

src/Stryker.Core/Stryker.Core.UnitTest/StringWriterExtenstions.cs
eNeRGy164: Glad you like it :)

richardwerkman:
Well it would be awesome if the colours went from left to right and start with red, go to yellow and turn to red again for example. But it would be a bit tricky with the current library. Maybe a gradient from top to bottom would be more easy.

I think a black background would look better

eNeRGy164:
I would suggest to create a separate ticket to make the banner fancier.
I'll remove the background color from this commit.

Issue #1154, opened 2020-07-27
Add absolute path to json report


p3pijn:
**Is your feature request related to a problem? Please describe.**
The path from where you execute `dotnet stryker` is also the location where Stryker will output its reports.
The json report currently only contains the relative path of the file under test: https://github.com/stryker-mutator/stryker-net/blob/fd278b1f783ea5948d90ea15f5af276eb316ca2d/src/Stryker.Core/Stryker.Core/Reporters/JsonReporter/JsonReport.cs#L98

But depending on from where you have executed `dotnet stryker` the json report is either persisted to the folder containing the test project or to the folder containing the project under test.
Because only an absolute path to the mutated file is available in the json reports, and the location where the `mutation-report.json` might vary, it is currently hard to reliably find the file on disk based on the relative path in the json report.

**Describe the solution you'd like**
Add metadata to the json report so that it is possible to determine the absolute path of a mutated file that was reported to the JSON report, similar to how other coverage and mutation formats do this (e.g. pitest, opencover, jacoco, cobertura, etc)
One light-weight option might be that the `mutation-report.json` contains a field with the absolute path of the project folder (or file) that was tested. And the reported relative paths are then relative to that folder.

Example json mutation report:
```
{
  "schemaVersion": "1.3",
  "projectRoot": "/some/location"
  "files": {
    "SomeClass.cs": {
		...
    }
  }
}
```
By using such a structure, automated tools can more easily and reliably determine the absolute location of `SomeClass.cs`.

**Additional context**
My context is that I'm building tools that automatically collect and correlate code & test metrics.

p3pijn:
It looks like the sonar-stryker plugin is also making a best effort guess based on a wildcard to find the actual file that was tested:
https://github.com/stryker-mutator/sonar-stryker-plugin/blob/b93e835498c60a0413d6e7f62c414216248ec89b/src/main/java/io/github/strykermutator/StrykerSensor.java#L114
This plugin could possibly also benefit from this solution.

Mobrockers:
> It looks like the sonar-stryker plugin is also making a best effort guess based on a wildcard to find the actual file that was tested:
> https://github.com/stryker-mutator/sonar-stryker-plugin/blob/b93e835498c60a0413d6e7f62c414216248ec89b/src/main/java/io/github/strykermutator/StrykerSensor.java#L114
> This plugin could possibly also benefit from this solution.

[...]

PR #1267, opened 2020-10-22
feat(json report): Add absolute path to json report #1154

mcliment:
In the scope of issue #1154 I added a way to extract the project root from the first `FileLeaf` as I could not get it from anywhere else in the `JsonReport`.

I'll add as well a PR to https://github.com/stryker-mutator/mutation-testing-elements schema if you agree on this PR.


Mobrockers:
Hi!

Every FolderComposite should have it's FullPath property set. Why could you not use this? Is something wrong with the way we set the FullPath?

mcliment:
Hi @Mobrockers, thanks for the quick response.

I see that actually `FolderComposite` also has a relative path to project set so I could use that instead of the `FileLeaf`, so probably is a safe bet. I didn't saw it in the beginning. Thanks for the tip.

Mobrockers:
> Hi @Mobrockers, thanks for the quick response.
> 
> I see that actually `FolderComposite` also has a relative path to project set so I could use that instead of the `FileLeaf`, so probably is a safe bet. I didn't saw it in the beginning. Thanks for the tip.

You can use the FullPath instead of using the RelativePath to build the fullpath, but yes.

mcliment:
I see a problem here:

imatge

This is supposed to be the root folder of the project and the first file contained there is:

imatge

As you can see the `FullPath` of the first folder is not right, in this case should be `C:\Projects\ToyProject\ToyProject\ToyProjectLib1` instead of just `C:\Projects\ToyProject\ToyProject`.

The .csproj of the project is actually in `C:\Projects\ToyProject\ToyProject\ToyProjectLib1`.

Mobrockers:
Right, I forgot about that. I think it's valid to just always take the first FolderComposite in the IReadonlyInputComponent. @richardwerkman ?

richardwerkman:
We add a FolderComposite on top of all files and folders to wrap other projects in. When run on a single project it looks a bit odd indeed. So when run on a single project you could always take to first FolderComposite from the root and that will be the folder where your .csproj is located.

mcliment:
According to your comments and what I understand `RelativePath` and `RelativePathToProjectFile` should be, I made some changes to `InputFileResolver` to match those. Actually, subfolders were lost due to using `sub` instead of `folder` when building the `FullPath` property.

What do you think?

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -143,8 +143,8 @@ private FolderComposite FindProjectFilesUsingBuildalyzer(ProjectAnalyzerResult a
             var rootFolderComposite = new FolderComposite()
             {
                 Name = string.Empty,
-                FullPath = projectRoot,
-                RelativePath = string.Empty,
+                FullPath = projectUnderTestDir,
```
Mobrockers, line 146: We need a top level 'container' FolderComposite in case there are multiple projects, so we can't just take the projectUnderTestDir as the root folder. It needs to be one folder level higher.

mcliment: I did some changes, I think it does more sense now. 

But then, I don't see the meaning in calling the property `ProjectRoot` as it's a bit misleading as paths are relative to the parent that holds all the projects in the case you have many.

Mobrockers: You are correct, it's probably a bad name. How about solutionRoot?

mcliment: Done!

Mobrockers: Hi sorry, I thought you were talking about the naming of the projectRoot variable in InputFileResolver.cs.


In the context of the mutation report schema I don't think it makes sense to call it SolutionRoot instead of ProjectRoot as dotnet has solutions and solution files but other languages and frameworks do not have this concept. The mutation report schema is generic, and in other places where we deal with the schema we call the 'combination' of 'modules' a project. 

Perhaps we should have the discussion about what we call it in mutation-testing-metrics instead of here since we are not the only consumers of the json schema.

Mobrockers:
Looks good to me. When it's been added to the schema and a new schema version has been released we can merge this into stryker.net as well.

mcliment:
Hi @Mobrockers, let's recap a bit on the original goal of the PR and the current status, so I can find out a simple yet effective solution to the current problem.

I have a Toy project with this structure:
```
ToyProjectLib1
\- Folder1
   \- Folder2
      \- Multiplier.cs
\- Adder.cs
```

Using the `master` branch code and setting the working directory on the test projects of my Toy project (that tests `Adder.cs` as well as `Multiplier.cs` I get this structure in the `mutationReport` parameter of the `JsonReport` constructor:

```
Type: FolderComposite
FullPath: null,
Name: null,
RelativePath: null,
RelativePathToProjectFile: null
Children: 3
\- Type: FolderComposite
   FullPath: "C:\Projects\ToyProject\ToyProject",
   Name: "",
   RelativePath: "",
   RelativePathToProjectFile: "."
   Children: 1
   \- Type: FileLeaf
      FullPath: "C:\Projects\ToyProject\ToyProject\ToyProjectLib1\Adder.cs",
      Name: "Adder.cs",
      RelativePath: "Adder.cs",
      RelativePathToProjectFile: "Adder.cs"
\- Type: FolderComposite
   FullPath: "C:\Projects\ToyProject\ToyProject\ToyProjectLib1\Folder1",
   Name: "Folder1",
   RelativePath: "ToyProjectLib1\\Folder1",
   RelativePathToProjectFile: "Folder1"
   Children: 1
   \- Type: FolderComposite
      FullPath: "C:\Projects\ToyProject\ToyProject\ToyProjectLib1\Folder2", // Wrong!!!
      Name: "Folder2",
      RelativePath: "ToyProjectLib1\Folder2",
      RelativePathToProjectFile: "Folder2"
      Children: 1
      \- Type: FileLeaf
         FullPath: "C:\Projects\ToyProject\ToyProject\ToyProjectLib1\Folder1\Folder2\Multiplier.cs",
         Name: "Multiplier.cs",
         RelativePath: "ToyProjectLib1\Folder2\Multiplier.cs", // Wrong!!!
         RelativePathToProjectFile: "Folder1\Folder2\Multiplier.cs"
\- Type: FolderComposite
   FullPath: "C:\Projects\ToyProject\ToyProject\ToyProjectLib1\obj",
   Name: "obj",
   RelativePath: "ToyProjectLib1\obj",
   RelativePathToProjectFile: "obj"
   Children: ...
```

And the computed `Files` for the report are like this:

```
[
    { "Key": "Adder.cs", Value: object },
    { "Key": "ToyProjectLib1\Folder2\Multiplier.cs", Value: object }
]
```

With my changes to the `InputFileResolver` included in this PR, the results are like this:

```
Type: FolderComposite
FullPath: null,
Name: null,
RelativePath: null,
RelativePathToProjectFile: null
Children: 2
\- Type: FolderComposite
   FullPath: "C:\Projects\ToyProject\ToyProject",
   Name: "",
   RelativePath: "",
   RelativePathToProjectFile: null
   Children: 0
\- Type: FolderComposite
   FullPath: "C:\Projects\ToyProject\ToyProject\ToyProjectLib1",
   Name: "ToyProjectLib1",
   RelativePath: "ToyProjectLib1",
   RelativePathToProjectFile: "."
   Children: 3
   \- Type: FileLeaf
      FullPath: "C:\Projects\ToyProject\ToyProject\ToyProjectLib1\Adder.cs",
      Name: "Adder.cs",
      RelativePath: "ToyProjectLib1\Adder.cs",
      RelativePathToProjectFile: "Adder.cs"
   \- Type: FolderComposite
      FullPath: "C:\Projects\ToyProject\ToyProject\ToyProjectLib1\Folder1",
      Name: "Folder1",
      RelativePath: "ToyProjectLib1\Folder1",
      RelativePathToProjectFile: "Folder1"
      Children: 1
      \- Type: FolderComposite
         FullPath: "C:\Projects\ToyProject\ToyProject\ToyProjectLib1\Folder1\Folder2",
         Name: "Folder2",
         RelativePath: "ToyProjectLib1\Folder1\Folder2",
         RelativePathToProjectFile: "Folder1\Folder2"
         Children: 1
         \- Type: FileLeaf
            FullPath: "C:\Projects\ToyProject\ToyProject\ToyProjectLib1\Folder1\Folder2\Multiplier.cs",
            Name: "Multiplier.cs",
            RelativePath: "ToyProjectLib1\Folder1\Folder2\Multiplier.cs",
            RelativePathToProjectFile: "Folder1\Folder2\Multiplier.cs"
   \- Type: FolderComposite
      FullPath: "C:\Projects\ToyProject\ToyProject\ToyProjectLib1\obj",
      Name: "obj",
      RelativePath: "ToyProjectLib1\obj",
      RelativePathToProjectFile: "obj"
      Children: ...
```

And the computed `Files` for the report are like this:

```
[
    { "Key": "ToyProjectLib1\Adder.cs", Value: object },
    { "Key": "ToyProjectLib1\Folder1\Folder2\Multiplier.cs", Value: object }
]
```

So in the end, the value we need to add to the Json report `ProjectRoot` property should be `C:\Projects\ToyProject\ToyProject\ToyProjectLib1` in my understanding should be the `FullPath` of the `FolderComposite` from root that has `RelativeToProjectFile == "."' . Maybe I'm wrong.

As well, I don't really understand the meaning of the first `FolderComposite` that points to `C:\Projects\ToyProject\ToyProject` and has 0 children.

Sorry for the long explanation but I think it's clearer this way to explain the problems I have with the original composite structure.

mcliment:
I changed a bit the `GenerateFileReportComponents` in the end so the path to the file is the newly introduced `ProjectRoot` + the key of the Files collection. I think makes sense that way.

As well, I added a PR in `mutation-testing-elements`: https://github.com/stryker-mutator/mutation-testing-elements/pull/698

Thanks for your help.

src/Stryker.Core/Stryker.Core/Reporters/JsonReporter/JsonReport.cs
```diff
@@ -95,12 +98,21 @@ public string ToJsonHtmlSafe()
 
         private IDictionary<string, JsonReportFileComponent> GenerateFileReportComponents(ReadOnlyFileLeaf fileComponent)
         {
-            return new Dictionary<string, JsonReportFileComponent> { { fileComponent.RelativePath, new JsonReportFileComponent(fileComponent) } };
+            return new Dictionary<string, JsonReportFileComponent> { { fileComponent.RelativePathToProjectFile, new JsonReportFileComponent(fileComponent) } };
```
Mobrockers, line 101: Thank you for that! Nice bug fix :)

src/Stryker.Core/Stryker.Core/Initialisation/CsharpProjectComponentsBuilder.cs
```diff
@@ -105,7 +105,7 @@ private FolderComposite FindProjectFilesUsingBuildalyzer(ProjectAnalyzerResult a
                     continue;
                 }
 
-                var relativePath = Path.GetRelativePath(projectUnderTestDir, sourceFile);
+                var relativePath = Path.GetRelativePath(projectRoot, sourceFile);
```
Mobrockers, line 108: I don't think this is correct. the projectRoot is the 'solution folder'. Project files are not relative to the solution folder, the are relative to the project folder. Which is the projectUnderTestDir.

Mobrockers (changes requested):

src/Stryker.Core/Stryker.Core/Initialisation/CsharpProjectComponentsBuilder.cs
Mobrockers: I will take a look at our project structure to see if I can make some sense of it. It is perhaps too deep in stryker for a contributor to dig in to with the amount of available documentation (none 😅)

mcliment: I see things are a bit broken with some inconsistencies here and there because to my understanding, `RelativePath` means relative to root folder and `RelativePathToProjectFile` means relative to project folder, so they can't be the same unless the project is on the root folder.

For example, in line 116, this is how `RelativePath` is set: `RelativePath = _fileSystem.Path.Combine(folderComposite.RelativePath, fileName),` which leads to a wrong path.

Maybe I don't understand exactly what we're trying to achieve here.

Mobrockers: A contributor added the RelativePathToProjectFile at some point and we failed to question why he did that. At this point I'm not entirely sure what we're trying to achieve here either so I'm gonna have to look it over with a colleague and decide on how we want the design to be. Then fix up the whole thing and document what's what and why so we don't mess it up again. It might take us some time to get to that. Apologies.

Mobrockers:
TODO:

Fix cleartext-tree reporter after project components refactor
Check that project root still works

richardwerkman:
@Mobrockers I think I've cleared all issues with this PR. Could you check it again?

src/Stryker.Core/Stryker.Core/StrykerRunner.cs
```diff
@@ -134,5 +134,28 @@ private void AnalyseCoverage(StrykerOptions options)
...
             }
         }
+
+        /// <summary>
+        /// In the case of multiple projects we wrap them inside a wrapper root component. Otherwise the only project root will be the root component.
+        /// </summary>
+        /// <param name="projectComponents">A list of all project root components</param>
+        /// <param name="options">The current stryker options</param>
+        /// <returns>The root folder component</returns>
+        private CsharpFolderComposite AddRootFolderIfMultiProject(IEnumerable<IProjectComponent> projectComponents, StrykerOptions options)
```
Mobrockers, line 144: This should return an IFolderComposite or IProjectComponent

src/Stryker.Core/Stryker.Core/StrykerRunner.cs
```diff
@@ -134,5 +134,28 @@ private void AnalyseCoverage(StrykerOptions options)
...
+        {
+            var rootComponent = new CsharpFolderComposite
```
Mobrockers, line 146: It would be better to have a new type for this (RootFolderComposite?) because CsharpFolderComposite is specific to csharp, which is not needed at this stage in the process.

richardwerkman: Why do we even have a CsharpFolder? A folder should not care what type of project it is in right? It can just contain files. The files are indeed language specific.

Mobrockers: Because we use the syntaxtrees on folder level. For example, when compiling we take all syntaxtrees from the root component. The syntaxtree type is csharp specific, so the folder must be as well.

Mobrockers: We also use the folder composite to add compilation only syntax trees to the project (for example the stryker helpers)

Mobrockers:
@mcliment Thank you for starting the work on this PR and for showing us all the mess we had to clean up 😅
Sorry it took so long

mcliment:
@Mobrockers @richardwerkman Thanks to you guys for maintaining this project!

PR #1271, opened 2020-10-23
docs: update docs for new website

hugo-vrijswijk:
🙋‍♀️! This PR does a couple things with documentation:

- Fix some links to be relative instead of from the root
- Removes the table of contents. I can add them back if you want, but Docusaurus already has TOC's in the side
- Add a title meta-header to the `Stryker-in-pipeline.md`. The other files have nice capitalized names so they don't need it.
- Add custom-edit-urls to fix 'Edit this page' links

image

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -1,3 +1,7 @@
+---
+custom_edit_url: https://github.com/stryker-mutator/stryker-net/edit/master/docs/Configuration.md
+---
+
 For .NET Core projects Stryker.NET can be run without any configuration. On .NET Framework projects the solution path is required.
 
 ## Use a config file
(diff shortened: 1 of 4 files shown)


Mobrockers:
LGTM 🐝

We have some issues with TOC links not working correctly due to longer titel names and guessing you need to do to account for that in markdown TOC's. Does docusaurus handle the links correctly?

hugo-vrijswijk:
> We have some issues with TOC links not working correctly due to longer titel names and guessing you need to do to account for that in markdown TOC's. Does docusaurus handle the links correctly?

The longest one I could find was Configuration#using-dashboard-compare-in-a-pull-request-pipeline and that worked fine

PR #1275, opened 2020-10-24
feat: Move FilterMutations status logging to reporter so the logging can be turned off

### LINKED ISSUE #1261: Move reporting on before-testing mutant statuses to reporter
We have some reporting as log statements here: https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs#L288-L315

This doesn't really fit as log statements, perhaps it should be in a reporter like the clear text reporter.

We should probably implement this reporter hook to build this reporting: https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/Reporters/IReporter.cs#L12

luciaconde:
Fix #1261, by moving the log statements in FilterMutations() (regarding mutants status) to a custom default reporter named _StatusReporter_.


Mobrockers:
Could you add an example of what the output looks like?

luciaconde:
Sure! In the end I used the logger instead of writing directly to the console, so the output is identical to the one before (with the same coloring and formatting). As an example output from a Stryker run on a project:

`[12:44:52 INF] Using 7619 ms as testrun timeout`
`[12:44:55 INF] Capture mutant coverage using 'perTest' mode.`
`[12:44:56 INF] 10    mutants got status NoCoverage.   Reason: Mutant has no test coverage`
`[12:44:56 INF] 10    total mutants are skipped for the above mentioned reasons`
`[12:44:56 INF] 86    total mutants will be tested`
`[12:44:56 INF] Coverage analysis will reduce run time by discarding 77,4% of tests because they would not change results.`

image

But I can also work on a different formatting if that's better.

On a different note, as there weren't explicit tests on `MutationTestProcessTests.cs` for the log statements, I was also thinking about working on unit tests for the `StatusReporter `similarly to the other reporters.

src/Stryker.Core/Stryker.Core/Reporters/StatusReporter.cs
```diff
@@ -0,0 +1,96 @@
...
+        private readonly StrykerOptions _options;
+        private readonly ILogger<StatusReporter> _logger;
+        private readonly TextWriter _consoleWriter;
+
+
+        public StatusReporter(StrykerOptions strykerOptions, ILogger<StatusReporter> logger = null, TextWriter consoleWriter = null)
+        {
+            _options = strykerOptions;
+            _logger = logger ?? ApplicationLogging.LoggerFactory.CreateLogger<StatusReporter>();
+            _consoleWriter = consoleWriter ?? Console.Out;
```
richardwerkman, line 25: You are not using the console writer right? So could you remove it from the constructor?

src/Stryker.Core/Stryker.Core/Reporters/StatusReporter.cs
```diff
@@ -0,0 +1,96 @@
...
+    {
+        private readonly StrykerOptions _options;
+        private readonly ILogger<StatusReporter> _logger;
+        private readonly TextWriter _consoleWriter;
+
+
+        public StatusReporter(StrykerOptions strykerOptions, ILogger<StatusReporter> logger = null, TextWriter consoleWriter = null)
+        {
+            _options = strykerOptions;
+            _logger = logger ?? ApplicationLogging.LoggerFactory.CreateLogger<StatusReporter>();
```
richardwerkman, line 24: Could you add tests with a mocked logger?

richardwerkman (changes requested):
Looking a lot better in a seperate reporter! Thanks :)

src/Stryker.Core/Stryker.Core/Reporters/StatusReporter.cs
luciaconde: Oh right, forgot to remove it, thanks!

src/Stryker.Core/Stryker.Core/Reporters/StatusReporter.cs
luciaconde: On it!

Mobrockers:
Looks good! Let us know when you're ready for final review.

src/Stryker.Core/Stryker.Core/Reporters/StatusReporter.cs
richardwerkman: Are you still on it? If you need help let us know 👍

richardwerkman (approved):

richardwerkman:
Thanks for your help @luciaconde

Issue #1277, opened 2020-10-24
Organize CLI help text in categories


eNeRGy164:
The current set of options for Stryker.NET is huge, resulting in a very large help text and it's hard to find settings you might be interested in.

I know #578 is about modifying the way of using options, and I have looked at it but this is a very large task and I didn't dare to start on this as most CLI libraries for .NET are all a little bit different as Stryker is doing things at the moment.
In the meantime, users are not helped. Therefore, I suggest to at least improve the readability of the help text.

I looked at the library currently used for options (`McMaster.Extensions.CommandLineUtils`) and it's fairly easy to overwrite the `HelpTextGenerator`

Locally I have added a category to each option (with my limited understanding of these options) and I think this improves the readability quite a bit
I also added a white line between options. Sometimes it's hard to recognize the text is already related to the next option.

An additional possibility is to add a weight (integer) to sort options within a category.

Example:
_(Reduced some options for limit the length of this post)_
```
❯ D:\github\stryker-net\src\Stryker.CLI\Stryker.CLI\bin\Debug\netcoreapp3.1\Stryker.CLI.exe --help
Stryker: Stryker mutator for .Net

Stryker mutator for .Net

Usage: Stryker [options]

Options:
  -cp|--config-file-path <path>                             Sets the config-file-path relative to current workingDirectory | stryker-config.json (default)

  -r|--reporters <reporters>                                Sets the reporter | Options[ (default)[ Progress, Html ], All, Dots, ClearText, ClearTextTree, Json, Dashboard, Baseline, (deprecated)
                                                            ConsoleProgressBar, (deprecated) ConsoleProgressDots, (deprecated) ConsoleReport ]]
                                                            This argument takes a json array as a value. Example: ['Progress', 'Html']

  -l|--log-level <logLevel>                                 Sets the console output logging level | Options [error, warning, info (default), debug, trace]

  -f|--log-file                                             Makes the logger write to a file (Logging to file always uses loglevel trace)


Build options:
  -p|--project-file <projectFileName>                       Used for matching the project references when finding the project to mutate. Example: "ExampleProject.csproj"

  -fte|--files-to-exclude <files-to-exclude>                (deprecated:Use '--mutate' instead.)Set files to exclude for mutation. Example:
                                                            ['C:/ExampleProject/Example.cs','C:/ExampleProject/Example2.cs']

  -m|--mutate <file-patterns>                               Allows to specify file that should in- or excluded for the mutations.
[...]

PR #1280, opened 2020-10-24
feat(CLI Options): Add categories

eNeRGy164:
Draft for #1277


rouke-broersma:
Only took us less than a year! I incorporated this idea in #1699 

What do you think?

PR #1282, opened 2020-10-24
Add regex look ahead and behind mutators

### LINKED ISSUE #1278: Feature Request: Regex mutations lookahead and lookbehind
@vngrp Had the idea to switch positive and negative lookahead and lookbehind regex statements. This is a clear negation mutation inside a regex that could prove usefull.

### The new mutations
| Original | Mutated | 
| ------------- | ------------- | 
| `(?=...)` | `(?!...)` |
| `(?!...)` | `(?=...)` |
| `(?<=...)` | `(?<!...)` |
| `(?<!...)` | `(?<=...)` |

vngrp:
## Description
In regex there is a way to match something, only if it is preceded or proceded by a group.
For example: (?<=foo)bar will match the word bar in foobar, but won't match bar in minibar.

This mutator flips these regexes to check at the end instead of beginning, and to ensure
that it does NOT start with the group. It flips the check.

## Issue
Closes #1278

### DIFF
--- a/src/Stryker.RegexMutators/Stryker.RegexMutators.UnitTest/Mutators/LookAroundMutatorTest.cs
+++ b/src/Stryker.RegexMutators/Stryker.RegexMutators.UnitTest/Mutators/LookAroundMutatorTest.cs
@@ -1,0 +1,89 @@
+﻿using RegexParser.Nodes;
+using Shouldly;
+using Stryker.RegexMutators.Mutators;
+using System.Collections.Generic;
+using System.Linq;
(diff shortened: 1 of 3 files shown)


richardwerkman (changes requested):
I think this solution is creating too many mutations. It is not fitting the docs, while I think the docs are correct

vngrp:
Good point, thanks for the review. Updated the code and tests, should be good now

richardwerkman (approved):

PR #1283, opened 2020-10-24
Add look ahead and behind mutators to docs

vngrp:
Docs for https://github.com/stryker-mutator/stryker-net/pull/1282

### DIFF
--- a/docs/Mutators.md
+++ b/docs/Mutators.md
@@ -140,4 +140,4 @@ Stryker supports a variety of mutators, which are listed below. Do you have a su
 | `a^b` | `~(a^b)` |
 
 ## Regular Expressions
-For the full list of all available regex mutations, see the [regex mutator docs](/docs/RegexMutators.md).
+For the full list of all available regex mutations, see the [regex mutator docs](./docs/RegexMutators.md).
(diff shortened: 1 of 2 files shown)


Mobrockers:
The regex mutators list is getting very long. Perhaps we need a new doc page `RegexMutators` with categories similar to the regular Mutators to be able to differentiate between different RegexMutators.

docs/Mutators.md
```diff
@@ -164,9 +164,13 @@ Stryker supports a variety of mutators, which are listed below. Do you have a su
 | `abc{5}` | `abc` |
 | `abc{5,}` | `abc` |
 | `abc{5,8}` | `abc` |
+| `(?=abc)` | `(?!abc)` |
```
richardwerkman, line 167: The docs don't match with the implementation. If I'm correct `(?=foo)` mutates to `(?!foo)` and `(?<=foo)`.

I think the docs are correct but the implementation should change. It should only flip the positivity of the node and not the way it looks (ahead/behind).

vngrp: Yes good point, it was making too many mutants. Changed the code to only flip the negative/positive and leave look ahead and behind.

vngrp:
@Mobrockers Good one, thanks. It's already getting quite long and it's only just the beginning. Moved it to a separate RegexMutators file. Used same categories as regex101.com

PR #1286, opened 2020-10-26
fix(ProgressReport):  Replace progress reporting

### LINKED ISSUE #519: Progress bar appending instead of replacing values inline (Macos 10.14.5)

eNeRGy164:
Implement ShellProgressBar and make MutantsResultReporter obsolete

This is a WIP, to first get some feedback.
Added Unit testing.

Closes #519


src/Stryker.Core/Stryker.Core/Reporters/Progress/StopWatchProvider.cs
```diff
@@ -5,17 +5,22 @@ namespace Stryker.Core.Reporters.Progress
...
         private Stopwatch _watch;
 
         public void Start()
         {
-            _watch = new Stopwatch();
-           _watch.Start();
+            _watch = Stopwatch.StartNew();
+        }
+
+        public void Stop()
```
Mobrockers, line 21: 🙈 nice

Mobrockers:
Looks good to me 🐝

src/Stryker.Core/Stryker.Core/Reporters/Progress/StopWatchProvider.cs
eNeRGy164: I thought so :P

Mobrockers:
Mhm after testing it looks like the fancy progress bar does not work correctly in the visual studio debug console

image

In git bash it fails completely

image

It does work in CMD:

image


Git bash and the visual studio debug console are probably not the main shells used to run stryker but it suggests to me that the used library does not support all shells so a bit more error handling could not hurt.

Especially the issue in git bash is concerning because that is the same issue we noticed on macOSX before and is why we disabled replacing values when Console.CursorTop caused an exception:

https://github.com/stryker-mutator/stryker-net/blob/b05a5f907f9db4ff0f0b4502a15bc675cb3a993f/src/Stryker.Core/Stryker.Core/Reporters/Progress/ConsoleOneLineLogger.cs#L62-L74

eNeRGy164:
https://github.com/Mpdreamz/shellprogressbar/issues/22
Some of these seem to be there for a while.

Mobrockers:
Does anyone know someone with a mac that could test this pr? 😅

richardwerkman:
@vngrp has a mac, maybe he could take a look?

eNeRGy164:
I have no environment to recreate these issues.

Works for me:
* Windows PowerShell 5.1
* PowerShell Core 7
* Windows cmd
* PowerShell terminal in VSCode
  image
* Visual Studio Debug Conole
  image

eNeRGy164:
> _Mhm after testing it looks like the fancy progress bar does not work correctly in the visual studio debug console_
> 
> This happens only when debug logging is turned on

Ah missed that remark.
Yes, you can't have multiple processes doing console output at the same time if one is messing with the positions, like in this case the progressbar. That is also why I had to merge the mutant reporting and the progressbar in one output.

Mobrockers:
> > _Mhm after testing it looks like the fancy progress bar does not work correctly in the visual studio debug console_
> > This happens only when debug logging is turned on
> 
> Ah missed that remark.
> Yes, you can't have multiple processes doing console output at the same time if one is messing with the positions, like in this case the progressbar. That is also why I had to merge the mutant reporting and the progressbar in one output.

Unfortunately that seems like a regression compared to the current implementation (on some terminals).
We're going to do more testing to decide what to do with this pr, as it stands it looks like we're gonna have to decide between regressions on the one hand or other issues on the other hand.

Sorry for the wait.

eNeRGy164:
No problem, it's your project after all.
I just can't reproduce the errors, so it's hard to fix them myself.

eNeRGy164:
For showing addition logging we could reroute to the WriteLine function mentioned here: https://github.com/Mpdreamz/shellprogressbar/issues/31#issuecomment-523022264

PR #1294, opened 2020-10-31
chore: Make Project Components language specific

dukedagmor:
is now exactly the same as the fsharp branch except the fileleaffsharp and foldercompositefsharp, updated docs and UML


dukedagmor:
ToReadOnly can maybe be played in the interface, this would ensure the possibility to convert to readonly for each implementation of said interface and remove duplicates from the UML

Mobrockers:
I prefer FsharpFileLeaf over FileLeafFsharp and if we're adding more IFileLeaf/IFolderComposite implemntations for specific languages I prefer that the difference is clear from the name for all implementations. Which means renaming FolderComposite/FileLeaf to CsharpFolderComposite/FileLeaf. Same goes for any other classes that are going to become specific to one language. That way you won't have to guess what it's for.

Mobrockers:
> ToReadOnly can maybe be played in the interface, this would ensure the possibility to convert to readonly for each implementation of said interface and remove duplicates from the UML

Are you intending to add that to this PR or is it ready to merge?

dukedagmor:
good suggestions! I will update this branch to do so

dukedagmor:
implemented the requested changes and added the toreadonly method to the interfaces

src/Stryker.Core/Stryker.Core/ProjectComponents/ReadOnlyFolderComposite.cs
```diff
@@ -1,16 +1,16 @@
...
     public class ReadOnlyFolderComposite : ReadOnlyProjectComponent
     {
 
-        private readonly FolderComposite _folderComposite;
+        private readonly IParentComponent _folderComposite;
         private readonly bool _belongsToProject;
 
         public IEnumerable<IReadOnlyProjectComponent> Children => _folderComposite.Children.Select(child => child.ToReadOnlyInputComponent());
-        public ReadOnlyFolderComposite(FolderComposite folderComposite, bool belongsToProjectUnderTest) : base(folderComposite)
+        public ReadOnlyFolderComposite(IParentComponent folderComposite, bool belongsToProjectUnderTest) : base(folderComposite)
```
Mobrockers, line 13: This reads weird. We're receiving an IParentComponent that we know is a folderComposite somehow. And from then on we also call it a folder composite. Perhaps the IParentComponent can just be IFolderComposite? We should think about this.

dukedagmor: done

src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/DiffMutantFilterTests.cs
```diff
@@ -218,7 +220,7 @@ public void ShouldNotFilterMutantsWhereCoveringTestsContainsChangedTestFile()
         {
             // Arrange
             string testProjectPath = "C:/MyTests";
-            var options = new StrykerOptions();
+            var options = new StrykerOptions(diff: false);
```
Mobrockers, line 223: Setting diff to false here makes no sense as this test is about testing the diff feature

dukedagmor: the other tests in the class have the same option

dukedagmor: https://github.com/stryker-mutator/stryker-net/blob/bbbc78eeb98d2f19a763afbb7d03054dffa8b6cc/src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/DiffMutantFilterTests.cs#L96

dukedagmor:
merge conflicts solved

src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/DiffMutantFilterTests.cs
Mobrockers: Mhm it seems some tests have it set to false and some don't. Interesting. In any case the original did not have diff set to false, so why are you changing it?

PR #1324, opened 2020-11-13
feat: add expression to body conversion for properties

dupdob:
as per Slack discussion, this PR mainly adds support for properties using arrow expressions for accessor(s) by converting them to body form when necessary. Also support rollbacking those conversions in case of build errors.
Added several tests to complete coverage and fix some bugs.


src/Stryker.Core/Stryker.Core/Instrumentation/BaseEngine.cs
```diff
@@ -1,11 +1,10 @@
...
 using Microsoft.CodeAnalysis;
 
 namespace Stryker.Core.Instrumentation
 {
     internal abstract class BaseEngine<T>: IInstrumentCode where T: SyntaxNode
     {
-
         protected BaseEngine(string markerId, string engineId)
         {
             InstrumentEngineID = engineId;
```
richardwerkman, line 10: I'd like to see we take the engineId from `GetType().Name`. That way all types get their own Id but we don't have to pass a string from each type to the base constructor. As we only create all engines once performance should not be impacted by this right?

src/Stryker.Core/Stryker.Core/Instrumentation/PropertyExpressionToBodyEngine.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Stryker.Core.Helpers;
+
+namespace Stryker.Core.Instrumentation
+{
+    internal class PropertyExpressionToBodyEngine : BaseEngine<PropertyDeclarationSyntax>
+    {
+        public PropertyExpressionToBodyEngine(string markerId) : base(markerId, "PropertyExpressionToBodyEngine")
```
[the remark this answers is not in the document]
richardwerkman: If we don't use the `GetType().Name` solution in the base type I'd like to see a usage of `nameof()` here and in the other engine types

src/Stryker.Core/Stryker.Core/Mutants/NodeOrchestrators/PropertyDeclarationOrchestrator.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            result = mutated.ReplaceNode(getter.Body!, SyntaxFactory.Block(
+                    MutantPlacer.PlaceStatementControlledMutations(
+                        getter.Body,
+                        context.StatementLevelControlledMutations.Union(context.BlockLevelControlledMutations).Select(
+                            m => (m.Id,
+                                SyntaxFactory.ReturnStatement(
+                                    sourceNode.ExpressionBody!.Expression
+                                        .InjectMutation(m.Mutation)) as StatementSyntax))))).
+                WithSemicolonToken(SyntaxFactory.MissingToken(SyntaxKind.SemicolonToken));
```
richardwerkman, line 41:
```suggestion
                                        .InjectMutation(m.Mutation)) as StatementSyntax)))))
                .WithSemicolonToken(SyntaxFactory.MissingToken(SyntaxKind.SemicolonToken));
```

src/Stryker.Core/Stryker.Core.UnitTest/Mutants/MutantOrchestratorTests.cs
```diff
@@ -840,13 +835,55 @@ public void ShouldMutateStaticProperties()
...
+{
+    get { return (StrykerNamespace.MutantControl.IsActive(0)?"""":""TestDescription"");}
+    set { value = (StrykerNamespace.MutantControl.IsActive(1)?"""":""TestDescription"");}
+}
+static TestClass(){using(new StrykerNamespace.MutantContext()){}}}";
+            ShouldMutateSourceToExpected(source, expected);
+        }
+
+        [Fact]
+        public void ShouldMutatePropertiesAsArraowExpression()
```
richardwerkman, line 877:
```suggestion
        public void ShouldMutatePropertiesAsArrowExpression()
```

richardwerkman (commented):
The code looks great! And I've tested it and it works as expected. But I can't seem to reproduce your issue on our master. Was the issue related to coverage? Could you explain what the exact issue is this PR fixes? Or could you add an example in our integration test with the validation that it actually works.

src/Stryker.Core/Stryker.Core/Instrumentation/PropertyExpressionToBodyEngine.cs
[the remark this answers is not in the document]
dupdob: yes, no discussion there. just did not though about it 😄

src/Stryker.Core/Stryker.Core/Instrumentation/BaseEngine.cs
dupdob: Good design.
done

dupdob:
There was no actual issue. At least not on stryker side. This is was a classic case of static cache not properly tested.
But I spotted work on properties was incomplete, and voila !

richardwerkman:
Thanks for the quick adjustments. Integration test is failing, if it succeeds I will merge this

dupdob:
fair, but I have no idea why it is failing. can you check the pipeline ?
the error seems to be about an invalid MsBuild version, and I failed to see what could trigger that!

https://dev.azure.com/stryker-mutator/Stryker/_build/results?buildId=6136&view=logs&j=99031f74-bf3d-58da-997c-4e095f08dc64&t=b1b7b885-dba8-5b22-0769-49e21e77890d&l=10

Mobrockers:
Hi @dupdob ! Happy new year :) 

I think with your last merge you accidentally merged in a different feature into this branch. See commit [6d79982](https://github.com/stryker-mutator/stryker-net/pull/1324/commits/6d799828262ce407df323b2311b22d70f93f05c4)

dupdob:
Thanks, fixed

Mobrockers:
I think we will have time to finally review this on friday 🙌🙌

Mobrockers:
Also looks like the test was fixed 🙌🙌🙌😅

PR #1362, opened 2020-12-26
feat(initial-build): Allow solutionpath for dotnet core

### LINKED ISSUE #1309: Support $(SolutionDir) for dotnet build
Projects may reference shared artifacts (e.g., editorconfig, or ruleset, or snk).  Such artifacts can be located from within the .csproj via the Solution root (i.e., $(SolutionDir).  e.g., <AssemblyOriginatorKeyFile>$(SolutionDir)\StrongNameKey.snk</AssemblyOriginatorKeyFile> 

In Stryker, Projects that are not Full-framework build a specific .csproj; in such scenarios, $(SolutionDir) is undefined, so the artifacts are not found, and the build fails. 

If --solution_path is specified, build the solution rather than a specific project, such that $(SolutionDir) is automatically defined.

Pass command-line parameters through to dotnet build (to manually specify $(SolutionDir)

**Additional context**
This request should not adversely affect this enhancement: https://github.com/stryker-mutator/stryker-net/pull/1039

Bforslund:
Fix #1309

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialBuildProcessTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialBuildProcessTests.cs
@@ -34,9 +34,11 @@ namespace Stryker.Core.UnitTest.Initialisation
             target.InitialBuild(false, "/", "/");
         }
 
-        [Fact]
+        [SkippableFact]
         public void InitialBuildProcess_ShouldRunMsBuildOnDotnetFramework()
         {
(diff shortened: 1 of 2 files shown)


Mobrockers:
@richardwerkman can you review if there's any unexpected impact on multi-solution? I'm biased since I helped with this PR ;)

richardwerkman:
@Mobrockers I think this will indeed have impact on solution runs. Stryker will call dotnet build for every project and this might be slower when building the whole solution every time? Maybe we could move the initial build from "per project" to the whole solution when a solution path is given.

PR #1363, opened 2020-12-26
feat(initial-build): Allow custom msbuild path

### LINKED ISSUE #666: Add path param for MSBuild.exe
I'm not being able to run Stryker in a project (netframework 4.5) inside a machine that doesn't have VS2017 installed (it has VS2015). Stryker fails when it tries to build the solution. 

I ran some tests and found out that it tries to use the MSBuild.exe located at the first path defined on _MSBuildHelper's.fallbackLocations_. 

A parameter that i could tell stryker which msbuild installation it should use, this way we would even be able to use those stand-alone installations without dependending on visual studio).

**Additional context**
I'm trying to setup a step for mutation test in our CI process. I've read some issues here about passing build variables to MSBuild.exe, but im not sure if that's the case. I manually tried the MSBuild installed in "C:\Program Files (x86)\MSBuild" and i was able to build the solution.
If necessary, i can provide more info later at work (it's not a machine that i personally use for development).

Thank you!

Bforslund:
fix #666

Needs option


src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialBuildProcessTests.cs
```diff
@@ -54,5 +54,25 @@ public void InitialBuildProcess_ShouldRunMsBuildOnDotnetFramework()
                 It.IsAny<int>()),
                 Times.Once);
         }
+
+        [Fact]
+        public void InitialBuildProcess_ShoulduseCustomMsbuildIfNotNull()
+        {
+
```
Mobrockers, line 61:
```suggestion
```

richardwerkman:
Add option to CLI: `msbuild-path` with no shorthand

rouke-broersma:
Needs docs

Issue #520, opened 2019-05-18
--project-file does not accept relative paths


hannahchan:
I have the following file structure for my Application.
```
ExampleApp
├─ Sources
│  ├─ Project1
│  ├─ Project2
│  ├─ Project3
│  └─ Project4
├─ Tests
│  ├─ Project1.Tests
│  ├─ Project2.Tests
│  │  └─ stryker-config.json
│  ├─ Project3.Tests
│  └─ Project4.Tests
└─ ExampleApp.sln
```
When my current directory is *Project1.Tests* and I run `dotnet stryker --project-file="../../Sources/Project1/Project1.csproj"`, I get the error message *Project reference issue.*

Likewise I get the same error message when my current directory is *Project2.Tests* and I run `dotnet stryker` with a *stryker-config.json* containing the following;

```
{
    "stryker-config":
    {
        "project-file": "../../Sources/Project2/Project2.csproj"
    }
}
```

The error does not occur when using absolute paths. The version of `dotnet-stryker` I am using is `Version 0.11.0 (beta)`. The issue happens both in Windows 10 and macOS.

Mobrockers:
Hey Hannah,

Apologies for the confusion but we don't actually match the project file with the file system, we match the project file name with your project references in the test project. This should be enough to get Stryker running:

`--project-file="Project1.csproj"`

Also if project1 is the only project reference in your project1.tests project then we can auto detect your project-file settings and you won't have to supply it yourself.

hannahchan:
Thanks for getting back so quick. I just confirmed what you said and it works. I got confused because the error message I get back looks like this.
```
Project reference issue.

Test project contains more than one project references. Please add the --project-file=[projectname] argument to specify which project to mutate.
Choose one of the following references:
  C:\Users\Hannah.Chan\Code\Scaffold.WebApi\Scaffold\Sources\Scaffold.Application\Scaffold.Application.csproj
  C:\Users\Hannah.Chan\Code\Scaffold.WebApi\Scaffold\Sources\Scaffold.Data\Scaffold.Data.csproj

Example: --project-file=C:\Users\Hannah.Chan\Code\Scaffold.WebApi\Scaffold\Sources\Scaffold.Application\Scaffold.Application.csproj
```
So I assumed it needed a file path. Maybe absolute paths shouldn't work?

Mobrockers:
We should probably improve the detecting and logging of the project under test settings, thank you for bringing it to our attention!

[...]

PR #1366, opened 2021-01-06
feat(InputFileResolver) : Validate Project File Option Value

fbd:
ValidateProjectUnderTestNameFilter checks if the filter looks
like a path to a project rather than just a project name (with or without file extension), the exception message will instruct the user in how to properly use the --project-file option.

Unit test added to verify that the exception is thrown when a path is discovered.

This hopefully "fixes" #520 by providing a more helpful exception message.

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InputFileResolverTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InputFileResolverTests.cs
@@ -1173,5 +1173,27 @@ Please specify a test project name filter that results in one project.
             ex.Message.ShouldBe("Project reference issue.");
             ex.Details.ShouldContain("no project", Case.Insensitive);
         }
+
+        [Theory]
+        [InlineData("../../Sources/ExampleProject.csproj")]
+        [InlineData("../ExampleProject.csproj")]
+        [InlineData("../../")]
+        [InlineData("..\\..\\")]
+        public void ShouldThrowWhenUsingRelativePath(string shouldMatchNone)
+        {
+            var analyzerResult = TestHelper.SetupProjectAnalyzerResult(
+                    projectReferences: new List<string> {
+                        "../ExampleProject/ExampleProject.csproj",
+                        "../AnotherProject/AnotherProject.csproj"
+                    }).Object;
+
+            var ex = Assert.Throws<StrykerInputException>(() =>
+            {
+                new InputFileResolver().FindProjectUnderTest(new List<IAnalyzerResult> { analyzerResult }, shouldMatchNone);
+            });
+
+            ex.Message.ShouldBe("Project reference issue.");
+            ex.Details.ShouldContain("Use the project name instead of an absolute or relative path", Case.Insensitive);
+        }
     }
 }
--- a/src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
+++ b/src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
@@ -166,12 +166,34 @@ namespace Stryker.Core.Initialisation
             }
             else
             {
+                ValidateProjectUnderTestNameFilter(projectUnderTestNameFilter);
                 projectUnderTestPath = DetermineProjectUnderTestWithNameFilter(projectUnderTestNameFilter, projectReferences);
             }
 
             _logger.LogDebug("Using {0} as project under test", projectUnderTestPath);
 
             return projectUnderTestPath;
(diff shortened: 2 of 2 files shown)


Mobrockers:
I think this is a bit restrictive. Full and relative paths are allowed as long as that's what your reference looks like in your project file. This change would mean that paths are no longer allowed. It could mean that some projects actually can't use stryker due to weird project structures. Example:

```
<ItemGroup>
    <ProjectReference Include="..\..\Stryker.Core\Stryker.Core\Stryker.Core.csproj" />
    <ProjectReference Include="..\..\Stryker.Temp.Core\Stryker.Core\Stryker.Core.csproj" />
  </ItemGroup>
```

This kind of (horrible) project structure would not be supported any longer.
The only thing we can do to fix that issue imo is improve the message itself to mention that the project file reference needs to partially or exactly match the project reference in your csproj.

fbd:
@Mobrockers Do you prefer I create a new PR or edit this one?

How about listing the current references in the project and suggesting the user pick any one of them or add a new reference is the project he intends to mutate is not one of them?  In most (sensible) unit test projects there would only be a handful of project references, so listing them is probably okay (or I could stop printing them once I reach a threshold and then suggest the user pickout a path from the .csproj file manually if he doesn't see anything relevant).

Mobrockers:
> @Mobrockers Do you prefer I create a new PR or edit this one?

You can edit this one, no problem. Whichever you prefer.

> How about listing the current references in the project and suggesting the user pick any one of them or add a new reference if the project he intends to mutate is not one of them? In most (sensible) unit test projects there would only be a handful of project references, so listing them is probably okay (or I could stop printing them once I reach a threshold and then suggest the user pickout a path from the .csproj file manually if he doesn't see anything relevant).

As you can see in the original issue, we do already list the references that the user can choose:

```
Test project contains more than one project references. Please add the --project-file=[projectname] argument to specify which project to mutate.
Choose one of the following references:
  C:\Users\Hannah.Chan\Code\Scaffold.WebApi\Scaffold\Sources\Scaffold.Application\Scaffold.Application.csproj
  C:\Users\Hannah.Chan\Code\Scaffold.WebApi\Scaffold\Sources\Scaffold.Data\Scaffold.Data.csproj
```

However the user here was confused about what was allowed because he sees file paths. He assumed that meant he could use relative paths. This means this users actually has full paths as project references.

fbd:
Ahh, I see. Alright, how about I make the change so that we only print the project file name (instead of the path) and suggest the user pick one or add a new reference? I believe you can not have two references with the same name in any case so it would be redundant to display any kind of path (relative or absolute) and it just confuses the user.  I will make the change if you agree.

Mobrockers:
> Ahh, I see. Alright, how about I make the change so that we only print the project file name, for example _Project.csproj_ (instead of the path _....\Folder\Project.csproj_) and suggest the user pick one or add a new reference? I believe you can not have two references with the same name in any case so it would be redundant to display any kind of path (relative or absolute) and it just confuses the user. I will make the change if you agree.
> 
> Side note: should I worry about the Integration Test runs failing? Seems like some problem with the environment.

If as you say it is illegal to list two references with the same name, then I can agree that the easiest fix is to just display the name and not the path.

fbd:
I was incorrect in believing that two references projects can not have the same name, I was probably confusing references to DLLs.
In any case, Buildalyzer *normalizes* the path, but we can create FindProjectsReferencedByAllTestProjects*WithOutNormalization* and use that when creating the exception messages:

        private static IEnumerable<string> FindProjectsReferencedByAllTestProjectsWithoutNormalizing(IEnumerable<IAnalyzerResult> testProjects)
        {
            var amountOfTestProjects = testProjects.Count();

            var projectReferences = testProjects
                .SelectMany(t => t.Items)
                .Where(i => i.Key == "ProjectReference")
                .SelectMany(t => t.Value)
                .GroupBy(x => x.ItemSpec)
                .Where(g => g.Count() == amountOfTestProjects).Select(g => g.Key)
                .ToList();
            return projectReferences;
        }

I made this change on my local fork and it works (that is, the printed out paths are the same as in the csproj file, not absolute paths used by the builder/compiler), but I break 8 unit tests because the mocked IAnalyzerResult.Items is empty.  I don't think there is a good way to mock this, since the constructor of the ProjectItems is internal to the Buildalyzer project.  So I could do a PR on Buildalyzer that adds an interface for ProjectItems (wait until next release and update Stryker) or do something weird in Stryker just to get those tests to run.

Mobrockers:
If you could add an interface for this in buildalyzer that would be great :)

PR #1410, opened 2021-02-12
Update Reporters.md

psfinaki:
I guess `filteredmutants` is something obsolete, haven't seen it elsewhere in the docs, also the example command throws like this.


Mobrockers:
Filtetedmutants was actually introduced recently but we renamed it during development. It's no problem to remove it from the example. Thank you for these docs improvements! It's much appreciated!

psfinaki:
Thanks for the responsiveness and for the awesome instrument :)

Mobrockers:
Glad you like it :)

Issue #1423, opened 2021-02-19
Compile Error: The name 'x' does not exist in the current context


bcollamore:
Compile Error: The name 'SizeSuffixes' does not exist in the current context

"[08:26:17 WRN] Safe Mode! Stryker will try to continue by rolling back all mutations in method. This should not happen, please report this as an issue on github with the previous error message."

Note:  The run succeeded after rolling back

 - OS: Linux
 - Type of project: Core
 - 3.1
 - Stryker Version: 0.21.1

Hopefully relevant logs attached, as well as original method being mutated:
[StrykerLogs.txt](https://github.com/stryker-mutator/stryker-net/files/6010457/StrykerLogs.txt)

Mobrockers:
Thank you for the reporting!

richardwerkman:
It seems you use a way of array decraration we haven't seen before. We should look into how this is mutated and add some tests for this secific way of declaration:

``` cs
string[] SizeSuffixes = { "bytes", "KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB" };
```

Thanks for reporting!

dupdob:
Well, this is embarrassing.
The issue is not due to the array declaration syntax, which woks fine. It simply that array declaration used for a variable declarations lead to this kind of issues.
This should have been identified earlier, but unit tests sadly used too simple examples to spot this.
Working on a fix.

Issue #1426, opened 2021-02-22
Stryker fails to mutate in version 0.21.1


laurenceSaes:
**Describe the bug**
Stryker will fail to mutate after upgrading from version 0.19 to 0.21.1. The ExpressionBody of a sourceNode is null what will result in a null reference exception:

The exception:
![image](https://user-images.githubusercontent.com/518177/108678634-9f14a900-74eb-11eb-8f59-49c0657592d3.png)

The source code to mutate:

```
get
      {
        var useValue = SettingsModel[XServerSettingsModel.FeatureBatchSizeSettingName].Value as int? ?? XModule.DefaultFeatureBatchSize;
        return useValue > 0 ? useValue : XModule.DefaultFeatureBatchSize;
      }
```

**Logs**

The following error is printed into the log file: [ErrorLog.txt](https://github.com/stryker-mutator/stryker-net/files/6020356/ErrorLog.txt)

**Expected behavior**
I would expect that Stryker does not fail to mutate this line of code. When it does, I would expect that Stryker would continue executing and log after completion that a specific mutation could not be performed.

**Desktop (please complete the following information):**

 - OS: Windows 10
 - Type of project: framework
 - Framework Version: net4.8
 - Stryker Version: 0.21.1

dupdob:
Hi, thanks for your detailed report. 
I need your help. I tried to reproduce the issue with the bit of code you provided and it works as expected with the WIP version. I am confident no one has tried to fix it yet, so I suspect the crash may  be due to some other part of the code.
Can you provide the complete definition of the property?

laurenceSaes:
Hello,

This is the complete definition of the property:

```
    /// <summary>
    /// Gets the configured feature batch size.
    /// </summary>
    public int FeatureBatchSize
    {
      get
      {
        var useValue = SettingsModel[XServerSettingsModel.FeatureBatchSizeSettingName].Value as int? ?? XModule.DefaultFeatureBatchSize;
        return useValue > 0 ? useValue : XModule.DefaultFeatureBatchSize;
      }
    }
```

The mutation generation succeeds on the master branch, so it seems like that this issue is already resolved.

dupdob:
Thanks for reverting with the requested details. You are right, the issue is indeed resolved and will be part of the next release.
Your problem relates to issue #1386.
[...]

PR #1427, opened 2021-02-22
fix: local declaration mutations are promoted from statement to block level mutations

dupdob:
Fix issues #1423 (and related). Statement level mutations are promoted as blocklevel mutations for LocalDeclaration statement, to prevent them being hidden in a subscope when mutated.


Mobrockers:
Quick fix as always 💖

Btw I changed the target branch to master. We have decided to move the main branch to develop to focus on a 1.0 release and to make sure no new features make it on to master, which has the side effect that the dev branch is automatically selected for pull requests. Bug fixes will still go to master so we can still release fixes without new features.

dupdob:
I failed to understand your remark about branches at first! Now I do. I did not notice that Github selected `v1.0' as the target.
I was not even aware there was v 1 branch.

I'll try to find a fix for the #1426, I can not reproduce for now. I can give details on the slack if interested.

src/Stryker.Core/Stryker.Core.UnitTest/Mutants/MutantOrchestratorTests.cs
```diff
@@ -886,6 +886,43 @@ public void ShouldMutatePropertiesAsArrowExpression()
             ShouldMutateSourceToExpected(source, expected);
         }
 
+
+        [Fact]
+        public void ShouldNotExhibitIssue1423()
```
Stijn-Rutten, line 891: Could you maybe add a more descriptive name to this test? I would like a clearer name that allows me to see the problem at first glace. Maybe reference the issue in the description so when we need more info we can reference that.

src/Stryker.Core/Stryker.Core/Mutants/NodeOrchestrators/LocalDeclarationOrchestrator.cs
```diff
@@ -0,0 +1,32 @@
...
+            if (node.IsConst)
+            {
+                // don't mutate const declaration statement
+                return node;
+            }
+
+            var result = base.OrchestrateChildrenMutation(node, context);
+            // statement level mutations need to be changed to block level
+            context.BlockLevelControlledMutations.AddRange(context.StatementLevelControlledMutations);
+            context.StatementLevelControlledMutations.Clear();
```
richardwerkman, line 24: I'm thinking this is the solution right? Can you explain why this solves our issue?

PR #1429, opened 2021-02-22
fix: Stryker.CLI fails to start on MacOSX due to missing microsoft.testplatform package

MindWrapper:
Added missed binary, corrected .gitignore, and added tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -161,6 +161,7 @@ PublishScripts/
 
 # NuGet Packages
 *.nupkg
+!microsoft.testplatform.portable*.nupkg
 # The packages folder can be ignored because of Package Restore
 **/packages/*
 # except build/, which is used as an MSBuild target.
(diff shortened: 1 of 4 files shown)


src/Stryker.Core/Stryker.Core/Stryker.Core.csproj
```diff
@@ -32,6 +32,8 @@
     <EmbeddedResource Include="Reporters\HtmlReporter\Files\mutation-report.html" />
 
     <EmbeddedResource Include="ToolHelpers\.vstest\Microsoft.TestPlatform.Portable\*\*.nupkg" />
+
+    <EmbeddedResource Include="microsoft.testplatform.portable.16.8.3.nupkg" />
```
Mobrockers, line 36: We already embed this just above this line. Why do you embed it again for a specific version number?

.gitignore
```diff
@@ -161,6 +161,7 @@ PublishScripts/
 
 # NuGet Packages
 *.nupkg
+!microsoft.testplatform.portable*.nupkg
```
Mobrockers, line 164: The nupkg should not be included in git.

Mobrockers:
Hello 🙋‍♀️

We actually load microsoft.testplatform.portable in the build pipeline. It should not be included in git. I am also pretty sure that the package is not missing from the currently released stryker package. I think something else is going wrong here.

Mobrockers:
Thank you for taking the time to try and fix this! Unfortunately I don't think this is actually a problem with the nuget package.

Why it's not working when running from source I do not know, but I don't think this is the fix for that problem. Appreciate that you were willing to fix a problem you found yourself anyway!

PR #1453, opened 2021-03-03
Update Configuration.md

psfinaki:
Looks like this diff section got outdated in docs - figured it out from `dotnet stryker --help`. The new description is taken from the help as well.


Mobrockers:
Thank you 🙈

Issue #1470, opened 2021-03-13
Mutator for individual statements


ivancea:
**Is your feature request related to a problem? Please describe.**
It would be useful to have single statement mutators (removing method calls, assignments, any statement)

**Describe the solution you'd like**
There is already a block statement mutator (not in Stryket.NET as I see in the docs). I think it may be modified to remove individual statements instead of complete blocks. It would generate more "unitary" mutations.

**Example:**
```cs
void Method() {
    a.Method(); // Mutation 1: Removed
    b.Method(); // Mutation 2: Removed
}
```

ivancea:
I'll work on this if there is not a problem with it 😃
I'll just replace every StatementSyntax with an EmptyStatementSyntax, with some exceptions:
- ReturnStatementSyntax: It may result in an error if it is the last return of a non-void function. The "last return" part is relatively complex, so I'll probably just check if it is a void return.
- LocalDeclarationStatementSyntax: Removing a declaration will probably generate an error, unless the variable isn't used, in which case there is already a compile time warning probably.
- BlockSyntax: Some BlockSyntax can be safely removed (if-else children, for example), but others cannot (method/lambda blocks), so a list of parent types to avoid may be fine

If I have time, I'll create a PR soon™.
It will be different from the JS block mutator, so if I should exclude blocks or move it to another mutator, just tell me 🙂

PR #1472, opened 2021-03-13
feat(statement mutator): Add mutator that removes statements

ivancea:
issue: #1470

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutators/StatementMutatorTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutators/StatementMutatorTests.cs
@@ -20,8 +20,11 @@ namespace Stryker.Core.UnitTest.Mutators
 
         public static IEnumerable<object[]> MutableStatements => new List<object[]>
         {
+            new object[] { SyntaxFactory.ReturnStatement() },
             new object[] { SyntaxFactory.BreakStatement() },
             new object[] { SyntaxFactory.ContinueStatement() },
+            new object[] { SyntaxFactory.GotoStatement(SyntaxKind.GotoStatement) },
+            new object[] { SyntaxFactory.ThrowStatement() },
             new object[] { SyntaxFactory.YieldStatement(SyntaxKind.YieldBreakStatement) },
             new object[]
             {
@@ -30,12 +33,9 @@ namespace Stryker.Core.UnitTest.Mutators
                     SyntaxFactory.LiteralExpression(SyntaxKind.NumericLiteralExpression, SyntaxFactory.Literal(1))
                 )
             },
-            new object[] { SyntaxFactory.ReturnStatement() },
-            new object[] {
-                SyntaxFactory.IfStatement(
-                    SyntaxFactory.LiteralExpression(SyntaxKind.TrueLiteralExpression),
-                    SyntaxFactory.EmptyStatement()
-                )
+            new object[]
+            {
+                SyntaxFactory.ExpressionStatement(SyntaxFactory.AwaitExpression(SyntaxFactory.LiteralExpression(SyntaxKind.NullLiteralExpression)))
             },
         };
 
@@ -61,13 +61,27 @@ namespace Test
 {{
     class Program
     {{
-        delegate void TestDelegate();
-
         static int Method()
         {{
             int variable = 0;
-            TestDelegate lambda = () => {{ }};
-            TestDelegate anonymousMethod = delegate () {{ }};
+
+            Method(out var x);
+
+            if (x is Type v)
+            {{
+                return;
+            }}
+
+            switch (x)
+            {{
+                case 1:
+                    return;
+                    break;
+                    continue;
+                    goto X;
+                    throw x;
+            }}
+
             return 1;
         }}
     }}
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutators/StatementMutatorTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutators/StatementMutatorTests.cs
@@ -18,31 +18,27 @@ namespace Stryker.Core.UnitTest.Mutators
             target.MutationLevel.ShouldBe(MutationLevel.Standard);
         }
 
-        public static IEnumerable<object[]> MutableStatements => new List<object[]>
-        {
-            new object[] { SyntaxFactory.ReturnStatement() },
-            new object[] { SyntaxFactory.BreakStatement() },
-            new object[] { SyntaxFactory.ContinueStatement() },
-            new object[] { SyntaxFactory.GotoStatement(SyntaxKind.GotoStatement) },
-            new object[] { SyntaxFactory.ThrowStatement() },
-            new object[] { SyntaxFactory.YieldStatement(SyntaxKind.YieldBreakStatement) },
-            new object[]
-            {
-                SyntaxFactory.YieldStatement(
-                    SyntaxKind.YieldReturnStatement,
-                    SyntaxFactory.LiteralExpression(SyntaxKind.NumericLiteralExpression, SyntaxFactory.Literal(1))
-                )
-            },
-            new object[]
-            {
-                SyntaxFactory.ExpressionStatement(SyntaxFactory.AwaitExpression(SyntaxFactory.LiteralExpression(SyntaxKind.NullLiteralExpression)))
-            },
-        };
-
         [Theory]
-        [MemberData(nameof(MutableStatements))]
-        private void ShouldMutate(StatementSyntax statement)
+        [InlineData("return;")]
+        [InlineData("break;")]
+        [InlineData("continue;")]
+        [InlineData("goto test;")]
+        [InlineData("throw null;")]
+        [InlineData("yield break;")]
+        [InlineData("yield return 0;")]
+        [InlineData("await null;")]
+        private void ShouldMutate(string statementString)
         {
+            var source = $@"class Test {{
+                void Method() {{
+                    {statementString}
+                }}
+            }}";
+
+            var tree = CSharpSyntaxTree.ParseText(source).GetRoot();
+
+            var statement = tree.DescendantNodes().OfType<StatementSyntax>().Where(s => s is not BlockSyntax).First();
+
             var target = new StatementMutator();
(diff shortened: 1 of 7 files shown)


ivancea:
Still working on MutantOrchestratorTests tests that didn't pass after this mutator

src/Stryker.Core/Stryker.Core/Mutators/Mutator.cs
```diff
@@ -27,6 +27,8 @@ public enum Mutator
         Linq,
         [Description("String literals")]
         String,
+        [Description("Statements")]
```
ivancea, line 30: Not sure about the order of these elements. Same for CsharpMutantOrchestrator

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/StatementMutatorTests.cs
```diff
@@ -0,0 +1,86 @@
...
+            new object[] {
+                SyntaxFactory.IfStatement(
+                    SyntaxFactory.LiteralExpression(SyntaxKind.TrueLiteralExpression),
+                    SyntaxFactory.EmptyStatement()
+                )
+            },
+        };
+
+        [Theory]
+        [MemberData(nameof(MutableStatements))]
```
richardwerkman, line 43: We prefer to use InlineData instead of MemberData. And since the `member` is not used anywhere else I'd like to see InlineData here as well

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/StatementMutatorTests.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        [Fact]
+        private void ShouldNotMutate()
+        {
+            SyntaxTree tree = CSharpSyntaxTree.ParseText($@"
+namespace Test
+{{
+    class Program
+    {{
+        delegate void TestDelegate();
```
richardwerkman, line 64: Why is this here?

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/StatementMutatorTests.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        static int Method()
```
richardwerkman, line 66: I'm missing the empty statement?

richardwerkman (changes requested):
Thanks for creating this mutator! This will be a nice addition to stryker 😄 

Ps. You will have to update the integration tests as well, since this will introduce a lot of new mutations. We should probably set the mutation level to `Complete` in the integration test to make sure we also test these mutations.

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/StatementMutatorTests.cs
ivancea: I used MemberData because the statements aren't constants, and I cannot construct them inside InlineData (unless I'm missing something)

ivancea:
I'll have to rethink this mutator completely. I'm finding compilation problems that aren't being rollbacked and are throwing errors:

1. Any variable assignation (and any statement having an asignation inside) may result in an error if they are assigning an unassigned variable. Example:
```Cs
string v;
if (..) {
    v = "";
} else {
    v = " ";
}
```

2. The last statement inside a switch case will throw an error if removed (return/throw/break/goto)

I'll switch from the "forbidden syntaxes" approach to a "allowed syntaxes" one, and document avery case I find. The most important problem now is the assignment thing, as it isn't trivial to find which assignments (or blocks containing them) are removable. As in devMode they are resulting in a critical error, I'm assumming it's not acceptable. I'm converting this PR to draft

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/StatementMutatorTests.cs
```diff
@@ -0,0 +1,86 @@
...
+
+namespace Stryker.Core.UnitTest.Mutators
+{
+    public class StatementMutatorTests
+    {
+        [Fact]
+        public void ShouldBeMutationlevelStandard()
+        {
+            var target = new StatementMutator();
+            target.MutationLevel.ShouldBe(MutationLevel.Standard);
```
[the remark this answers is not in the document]
ivancea: After the rework, it added about ~25% extra mutators to a project I tested (NFluent, ~2000 -> ~2500).
So, as you like. I would say it's ok in standard. Method calls may produce untested side effects, as well as breaks returns..., but there are some "repeated" mutations (x++; -> x--; & ;) that may be considered redundant, so maybe advanced?

src/Stryker.Core/Stryker.Core/Mutators/StatementMutator.cs
```diff
@@ -0,0 +1,53 @@
...
+using Stryker.Core.Mutants;
+using System.Linq;
+
+namespace Stryker.Core.Mutators
+{
+    public class StatementMutator: MutatorBase<StatementSyntax>, IMutator
+    {
+        public override MutationLevel MutationLevel => MutationLevel.Standard;
+
+        private static readonly HashSet<SyntaxKind> ForbiddenSyntaxes = new HashSet<SyntaxKind>() {
```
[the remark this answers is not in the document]
ivancea: 100% reworked to just allow some syntaxes

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/StatementMutatorTests.cs
ivancea: Removed. used it to check lambdas and anonymous methods, but it isn't necessary now

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/StatementMutatorTests.cs
ivancea: I'm not sure what you mean. Te test it isn't mutated?
After the rework, with a whitelist of mutable statements, I'm not sure if I should add every non-whitelisted statement to the test. What do you think? Every other StatementSyntax is commented in the StatementMutator whitelist

ivancea:
Reopenign after reworking it :)

ivancea:
Not sure about why are the Windows integration tests failing

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/StatementMutatorTests.cs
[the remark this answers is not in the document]
Mobrockers: With the rewrite of this mutator we think standard is fine for now.

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/StatementMutatorTests.cs
Mobrockers: If you look at other tests you'll see that we place the syntax as string in InlineData and parse it to syntax in the unit test. This is more readable because you can clearly see what syntax we're testing instead of having to build up a mental model from a SyntaxTree.

Mobrockers:
> Not sure about why are the Windows integration tests failing

It is not the integration test that is failing, it is a unit test:

https://dev.azure.com/stryker-mutator/Stryker/_build/results?buildId=6939&view=ms.vss-test-web.build-test-results-tab&runId=1067866&resultId=100404&paneView=debug

```
Shouldly.ShouldAssertException : isSame
should be
True
but was
False

Additional Info:
AST's are not equivalent. Line[1]
actual:(Stryker6Gp9tRNthZmz0ED.MutantControl.IsActive(ID)?1-8:1+8)
expect:(Stryker6Gp9tRNthZmz0ED.MutantControl.IsActive(16)?1-8:1+8)
Actual(full):
(Stryker6Gp9tRNthZmz0ED.MutantControl.IsActive(ID)?1-8:1+8)
, expected:
(Stryker6Gp9tRNthZmz0ED.MutantControl.IsActive(16)?1-8:1+8)
```

```
at Stryker.Core.UnitTest.AssertionExtensions.ShouldBeSemantically(SyntaxNode actual, SyntaxNode expected) in D:\a\1\s\src\Stryker.Core\Stryker.Core.UnitTest\AssertExtensions.cs:line 52
   at Stryker.Core.UnitTest.AssertionExtensions.ShouldBeSemantically(String actual, String expected) in D:\a\1\s\src\Stryker.Core\Stryker.Core.UnitTest\AssertExtensions.cs:line 22
   at Stryker.Core.UnitTest.Mutants.MutantPlacerTests.MutantPlacer_ShouldPlaceWithConditionalExpression(Int32 id) in D:\a\1\s\src\Stryker.Core\Stryker.Core.UnitTest\Mutants\MutantPlacerTests.cs:line 61
```

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/StatementMutatorTests.cs
ivancea: Ok, changed

ivancea:
The unit test that were failing in the pipeline (not in my local) isn't failing now; I don't know what happened 🙃

src/Stryker.Core/Stryker.Core/Mutators/StatementMutator.cs
```diff
@@ -0,0 +1,110 @@
...
+    public class StatementMutator: MutatorBase<StatementSyntax>, IMutator
+    {
+        public override MutationLevel MutationLevel => MutationLevel.Standard;
+
+        private static readonly HashSet<SyntaxKind> AllowedSyntaxes = new HashSet<SyntaxKind>()
+        {
+            // SyntaxKind.EmptyStatement, // useless mutation
+
+            // SyntaxKind.Block, // unitary mutation should prevail over blocks
+            // SyntaxKind.CheckedStatement, // not unitary
```
Mobrockers, line 18: Could you clarify what the comment not unitary means on these types of statements

ivancea: As "not unitary" I meant every statement that contains more statements inside (usually, containing blocks)

PR #1492, opened 2021-03-26
Improved debug message.

psfinaki:
Suggesting this after I noticed a message `Mutations will be tested in 10 test runs instead of 10.` :)

Discard if this actually represents a more serious problem.


richardwerkman:
This can certainly happen so this is a nice improvement. Thanks for your contribution 👍

PR #1495, opened 2021-03-30
feat(multi-target): Prepare project analyzer for multi-target support

psfinaki:
As agreed, this doesn't implement the option itself, just a minimal API change.

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
+++ b/src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
@@ -11,7 +11,10 @@ namespace Stryker.Core.Initialisation
 {
     public interface IProjectFileReader
     {
-        IAnalyzerResult AnalyzeProject(string projectFilePath, string solutionFilePath);
+        IAnalyzerResult AnalyzeProject(
+            string projectFilePath,
+            string solutionFilePath,
+            string targetFramework = null);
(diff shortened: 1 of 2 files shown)


src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
```diff
@@ -25,7 +28,10 @@ public ProjectFileReader(INugetRestoreProcess nugetRestoreProcess = null)
             _logger = ApplicationLogging.LoggerFactory.CreateLogger<ProjectFileReader>();
         }
 
-        public IAnalyzerResult AnalyzeProject(string projectFilePath, string solutionFilePath)
+        public IAnalyzerResult AnalyzeProject(
+            string projectFilePath,
+            string solutionFilePath,
+            string targetFramework)
```
richardwerkman, line 34: Why does the targetFramework not have a default null value? It has on the interface. If we add that the unit tests also could use the default

richardwerkman (commented):
Thanks for your contribution

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
Mobrockers, line 34:
```suggestion
            string targetFramework = null)
```

Issue #1501, opened 2021-04-07
Bug? in Mono.Cecil causes confusing warnings/errors in CLI runner


FrankvdStam:
**Describe the bug**
When running on a .NET framework project, I'm getting a confusing warning and the same exception is raised 3 times. After this, it properly runs the tests with the mutations.

**Logs**
[16:39:25 INF] Identifying project to mutate.
[16:39:27 INF] The project [path to class library I want mutated] will be mutated.
[16:39:29 INF] Analysis complete.
[16:39:29 INF] Building test project [Path to project with unit tests] (1/1)
[16:39:45 INF] Initializing test runners (VsTest)
[16:39:49 INF] Total number of tests found: 28
[16:39:49 INF] Initial testrun started
[16:39:53 INF] Using 10764 ms as testrun timeout
[16:39:56 WRN] Original project under test [path to class library I want mutated] could not be loaded.
Embedded Resources might be missing.
Mono.Cecil.AssemblyResolutionException: Failed to resolve assembly: 'System.Data.Entity, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'
   at Mono.Cecil.BaseAssemblyResolver.Resolve(AssemblyNameReference name, ReaderParameters parameters)
   at Mono.Cecil.BaseAssemblyResolver.Resolve(AssemblyNameReference name)
   at Mono.Cecil.DefaultAssemblyResolver.Resolve(AssemblyNameReference name)
   at Mono.Cecil.MetadataResolver.Resolve(TypeReference type)
   at Mono.Cecil.ModuleDefinition.Resolve(TypeReference type)
   at Mono.Cecil.TypeReference.Resolve()
   at Mono.Cecil.Mixin.CheckedResolve(TypeReference self)
   at Mono.Cecil.SignatureReader.ReadCustomAttributeEnum(TypeReference enum_type)
   at Mono.Cecil.SignatureReader.ReadCustomAttributeElementValue(TypeReference type)
   at Mono.Cecil.SignatureReader.ReadCustomAttributeElement(TypeReference type)
   at Mono.Cecil.SignatureReader.ReadCustomAttributeFixedArgument(TypeReference type)
   at Mono.Cecil.SignatureReader.ReadCustomAttributeConstructorArguments(CustomAttribute attribute, Collection`1 parameters)
   at Mono.Cecil.MetadataReader.ReadCustomAttributeSignature(CustomAttribute attribute)
[...]

PR #1502, opened 2021-04-07
Potential fix for assembly resolving issues

FrankvdStam:
Potential fix for https://github.com/stryker-mutator/stryker-net/issues/1501

Please run the tests for the core project - I was unable to get them running so am unsure if they still pass.

Commit message:
Copied BaseAssemblyResolver from Mono.Cecil into CustomAssemblyResolver.cs, removed conditional compilation. First it tries to resolve the .NET core way, if that doesn't work it proceeds with the .NET framework way.

EmbeddedResourcesGenerator.cs now passes the added CustomAssemblyResolver to LoadModule call to override the default resolving behavior. This object is re-created for each call, maybe it can be instantiated once? Not sure what the lifecycle should be.

### DIFF
--- a/src/Stryker.CLI/Stryker.CLI.UnitTest/packages.lock.json
+++ b/src/Stryker.CLI/Stryker.CLI.UnitTest/packages.lock.json
@@ -1,7 +1,7 @@
 {
   "version": 1,
   "dependencies": {
-    ".NETCoreApp,Version=v3.1": {
+    ".NETCoreApp,Version=v5.0": {
       "Microsoft.CodeCoverage": {
         "type": "Direct",
         "requested": "[16.9.4, )",
@@ -297,8 +297,7 @@
           "Microsoft.Extensions.Configuration": "5.0.0",
           "Microsoft.Extensions.Configuration.Abstractions": "5.0.0",
           "Microsoft.Extensions.Configuration.FileExtensions": "5.0.0",
-          "Microsoft.Extensions.FileProviders.Abstractions": "5.0.0",
-          "System.Text.Json": "5.0.0"
+          "Microsoft.Extensions.FileProviders.Abstractions": "5.0.0"
         }
       },
       "Microsoft.Extensions.DependencyInjection": {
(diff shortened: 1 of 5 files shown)


rouke-broersma:
Hi @FrankvdStam 👋 

I've been thinking about this fix for some time and I am a bit hesitant to merge this.

1. I don't like the idea of being responsible for this piece of code, since I have very little understanding of what it does
2. Because I have very little understanding of what it does, I cannot judge the impact this might have on projects where embedded resource resolving currently works perfectly fine

For that reason I'm going to attempt to get in contact with the Mono.Cecil team first and ask them to judge if this is a good idea. Hope you don't mind, I really appreciate the effort!

FrankvdStam:
Understandable - it is a bit dubious. 

Looked at the fix you included, does not look like it would be relevant here - it still uses conditional compilation. See, Ryder is .NET core. This causes only .NET core specific resolving code to be used in the cecil default resolver. The proposed fix would solve this for .NET standard - but only at compile - building Ryder as .NET standard might work, but will in the end probably have the same issues when resolving framework libraries. 

As far as I can see you need dynamic, runtime checking on what path to take: core/standard/framework. Or you're stuck in your branch: Core applications will likely have problems resolving framework libraries and framework apps will likely have issues resolving core libraries. It does not look like that PR includes a fix like that.

Maybe someone from cecil will end up reading this and help them along or maybe I don't really understand it enough either and am wasting everyone's time, these are just my findings.

rouke-broersma:
Maintainers of Mono.Cecil seem to be quite active with replying to questions. If they end up not answering soon enough we will probably still go with this fix you propose. I am just looking for ways to avoid having to maintain this particular piece of magic if at all possible :p

FrankvdStam:
Maybe I should propose this fix over at cecil? I just figured it would take far too long.

rouke-broersma:
Response to pull requests at Cecil seem kinda hit or miss. Not sure why some PR's get merged within the same day and some don't even get a comment. The questions seem to get more response. I agree that it would probably take a long time before a change like this would be accepted by Cecil if at all. Let's see what they reply to my questions and go from there.

rouke-broersma:
@FrankvdStam The maintainer of Cecil has replied that the default assembly resolver in Cecil is for the generic case and that any project is free to implement their own resolver to their own requirements. It seems like this PR is the way to go.

I would like to take some time to clean test, understand (ish) and clean up the custom resolver. There are things marked 'internal for testing' for example that we're not using and can probably safely be removed.

FrankvdStam:
@rouke-broersma okay, my time is limited (priorities yadayada) but let me know what you need from me.

rouke-broersma:
Integrated in #1564

PR #1518, opened 2021-04-18
Fix issues with resolving vstest location

sbergen:
I didn't run into any of these issues when running the package gotten from NuGet, but rather had them in local development.

* Look in `$HOME` in addition to `$USERPROFILE` (for UNIX-systems)
* Use `netcoreapp2.1` instead of `netcoreapp2.0`
  * 2.0 has reached end-of-life, so this should be ok, right?
* Ignore case when looking at embedded vstest resource names
  * I faced this issue before I noticed we should look in `$HOME` also, and had ran the manual `dotnet restore`, which is used in CI

### DIFF
--- a/src/Stryker.Core/Stryker.Core/ToolHelpers/VsTestHelper.cs
+++ b/src/Stryker.Core/Stryker.Core/ToolHelpers/VsTestHelper.cs
@@ -124,7 +124,7 @@ namespace Stryker.Core.ToolHelpers
                     dllPath = FilePathUtils.NormalizePathSeparators(
                         Path.Combine(
                             nugetPackageFolder, portablePackageFolder, versionString,
-                            "tools", "netcoreapp2.0", "vstest.console.dll"));
+                            "tools", "netcoreapp2.1", "vstest.console.dll"));
                     exePath = FilePathUtils.NormalizePathSeparators(
                         Path.Combine(
                             nugetPackageFolder, portablePackageFolder, versionString,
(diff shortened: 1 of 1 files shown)


sbergen:
Hmm, seems I wasn't running all the tests locally. Somehow test discovery isn't working very well in Rider with this project. I'll see if I can figure out why that is also...

rouke-broersma:
> Hmm, seems I wasn't running all the tests locally. Somehow test discovery isn't working very well in Rider with this project. I'll see if I can figure out why that is also...
> 
> EDIT: Well, actually it seems like all tests pass in Rider, but not on the command line. Will look into it.
> 
> EDIT2: Seems like test results on the CLI are flaky, even on master :( 

Yes sorry about that. We have one or two flaky tests and no clue how they can be flaky..

sbergen:
> > Hmm, seems I wasn't running all the tests locally. Somehow test discovery isn't working very well in Rider with this project. I'll see if I can figure out why that is also...
> > EDIT: Well, actually it seems like all tests pass in Rider, but not on the command line. Will look into it.
> > EDIT2: Seems like test results on the CLI are flaky, even on master :(
> 
> Yes sorry about that. We have one or two flaky tests and no clue how they can be flaky..

Could it be related to test being run in parallel? I ran `dotnet test -p:ParallelizeTestCollections=false` many times, and didn't get a single failure yet.

rouke-broersma (approved):

PR #1519, opened 2021-04-18
fix: Allow setting break threshold to 100%

sbergen:
Previously only the low and high thresholds were allowed to be the same, and thus the maximum break threshold you could set was 99%. I don't see any reason why this should be so, and would like to set the break threshold to 100% in my project, which was [fairly easy to reach](https://github.com/sbergen/Responsible/pull/72).


rouke-broersma:
I think this change is valid in the case of all thresholds being 100, but on lower thresholds this leads to a weird scenario.

The thresholds are the lower boundary for the state so let's say you have high 60, low 60 and break 60. If you have a score of 60 this would mean that your score is both good enough to be green and bad enough to break the build.

We are also reworking all the options, I would prefer to track this change with the changing of the options. That does mean that this change would not be available until we are done with the options rework and can release v1.0.

PR #1522, opened 2021-04-18
dev: Simpler vstest.console discovery on *nix

0xced:
Pull request #1518 fixed the discovery of the vstest console inside the NuGet packages directory for macOS and Linux. This commits work on top of it to make it even more future-proof.

* Safer because the `netcoreapp2.1` and `net451` target framework monikers are not hardcoded anymore so it becomes more future-proof if Microsoft ever updates those.

* Simpler because it uses `Environment.GetFolderPath(Environment.SpecialFolder.UserProfile` instead of probing both `USERPROFILE` and `HOME` environment variables.


rouke-broersma:
Awesome :)

0xced:
Thanks for merging this pull request so quickly!

PR #1523, opened 2021-04-18
Automatically disable DiffEngine when running tests

0xced:
Disable DiffEngine so that approval tests frameworks such as [Verify](https://github.com/VerifyTests/Verify) or [ApprovalTests.Net](https://github.com/approvals/ApprovalTests.Net) (which both use DiffEngine under the hood) don't launch a diffing tool GUI on each failed test.

### DIFF
--- a/src/Stryker.Core/Stryker.Core/TestRunners/VsTest/StrykerVstestHostLauncher.cs
+++ b/src/Stryker.Core/Stryker.Core/TestRunners/VsTest/StrykerVstestHostLauncher.cs
@@ -41,6 +41,14 @@ namespace Stryker.Core.TestRunners.VsTest
                 UseShellExecute = false,
                 RedirectStandardOutput = true,
                 RedirectStandardError = true,
+                EnvironmentVariables =
+                {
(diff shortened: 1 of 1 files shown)


rouke-broersma:
This sounds more like a feature than a bugfixes, though the impact is likely small. @richardwerkman you think this is fine to put on master?

@0xced  Would there ever be a case where someone would actually want to disable this behavior? My instinct says probably not but I am not familiar with this class of test tooling.

0xced:
> Would there ever be a case where someone would actually want to disable this behavior?

Approval testing is useful when doing TDD and running test by test but running mutation tests is the opposite, many tests are going to fail (the diffing tool are launched only on failed tests). So I can't imagine a scenario in which you would want dozen or maybe even hundreds of diffing tool instances to start when running the mutation tests.

richardwerkman:
I think we can see this as a bug fix. It's undesired behaviour and I don't see any harm in releasing this before v1.

rouke-broersma:
@0xced this fix will be released shortly

PR #1524, opened 2021-04-18
Fix flaky tests by not running them in parallel

sbergen:
I observed that one of the issues was related to the static `JsonReport._report` field. It would be better to get rid of this static caching, by passing the state around explicitly instead, but for now, fixing the flaky tests is probably still a good idea.

I've had the test running with `while dotnet test --no-build; do :; done` for several minutes now without failure.


sbergen:
A quick regex search reveals quite a few of these possibly problematic static and not `readonly` fields:
image

rouke-broersma:
I don't think passing the static cache explicitly is realistic with the current reporting infrastructure. 

I feel like this fix is a bit too aggressive but at the same time it is too easy to forget to place conflicting tests in the same collection and solve the parallelism problem that way. It seems like the impact on test duration is also negligible due to the way xunit does parallelism anyway. Seems good enough for now.

PR #1526, opened 2021-04-19
feat(baseline): Azure fileshare uses projectname as output path if specified

### LINKED ISSUE #1525: Subfolders are not created on AzureStorageFile
**Describe the bug**
When specifying a subfolder in the azure-storage-url (https://storageaccountname.file.core.windows.net/azurefilesharename/my-project-key-01), then it results in an error creating the directories on storage.

I'm not sure if it a bug or a feature. Let me describe the use case we looking for.

We have several projects, which should upload their mutation test reports. As they are microservices or individual libraries, they are usually independent artifacts and have their own release cycle and also their own baselines.

**Azure Storage / Azure Files**
Azure File Storage Immutable Url: https://storageaccountname.file.core.windows.net
Azure File Share Name: azurefilesharename
Resulting URL: https://storageaccountname.file.core.windows.net/azurefilesharename

So, resulting URL is the base for all projects.

Every Project have a key like: `my-project-key-01`, ...

Resulting URL: https://storageaccountname.file.core.windows.net/azurefilesharename/my-project-key-01, ...

**Predefined  `_outputPath` directories from Stryker.NET**

OutputPath: `StrykerOutput/Baselines/<version>`
Parameter Version: 2.0.0-beta0001 

Resulting URL: https://storageaccountname.file.core.windows.net/azurefilesharename/my-project-key-01/StrykerOutput/Baselines/2.0.0-beta0001

I'm happy to contribute with PR again on this, if this is expected behavior.

**Logs**
```
[16:25:43 DBG] Could not locate the current branch name, using project version instead: 2.0.0-beta0270
[16:25:44 DBG] No baseline was found at https://xxx.file.core.windows.net/yyy/the-sub-folder/StrykerOutput/Baselines/dashboard-compare/2.0.0-beta0270/stryker-report.json
[16:25:44 DBG] Creating directories for file https://xxx.file.core.windows.net/yyy/the-sub-folder/StrykerOutput/Baselines/dashboard-compare/2.0.0-beta0270/stryker-report.json
[16:25:44 DBG] Creating directory https://xxx.file.core.windows.net/yyy/the-sub-folder/StrykerOutput/
[16:25:44 ERR] Creating directory failed with status Forbidden and message <?xml version="1.0" encoding="utf-8"?><Error><Code>AuthenticationFailed</Code><Message>Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
RequestId:ff76b1ea-401a-0066-0495-3300f5000000
Time:2021-04-17T14:25:44.3194645Z</Message><AuthenticationErrorDetail>sr is mandatory. Cannot be empty</AuthenticationErrorDetail></Error>
[16:25:44 INF] Time Elapsed 00:14:19.0889376
[16:25:44 INF] The final mutation score is 57.59 %
```

**Expected behavior**
1. All subfolders are created, if they not exists. 
2. Job should fail on unsuccessful upload

 - OS: Windows / Linux
 - Type of project: Core
 - Framework Version: .NET SDK 5.0.202
 - Stryker Version: 0.22.3 (latest)

cmenzi:
Closes #1525

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Baseline/Providers/AzureFileShareBaselineProviderTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Baseline/Providers/AzureFileShareBaselineProviderTests.cs
@@ -170,6 +170,168 @@ namespace Stryker.Core.UnitTest.Baseline.Providers
                 ItExpr.IsAny<CancellationToken>());
         }
 
+        [Fact]
+        public async Task Save_Calls_CreateDictionaryWithProjectName_And_AllocateFileLocation_When_Baseline_Does_Not_Exists()
+        {
+            // arrange
+            var projectName = "my-project-name";
+            var projectVersion = "my-project-version";
+
+            var options = new StrykerOptions(azureFileStorageUrl: "https://www.filestoragelocation.com", azureSAS: "AZURE_SAS_KEY", baselineStorageLocation: "azurefilestorage", projectName: projectName);
+
+            var handlerMock = new Mock<HttpMessageHandler>(MockBehavior.Strict);
+
+            var readonlyInputComponent = new Mock<IReadOnlyProjectComponent>(MockBehavior.Loose).Object;
+
+            var jsonReport = JsonReport.Build(options, readonlyInputComponent);
+
+            var expectedGetUri = new Uri($"https://www.filestoragelocation.com/{projectName}/Baselines/{projectVersion}/stryker-report.json?sv=AZURE_SAS_KEY");
+
+            var expectedCreateProjectOutputDirectoryUri = new Uri($"https://www.filestoragelocation.com/{projectName}/?restype=directory&sv=AZURE_SAS_KEY");
+            var expectedCreateBaselinesDirectoryUri = new Uri($"https://www.filestoragelocation.com/{projectName}/Baselines/?restype=directory&sv=AZURE_SAS_KEY");
+            var expectedCreateVersionDirectoryUri = new Uri($"https://www.filestoragelocation.com/{projectName}/Baselines/{projectVersion}/?restype=directory&sv=AZURE_SAS_KEY");
+
+            var expectedFileAllocationUri = new Uri($"https://www.filestoragelocation.com/{projectName}/Baselines/{projectVersion}/stryker-report.json?sv=AZURE_SAS_KEY");
+
+            var expectedUploadContentUri = new Uri($"https://www.filestoragelocation.com/{projectName}/Baselines/{projectVersion}/stryker-report.json?comp=range&sv=AZURE_SAS_KEY");
+
+            handlerMock
+                .Protected()
+                .Setup<Task<HttpResponseMessage>>(
+                "SendAsync",
+                ItExpr.Is<HttpRequestMessage>(requestMessage => requestMessage.RequestUri == expectedGetUri && requestMessage.Method == HttpMethod.Get),
+                ItExpr.IsAny<CancellationToken>())
+                .ReturnsAsync(() => new HttpResponseMessage()
+                {
+                    StatusCode = System.Net.HttpStatusCode.NotFound,
+                    Content = new StringContent(jsonReport.ToJson(), Encoding.UTF8, "application/json")
+                })
+                .Verifiable();
+
+            handlerMock
+                .Protected()
+                .Setup<Task<HttpResponseMessage>>(
+                "SendAsync",
+                ItExpr.Is<HttpRequestMessage>(requestMessage => requestMessage.RequestUri == expectedCreateProjectOutputDirectoryUri && requestMessage.Method == HttpMethod.Put),
+                ItExpr.IsAny<CancellationToken>())
+                .ReturnsAsync(new HttpResponseMessage()
+                {
+                    StatusCode = System.Net.HttpStatusCode.Created
+                })
+                .Verifiable();
+            handlerMock
+                .Protected()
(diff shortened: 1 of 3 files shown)


cmenzi:
Closes #1525

rouke-broersma (changes requested):
This change in itself looks good, but this will change existing behavior and is therefore not a fix. It should go to the v1 branch instead of master.

I should have specified that in your issue sorry.

cmenzi:
Ok, is there any timeline for v1.0?

How does it change the current behavior? Before the `ProjectName` was always explicitly set to `null` if the reporter was not `Dashboard.` And if the reporter was Dashboard then also an API key was mandatory and only then the `ProjectName` was set to the `StrykerOptions`

See: https://github.com/stryker-mutator/stryker-net/pull/1526/commits/6af7547fea72b59913b203f2eec5858690ca5e7c#diff-d59d5228b318e124824edc80daf18bc43fdbcf340305004fc13306f35f7b4bd5L310

This means for everybody who used Dashboard reporter, they haven't used AzureFileStorage reporter at the same time.
If they have used AzureFileStorage reporter, their ProjectName was always set to null.

The only case would be, they specified the `project-name` and have used AzureFileStorage. 

Just trying arguing to to v0.22.4 :-)

src/Stryker.Core/Stryker.Core/Baseline/Providers/AzureFileShareBaselineProvider.cs
```diff
@@ -14,16 +14,25 @@ namespace Stryker.Core.Baseline.Providers
...
         public AzureFileShareBaselineProvider(IStrykerOptions options, HttpClient httpClient = null)
         {
             _options = options;
             _httpClient = httpClient ?? new HttpClient();
             _logger = ApplicationLogging.LoggerFactory.CreateLogger<AzureFileShareBaselineProvider>();
+
+            _outputPath = $"{DefaultOutputDirectoryName}/{BaselinesDirectoryName}";
+            if (!string.IsNullOrWhiteSpace(options.ProjectName))
+            {
+                _outputPath = $"{options.ProjectName}/{BaselinesDirectoryName}";
```
richardwerkman, line 34: I'm not sure about using ProjectName for this. That option already has a purpose in Stryker. If someone needs that option to run Stryker on his project and doesn't need a sub folder in azure fileshare it might be confusing. @rouke-broersma what about creating a new option for this? (we could do that in a seperate PR on the V1 branch after the options rework)

src/Stryker.Core/Stryker.Core.UnitTest/Baseline/Providers/AzureFileShareBaselineProviderTests.cs
```diff
@@ -170,6 +170,168 @@ public async Task Save_Doesnt_Call_CreateDictionary_And_AllocateFileLocation_Whe
...
+
+            handlerMock
+              .Protected()
+              .Verify(
+               "SendAsync",
+               Times.Exactly(1),
+               ItExpr.Is<HttpRequestMessage>(req =>
+                  req.Method == HttpMethod.Put
+                  && req.RequestUri == expectedUploadContentUri),
+                ItExpr.IsAny<CancellationToken>());
```
richardwerkman, line 332: For me just calling `handlerMock.VerifyAll()` is ok. This test is now a lot of lines of code. It makes it harder to maintain. The only downside of VerifyAll is that it doesn't check for Times.Once but I accept that.

richardwerkman (commented):
Thanks for your contribution :)

src/Stryker.Core/Stryker.Core/Baseline/Providers/AzureFileShareBaselineProvider.cs
rouke-broersma: I dislike having so many different options, this is why I already moved projectname to a more general project-info.name in the new options instead of dashboard.name.

Options should be generally applicable to promote reuse imo.

rouke-broersma:
> Ok, is there any timeline for v1.0?
> 
> How does it change the current behavior? Before the `ProjectName` was always explicitly set to `null` if the reporter was not `Dashboard.` And if the reporter was Dashboard then also an API key was mandatory and only then the `ProjectName` was set to the `StrykerOptions`

> This means for everybody who used Dashboard reporter, they haven't used AzureFileStorage reporter at the same time.
> If they have used AzureFileStorage reporter, their ProjectName was always set to null.
> 
**> The only case would be, they specified the `project-name` and have used AzureFileStorage.**

Yep, agree. This would be the only case. But someone could have done this, and now their baselines are suddenly in a different place. We cannot know who uses stryker this way. It's not very likely someone did this, it's not even likely anyone except you uses the AzureFileStorage baseline provider at all. But strictly speaking it's at best a new feature and at worst even a breaking change (existing baselines will not be used) and definitely not a fix. If we were on v1 I would for sure not accept this as a fix but since we are still at version 0 and this change is very very unlikely to affect anyone at all I am on the fence. @richardwerkman what do you think? Can we bring this to v0 as a fix since it's so unlikely anyone would be impacted?

rouke-broersma:
> So, I would expect breaking changes in a pre-1.0 version. Even SemVer states this in [clause 4](https://semver.org/#spec-item-4).

It is allowed per semver but so far we have tried to be reasonable about it (at least no breaking changes in patch versions). Like I said I am open to merging this as a fix if @richardwerkman agrees.

rouke-broersma:
@richardwerkman

Issue #1511, opened 2021-04-16
System.InvalidOperationException: Expected a block containing a conditional expression


bcollamore:
**Describe the bug**
Stryker fails

**Logs**
See attached

**Expected behavior**
Stryker succeeds

**Desktop (please complete the following information):**
 - OS: Windows
 - Type of project: Full Framework
 - Framework Version: net48
 - Stryker Version: 0.22.2

Here is the original method:
		private List<string> ExtractXmlMessages(string message, out string partialMessage)
		{
			List<string> xmlMessages = new List<string>();

			// While we think we have a valid message
			while (true)
			{
				// Remove any junk starting characters which will prevent
				// the ExtractXmlMessage method from working.
				if (StringHelper.Contains(message, XmlStartSymbol) && !StringHelper.StartsWith(message, XmlStartSymbol))
				{
					int numberOfJunkChars = message.IndexOf(XmlStartSymbol, StringComparison.OrdinalIgnoreCase);
					message = message.Remove(0, numberOfJunkChars);
				}

				string completeMessage = ExtractXmlMessage(message, out partialMessage);
				completeMessage = completeMessage.Trim();

				if (String.IsNullOrEmpty(completeMessage))
				{
					break;
				}

				xmlMessages.Add(completeMessage);
				message = partialMessage;
			}
			return xmlMessages;
		}

Here is the log output:
[12:17:38 DBG] Remove a mutation helper.
[12:17:38 ERR] An error occurred during the mutation test run 
System.InvalidOperationException: Expected a block containing a conditional expression, found:

		/// <summary>
		/// Extracts all complete xml strings from the given message string.
		/// </summary>
		/// <param name="message"></param>
		/// <param name="partialMessage"></param>
		/// <returns></returns>
		private List<string> ExtractXmlMessages(string message, out string partialMessage)
		{
			List<string> xmlMessages = new List<string>();

			// While we think we have a valid message
			while ((Stryker7YxyYnQrRbPl8fG.MutantControl.IsActive(776)?!(true):(Stryker7YxyYnQrRbPl8fG.MutantControl.IsActive(775)?false:true)))
			{
				// Remove any junk starting characters which will prevent
				// the ExtractXmlMessage method from working.
[...]

Issue #1527, opened 2021-04-20
Cannot inject mutation '{}' in '_collection' because we cannot find the original code.


marcin-golebiowski:
**Describe the bug**

Stryker.NET has a problem with mutating following code:

```
using System;
using System.Collections.Generic;
using System.Text;

namespace TestApp
{
    public static class ExampleExtension
    {
        private static string[] tabs = { "tab1", "tab2"};

        private List<string> _collection;

        public List<string> Collection
        {
            get => _collection;

            set
            {
                _collection = value;
            }
        }
    }
}
```

Message: 
    System.InvalidOperationException : Cannot inject mutation '{}' in '_collection' because we cannot find the original code.
  Stack Trace: 
    RoslynHelper.InjectMutation[T](T original, Mutation mutation) line 52
    <>c__DisplayClass1_0.<InjectMutations>b__0(Mutation toConvert) line 37
    <>c__DisplayClass1_0.<InjectMutations>b__2(Mutant m) line 44
    SelectEnumerableIterator`2.MoveNext()
    Enumerable.Aggregate[TSource,TAccumulate](IEnumerable`1 source, TAccumulate seed, Func`3 func)
    MutantPlacer.PlaceStatementControlledMutations(StatementSyntax original, IEnumerable`1 mutations) line 85
    AccessorSyntaxOrchestrator.InjectMutations(AccessorDeclarationSyntax sourceNode, SyntaxNode targetNode, MutationContext context) line 41
    NodeSpecificOrchestrator`2.Mutate(SyntaxNode node, MutationContext context) line 117
    CsharpMutantOrchestrator.Mutate(SyntaxNode currentNode, MutationContext context) line 108
    <>c__DisplayClass11_0.<OrchestrateChildrenMutation>b__0(SyntaxNode original, SyntaxNode _) line 93
    Replacer`1.Visit(SyntaxNode node)
    CSharpSyntaxRewriter.VisitListElement[TNode](TNode node)
    CSharpSyntaxRewriter.VisitList[TNode](SyntaxList`1 list)
    CSharpSyntaxRewriter.VisitAccessorList(AccessorListSyntax node)
    AccessorListSyntax.Accept[TResult](CSharpSyntaxVisitor`1 visitor)
    CSharpSyntaxRewriter.Visit(SyntaxNode node)
    Replacer`1.Visit(SyntaxNode node)
[...]

Issue #1512, opened 2021-04-16
System.InvalidOperationException: No return at the end of


bcollamore:
**Describe the bug**
Stryker failed

**Logs**
Logs attached.  Notes below.

**Expected behavior**
Stryker succeeds

**Desktop (please complete the following information):**
 - OS: Linux
 - Type of project: .NET Standard
 - Framework Version: Standard 1.3
 - Stryker Version: 0.22.2

**Additional context**

Version 0.21.1 reported "this should not happen" but we missed it in the CI pipeline.

[strykerlogs.txt](https://github.com/stryker-mutator/stryker-net/files/6326685/strykerlogs.txt)
Original method:
		public bool AddRelationship(T dependent, T dependsOn, out string cyclePath)
		{
			cyclePath = string.Empty;
			lock (Graph)
			{
				// If either dependent or dependee is new to the graph, a cycle isn't possible.
				bool isCyclePossible = !_graphAdder.Add(Graph, dependent, out Node<T> dependentNode);
				isCyclePossible |= !_graphAdder.Add(Graph, dependsOn, out Node<T> dependsOnNode);

				bool cycleFound = false;
				var answerKey = new List<Node<T>>();
				if (isCyclePossible)
				{
					cycleFound = DoesPathExist(start: dependsOnNode, end: dependentNode, visited: new List<Node<T>>(), answerKey: answerKey);
				}

				if (cycleFound)
				{
					answerKey.Add(dependsOnNode);
					cyclePath = ToString(answerKey);
					return false;
				}
				dependentNode.AddDependency(dependsOnNode);
				return true;
			}
		}

Error:
2021-04-16T16:37:27.7609948Z [12:37:27 DBG] Remove a mutation helper.
2021-04-16T16:37:27.7758873Z [12:37:27 ERR] An error occurred during the mutation test run 
2021-04-16T16:37:27.7760014Z System.InvalidOperationException: No return at the end of: {
2021-04-16T16:37:27.7761564Z 			cyclePath = (StrykerJtpdbb49QO1lapm.MutantControl.IsActive(292)?"Stryker was here!":string.Empty);
2021-04-16T16:37:27.7762715Z 			lock (Graph)
2021-04-16T16:37:27.7763087Z {			{
2021-04-16T16:37:27.7764314Z 				// If either dependent or dependee is new to the graph, a cycle isn't possible.
[...]

PR #1529, opened 2021-04-22
Fix rollback of helpers and injections of array initializers in the following method

dupdob:
Include several fixes and one improvement:

1. Ensure helpers are properly rollbacked and add unit test to prevent regression (issues #1511, #1512).
1. Stryker drops any pending mutants during mutation orchestrating when moving to the next member definition (issue #1527)
1. Stryker injects initialization to default value for all out parameters (at the beginning of methods). This should reduce compile error mutants (should improve #1511 and #1512)

Also refactored injection engine testing, fixed some error messages and improved some naming.
Fixes issues: #1511, #1512, #1527

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
+++ b/src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
@@ -94,21 +94,16 @@ namespace Stryker.Core.Compiling
 
         private int? ExtractMutationIfAndId(SyntaxNode node)
         {
-            var first = node.GetAnnotations(MutantPlacer.MutationMarkers).FirstOrDefault();
-            if (first == null)
+            var (engine, id) = MutantPlacer.FindEngine(node);
+
+            if (engine == null)
             {
                 return null;
             }
 
-            if (first.Data == null)
-            {
-                Logger.LogDebug("Remove a mutation helper.");
-                return _dummyId;
-            }
-            var mutantId = int.Parse(first.Data);
-            Logger.LogDebug("Found id {0} in MutantIf annotation", mutantId);
-            return mutantId;
+            Logger.LogDebug(id == -1 ? $"Found a helper: {engine}." : $"Found id {id} in {engine} annotation.");
 
+            return id;
         }
 
         private static SyntaxNode FindEnclosingMember(SyntaxNode node)
@@ -208,7 +203,7 @@ namespace Stryker.Core.Compiling
                         if (!brokenMutations.Contains(key))
                         {
                             brokenMutations.Add(key);
-                            if (value != _dummyId)
+                            if (value != -1)
                             {
                                 RolledBackIds.Add(value);
                             }
--- a/src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
+++ b/src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
@@ -24,7 +24,6 @@ namespace Stryker.Core.Compiling
     {
         private List<int> RolledBackIds { get; }
         private ILogger Logger { get; }
-        private const int _dummyId = int.MinValue;
 
         public RollbackProcess()
         {
@@ -89,7 +88,7 @@ namespace Stryker.Core.Compiling
                 }
             }
 
-            return (null, 0);
+            return (null, -1);
         }
 
         private int? ExtractMutationIfAndId(SyntaxNode node)
--- a/src/Stryker.Core/Stryker.Core/Instrumentation/DefaultInitializationEngine.cs
+++ b/src/Stryker.Core/Stryker.Core/Instrumentation/DefaultInitializationEngine.cs
@@ -1,0 +1,65 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Stryker.Core.Helpers;
+
+namespace Stryker.Core.Instrumentation
+{
+    internal class DefaultInitializationEngine : BaseEngine<BaseMethodDeclarationSyntax>
+    {
+        private static SyntaxAnnotation blockMarker = new SyntaxAnnotation("InitializationBlock");
+        public DefaultInitializationEngine(string injector): base(injector)
+        {
+        }
+
+        protected override SyntaxNode Revert(BaseMethodDeclarationSyntax node)
+        {
+            if (node.Body == null || node.Body.Statements.Count == 0 || node.Body.Statements[0].Kind() != SyntaxKind.Block)
+            {
+                throw new InvalidOperationException(
+                    "Can't find initializer block at the beginning of method.");
+            }
+
+            return node.WithBody(
+                    node.Body.WithStatements(new SyntaxList<StatementSyntax>(node.Body.Statements.Skip(1))))
+                .WithoutAnnotations(Marker);
+        }
+
+        public BaseMethodDeclarationSyntax AddDefaultInitializer(BaseMethodDeclarationSyntax node, SyntaxToken outParameterParameterName, TypeSyntax outParameterParameterType)
+        {
+            if (node.Body == null)
+            {
+                throw new InvalidOperationException(
+                    "Cant' add default initializer(s) to expression bodied or virtual method.");
+            }
+
+            IEnumerable<StatementSyntax> initializers = null;
+            IEnumerable<StatementSyntax> originalStatements = null;
+            if (node.Body.Statements.Count > 0 && node.Body.Statements[0].HasAnnotation(blockMarker))
+            {
+                // we add a new initializer, we need to get the existing ones
+                initializers = (node.Body.Statements[0] as BlockSyntax).Statements;
+                // we can skip the initializer helper block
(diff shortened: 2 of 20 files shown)


src/Stryker.Core/Stryker.Core/Mutants/NodeOrchestrators/MemberDeclarationOrchestrator.cs
```diff
@@ -0,0 +1,18 @@
...
+    internal class MemberDeclarationOrchestrator<T, TBase> : NodeSpecificOrchestrator<T, TBase> where T : TBase where TBase : MemberDeclarationSyntax
+    {
+        public MemberDeclarationOrchestrator(CsharpMutantOrchestrator mutantOrchestrator) : base(mutantOrchestrator)
+        {
+        }
+
+        protected override TBase OrchestrateChildrenMutation(T node, MutationContext context)
+        {
+            var result = base.OrchestrateChildrenMutation(node, context);
+            context.Discard();
```
rouke-broersma, line 14: Could you add why the context needs to be discarded here?

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
```diff
@@ -94,21 +94,16 @@ public RollbackProcessResult Start(CSharpCompilation compiler, ImmutableArray<Di
...
 
-            if (first.Data == null)
-            {
-                Logger.LogDebug("Remove a mutation helper.");
-                return _dummyId;
-            }
-            var mutantId = int.Parse(first.Data);
-            Logger.LogDebug("Found id {0} in MutantIf annotation", mutantId);
-            return mutantId;
+            Logger.LogDebug(id == -1 ? $"Found a helper: {engine}." : $"Found id {id} in {engine} annotation.");
```
rouke-broersma, line 104:
```suggestion
            Logger.LogDebug(id == -1 ? $"Found a mutation engine: {engine}." : $"Found id {id} in {engine} annotation.");
```

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
```diff
@@ -94,21 +94,16 @@ public RollbackProcessResult Start(CSharpCompilation compiler, ImmutableArray<Di
...
+
+            if (engine == null)
             {
                 return null;
             }
 
-            if (first.Data == null)
-            {
-                Logger.LogDebug("Remove a mutation helper.");
-                return _dummyId;
```
rouke-broersma, line 106: If we no longer need the dummyId we should remove the declaration

src/Stryker.Core/Stryker.Core/Mutants/MutationContext.cs
```diff
@@ -47,10 +56,28 @@ public MutationContext Clone()
             return new MutationContext(this);
         }
 
+        public void Discard()
+        {
+            if (HasStatementLevelMutant)
+            {
+                // some mutants 
+                Logger.LogInformation($"{BlockLevelControlledMutations.Count+StatementLevelControlledMutations.Count} mutations were not injected.");
```
rouke-broersma, line 64: It seems like this might be logged often. Is this important for the user to know about? If so there should be some clarification on what the user should do with this information. If not this should at a minimum be LogDebug.

src/Stryker.Core/Stryker.Core/Mutants/NodeOrchestrators/StaticFieldDeclarationDeclarationOrchestrator.cs
```diff
@@ -1,10 +1,10 @@
-﻿using System.Linq;
+using System.Linq;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 
 namespace Stryker.Core.Mutants.NodeOrchestrators
 {
-    internal class StaticFieldDeclarationOrchestrator: NodeSpecificOrchestrator<FieldDeclarationSyntax, BaseFieldDeclarationSyntax>
+    internal class StaticFieldDeclarationDeclarationOrchestrator: MemberDeclarationOrchestrator<FieldDeclarationSyntax, BaseFieldDeclarationSyntax>
```
rouke-broersma, line 7: This name 😅😂

src/Stryker.Core/Stryker.Core/Instrumentation/DefaultInitializationEngine.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Stryker.Core.Helpers;
+
+namespace Stryker.Core.Instrumentation
+{
+    internal class DefaultInitializationEngine : BaseEngine<BaseMethodDeclarationSyntax>
```
rouke-broersma, line 11: This engine can be used to place default initializers on method scope variables, correct? Could you add a small docstring to this class explaining the intent.

rouke-broersma:
@dupdob I see you opened this PR twice, once to master and once to v1. Did you intend to keep the v1 PR? Because I think this fix can go to master.

src/Stryker.Core/Stryker.Core/Mutants/NodeOrchestrators/MemberDeclarationOrchestrator.cs
dupdob: done

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
dupdob: I am not confortable with your proposal, as `engine` is not a _mutation_ engine, but more of an `injection`engine. Just to clarify any potential misunderstanding:
- if `id` is positive or null, this is a mutation, and the engine will describe the control logic: `IfInstrumentationEngine` (`if (Stryker.MutantIsActive(1)){...}else{...}`) or `ConditionalInstrumentationEngine` (`Stryker.MutantIsActive(1)?...:...`).
- if ' `id` is `-1` (magic number ☹️), this is some other helping construction, such `return default(...)` or an expression to body conversion, or out parameters initialization.... Definitely not a **mutation**, more a minor refactoring that helps mutation to succeed

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
dupdob: done, sorry about that one

src/Stryker.Core/Stryker.Core/Mutants/NodeOrchestrators/StaticFieldDeclarationDeclarationOrchestrator.cs
dupdob: oups, something went wrong here. 🤣 .

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
rouke-broersma: I'm fine with removing mutation from the logrule. I mainly meant to place engine instead of helper because we definitely know it's an engine.

src/Stryker.Core/Stryker.Core/Instrumentation/DefaultInitializationEngine.cs
dupdob: yes. I added documentation to every engine for clarification.

rouke-broersma: Awesome :)

src/Stryker.Core/Stryker.Core/Mutants/MutationContext.cs
dupdob: This should not happen often. Currently, AFAIK it only happens with array initializers in fields, which result in mutations that cannot be injected.
I will put it at debug level for the time being.

dupdob:
I will push a new version that adresses comment here and reopen a PR for master. You will have to pick only one.

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
dupdob: well that's an engine that injects and remove helpers :-)

dupdob: Anyway, I improved the error message on the 'master' PR, assuming it will be the one you pick. I also fixed another mistake: https://github.com/stryker-mutator/stryker-net/pull/1529/files#diff-edba80c123cba4b1cd940aeec21549583d3abeb379d33e2ea3545516649925baR127

src/Stryker.Core/Stryker.Core/Mutants/MutationContext.cs
dupdob: see PR #1530.
https://github.com/stryker-mutator/stryker-net/pull/1530/files#diff-41b1bf6c0a66c50dad49e7ad055c28211a76f2e663cd9bcb9fcacc1477e98b0fR64

PR #1538, opened 2021-04-27
split xunit and xunit.runner.visualstudio packages

SimonCropp:
unfortunately the xunit and xunit.runner.visualstudio packages are not kept in version sync. so there is no value in using a shared version property for them


rouke-broersma:
Is it only the patch versions or also it minor versions that are not kept in sync? Otherwise we might use a wildcard for the patch version?

Issue #1546, opened 2021-04-30
Options to include or exclude LinqExpressions on LinqMutator


leonardosimoura:
**Is your feature request related to a problem? Please describe.**
Options to include or exclude LinqExpressions on LinqMutator. Ex. Not run FirstOrDefault mutations

**Describe the solution you'd like**
A additional parameter to configure

**Describe alternatives you've considered**
I tried the mutate example !MyFolder/MyService.cs{10..100} but it ignore logic inside some linq methods ( example ( First(f => f.Amount > 10)  where f.Amount > 10 is ignored to

**Additional context**
I trying to run stryker againt a big project and if we can choose when to enabled some of linq mutations will help us to focus on the core logic.

richardwerkman:
I'm curious why you want to exclude FirstOrDefault mutations. If you think the mutation is irrelevant we could make this issue about that. It is a very hard mutation to kill so I do understand your issue with it.

I was thinking about moving that mutation to a higher mutation level. So that some linq mutations will be turned on by default and some only if you increase the mutation level.

leonardosimoura:
The ideia is just disable them temporarily so is possible to focus on other mutations first.
Think in a project with half of its mutations as Linq mutations, but for the bussiness team  the most impotant mutations are others mutations and only a few linq mutations.

Sorry but i made the PR #1660  before read this,
I like the ideia to filter the expressions based on mutation level.
I created a ExcludeLinqExpressionFilter is possible to use it to apply some rules
What you think about something like:
**Basic:** Any, Count,
**Standard:**  Min, Max, Sum, Average,
**Advanced:**  OrderBy, OrderByDescending, FirstOrDefault, First, SingleOrDefault, Single, Last, All, Skip, Take, SkipWhile, TakeWhile, ThenBy, ThenByDescending,
**Complete:** All Expressions

richardwerkman:
What about?
Basic: Any, Count, All
Standard: Min, Max, Sum, Average, OrderBy, OrderByDescending, First, Single, Last,
Advanced: Skip, Take, SkipWhile, TakeWhile, ThenBy, ThenByDescending,
Complete: All Expressions

I think order of items is very important to test. So it should be enabled standard. This means OrderBy but also First and Last as they are ordering as well. But they won't be mutated to FirstOrDefault. We can do that in complete.

rouke-broersma:
Perhaps this should be a separate issue

leonardosimoura:
[...]

PR #1273, opened 2020-10-23
feat: Rework how stryker consumes options

rouke-broersma:
Fixes #578 

### Overview of options migration progress:
https://docs.google.com/spreadsheets/d/1fyKeqWhROup1yxZLqeQ5OV5bOSV5hej1eNKiH7lVRBU/edit?usp=sharing

- [x] Add unit tests
- [ ] Check docs
- [ ] Review comments
- [ ] Fix integration test and stryker stryker-config files
- [ ] Manual test all options
- [ ] Prepare migration guide
- [ ] Prepare blog post for migration guide

PR #1547, opened 2021-04-30
Options to include or exclude LinqExpressions on LinqMutator

leonardosimoura:
PR for #1546

### DIFF
--- a/src/Stryker.CLI/Stryker.CLI.UnitTest/ConfigFileTests.cs
+++ b/src/Stryker.CLI/Stryker.CLI.UnitTest/ConfigFileTests.cs
@@ -1,4 +1,4 @@
-﻿using DotNet.Globbing;
+using DotNet.Globbing;
 using Microsoft.CodeAnalysis.Text;
 using Moq;
 using Serilog.Events;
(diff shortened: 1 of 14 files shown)


rouke-broersma:
Hi @leonardosimoura 👋 

We are currently reworking all options in #1273 
I don't think we will want an option like this to be a commandline option (so config file only) in the new design. We will take a look at your pull request when we are done reworking all the options infrastructure. We will implement the needed changes to the options at that time. Hope that's alright with you.

leonardosimoura:
Hi @rouke-broersma 

It's OK,  i will follow the #1273 for new updates
Thanks for the attention

richardwerkman:
@leonardosimoura just a heads up, #1273 has been merged so you can update your PR with our new options.

I do have a note though. I'd rather have this not a new option. We could add it to the current option like this:
``` json
{
  "stryker-config": {
    "ignore-mutations": [
      "string",
      "linq.FirstOrDefault"
    ]
}
```
this option is already config file only like @rouke-broersma described. 

And some other remarks:
We have a mechanism to filter mutations without changing the mutator like you did. In the folder `MutationFilters` you can find some [examples](https://github.com/stryker-mutator/stryker-net/blob/v1.0/src/Stryker.Core/Stryker.Core/MutantFilters/ExcludeMutationMutantFilter.cs). If you need help working with the new options or the mutant filters please let us know. We are happy to help.

richardwerkman:
I've created #1656 for some guidance.

leonardosimoura:
Hi @richardwerkman  thanks for attention.
I will abandon this PR, refactor the solution and open a new PR.

PR #1550, opened 2021-05-02
Feature: Added --filter option to VsTestConsoleWrapper

corentinaltepe:
This PR is a work in progress and is clearly not ready for review. However, I need help. I am stuck with the problem described below.

Contributes to #420.
This PR depends on #1542 and #1273 which will need both to be merged first. Then this PR will be updated before ready for final review.

### Progress
As suggested by @rouke-broersma:
- [ ] user passes filter to stryker as an option according to new standard: #1273 (wait for the PR to be merged).
- [ ] filter gets used to filter the discovered tests in the initial testrun
- [ ] filter gets used to filter the executed tests during coverage analysis (all modes), this effectively filters the used tests
- [ ] if coverage analysis is turned off, the filter still needs to be applied because coverage analysis did not perform the filtering for us

### Problem
tl;dr: vstest.console.exe applied `/TestCaseFilter:` option works just fine, but the `VsTestConsoleWrapper` does not.

@dupdob: since you so kindly offered support in #420, may I ask your help here. I'm kinda stuck on this issue and running out of ideas.

I re-used a `xUnit` test project `Utils.Test.csproj` containing two dummy tests:
- `Utils.Test.MathTest.Ceiling` with category `unit`
- `Utils.Test.MathTest.Floor` with category `integration`

I managed to properly list and filter them with `dotnet test` and `vstest.console.exe` using either `--filter Category=unit` or `/TestCaseFiler:Category=unit` options.

```bash
"C:\Users\Corentin\.nuget\packages\microsoft.testplatform.portable\16.9.4\tools\net451\vstest.console.exe" -lt /TestCaseFilter:Category=unit C:\Users\Corentin\source\repos\ConteneurWeb\Tests\Common\Utils.Test\bin\Debug\net5.0\Utils.Test.dll
Microsoft (R) Test Execution Command Line Tool Version 16.9.4
Copyright (c) Microsoft Corporation.  All rights reserved.

The following Tests are available:
[xUnit.net 00:00:00.00] xUnit.net VSTest Adapter v2.4.3+1b45f5407b (64-bit .NET 5.0.5)
[xUnit.net 00:00:00.33]   Discovering: Utils.Test
[xUnit.net 00:00:00.38]   Discovered:  Utils.Test
    Utils.Test.MathTest.Ceiling
```

[...]

PR #1542, opened 2021-04-28
Refactor test and coverage logic

dupdob:
This PR improves on the existing test and coverage logic on several front:

- [x] **Drastically reduce the memory pressure**  during testing phase thanks to the suppression of several wrapping classes. This results in a slight performance increase for large projects
- [x] **Improve support for dynamic test cases** (i.e. using `TestCaseSource` attribute) thanks to extensive testing of xUnit and NUnit behaviour. This should fixes several issues related to those as well as increase Stryker accuracy and efficiency.
- [x] **Improve support for test cases providing multiple results**. This should fix potential bus with those (mostly false negative, i.e. mutants flagged as survivor while they should be killed by some test).
- [x] **Improved timeout computation** based on the list of tests to executes (vs reusing the default test time as before). The current logic is: fix/initialization time + sum of the run time of the tests (as established during initial test run), which then goes through the computation logic. the fix/initialization time is estimated by subtracting the sum of test times to the initial run time. 
- [x] **tolerate failing tests** in the initial test run (i.e. with the non mutated assembly).

see #1476 
Other changes and details
- remove `DotNetTestRunner` code
- remove named pipes and related classes
- tests failing at initial run:
  -  Stryker now tolerates up to 10% of failed tests during initial test run. This is an arbitrary decision that can easily be changed, but I felt having a threshold was sensible
  - _'failing tests'_ are still run during mutation tests and **considered as successful**
  - corollary : a mutant covered only by _'failing tests'_ is flagged as _Survived_. It could have been flagged as 'No Coverage' but I felt user will be surprised by this
- I refactored the backup/restore of the assembly so that the same class (`ProjectInfo`) was in charge of both backup AND restore
- the `MutationInput` class now hosts an `ITimeoutValueCalculator' and the results of the initial Testrun. Several method signatures, interfaces and test have been refactored as a consequence
- the `Mutant` class no longer stores complete description of the covering tests, only the test ids, for efficiency reason. As a consequence, I had to slight change the MutantFilter class so the `DiffMutantFilter` still support changed test source files
- several tests have been added to cover testcases with multiple test results
- Stryker now detects when mutation are executed outside normal test run (usually before the test case starts). Those mutations are marked to be run against all tests as a precaution; we may adopt a finer strategy at later time
- Stryker's data collector set the correct mutant ID when each test starts, except when there is only one mutant to test in the while session. Then the mutation ID is set at the beginning of the test session. This is required for mutation appearing outside tests
[...]

Issue #420, opened 2019-03-22
Feature request: Support for test filters


kwlin:
It would be nice to provide a way which tests should be executed within stryker-mutator like: https://docs.microsoft.com/en-us/dotnet/core/testing/selective-unit-tests 

Just to prevent (integration/long running)tests will be executed during a stryker run.

richardwerkman:
This could potentially conflict with our new feature #388. We have to use the filter there too. If we have a way of combining the users filter with our own filters this would be a nice improvement!

rubberduck203:
We heavily use test categories and filters to specify which tests can be run without external hardware connected.
Without a way to filter out tests that won't pass without that hardware, we can't really leverage this great tool.
Is there another option for filtering?

richardwerkman:
@rubberduck203 I guess there is a way. When you enable coverage analyse `dotnet stryker --coverage-analysis perTest` stryker automatically filters out test cases that are not hit by any tests.

So if you enable test coverage you should be able to run stryker on your project! You might have to exclude some files to make sure that the hardware tests won't run.

jnyrup:
It would be nice to have this filtering performed before coverage analysis.
Especially when  coverage analysis cannot filter out tests due to
```
WRN: Each mutant will be tested against XXX because we can't get coverage info for test case generated at run time
```

I assume this could both speed up the initial test run and the mutation test runs.

Mobrockers:
We are open to pull requests to enable custom test case filtering as long as this works alongside coverage analysis.

corentinaltepe:
I would be very interested in implementing such feature. I'm unfamiliar with the code base. Would anyone have pointers to how to get started, which classes the option is likely to be added, what implementation you think would work? Thanks.

rouke-broersma:
> I would be very interested in implementing such feature. I'm unfamiliar with the code base. Would anyone have pointers to how to get started, which classes the option is likely to be added, what implementation you think would work? Thanks.

We execute the tests using vstest here: https://github.com/stryker-mutator/stryker-net/blob/v1.0/src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs

This class also does test discovery. It's been a while since I touched this but I think the flow would generally be as follows:

1. user passes filter to stryker as an option according to new standard: https://github.com/stryker-mutator/stryker-net/pull/1273
2. filter gets used to filter the discovered tests in the initial testrun
[...]

PR #1557, opened 2021-05-09
Feature: Added --filter option to VsTestConsoleWrapper

corentinaltepe:
This PR is a work in progress.

Contributes to #420.

This PR depends on #1542 and #1273 which will need both to be merged first. Then this PR should be updated before ready for final review. It replaces #1550 by branching off v1.0 for an easier review of the changes.

### Progress
- [ ] user passes filter to stryker as an option according to new standard: #1273 (wait for the PR to be merged). In the meantime, I hardcoded a dummy test case fitler as `Category=unit` to fit my use case.
- [X] filter gets used to filter the discovered tests in the initial testrun
- [X] filter gets used to filter the executed tests during coverage analysis (all modes), this effectively filters the used tests
- [ ] if coverage analysis is turned off, the filter still needs to be applied because coverage analysis did not perform the filtering for us

### Comments
Your feedback is welcome. I have made other unrelated, opinionated changes that I would gladly reverse if they don't fit here: 
- allowed overwriting `.stryker-unchanged` file when moving it. The file is created when running stryker-cli but not removed if the CLI crashes, preventing the next run of the CLI if overwriting is not allowed.
- line returns on Linq expressions for better readability

VsTestConsole behavior around `TestCaseFilter` option is... strange. Passing in a `new TestPlatformOptions { TestCaseFilter = _options.TestCaseFilter }` argument to vsTestConsole when discovering tests has no effect. We must pass the option as a runsettings XML configuration, as suggested by @rouke-broersma. But the runsettings XML config has no effect when running the tests, we must then use the `TestPlatformOptions`.

Please find attached the logs (trace level) of a run of Stryker in a very small test project (2 tests, with a different category, a single test matches the Category=unit option). 
[logs.zip](https://github.com/stryker-mutator/stryker-net/files/6447293/logs.zip)

### What's Next

Where do you suggest we write tests for that option? Should we wait for #1273 to be complete first?
Are there cases I forgot to cover? I only ran stryker on my limited test project, on xUnit.

Thank you!

### DIFF
--- a/src/Stryker.Core/Stryker.Core/MutationTest/CsharpMutationProcess.cs
+++ b/src/Stryker.Core/Stryker.Core/MutationTest/CsharpMutationProcess.cs
@@ -11,7 +11,6 @@ using Stryker.Core.MutantFilters;
 using Stryker.Core.Mutants;
 using Stryker.Core.Options;
 using Stryker.Core.ProjectComponents;
-using Stryker.Core.Reporters;
 
 namespace Stryker.Core.MutationTest
 {
@@ -84,9 +83,9 @@ namespace Stryker.Core.MutationTest
                 }
                 if (_fileSystem.File.Exists(injectionPath))
                 {
-                    _fileSystem.File.Move(injectionPath, injectionPath + ".stryker-unchanged");
+                    _fileSystem.File.Move(injectionPath, injectionPath + ".stryker-unchanged", overwrite: true);
                 }
-                
+
                 // inject the mutated Assembly into the test project
                 using var fs = _fileSystem.File.Create(injectionPath);
                 ms.Position = 0;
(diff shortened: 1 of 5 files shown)


rouke-broersma:
@corentinaltepe so sorry this took so long! If you are still around and interested to finish this feature it would be much appreciated! Otherwise let us know and we will try to finish this PR for you.

corentinaltepe:
@rouke-broersma not a problem. Thank you for getting back to this PR. I'm sorry but I won't be available in the next few weeks. Then, I could start back where I left off. In the meantime, I'm totally okay if you decide to carry on. 

My understanding is this change had dependencies on major refactoring of the options, prior to releasing v1. I'll have to check the status on that before resuming. I might need help designing appropriate unit/integration tests. I'll get in touch in slack. Thank you!

corentinaltepe:
Closing in favor of #1669.

Issue #1570, opened 2021-05-21
Offer source level fine grained control of mutations and mutators


dupdob:
**Is your feature request related to a problem? Please describe.**
Controlling mutations and mutators within the tested project source code for fine grained control.
Proposed syntax looks like this
```csharp
// Stryker disable once all : Mutation has been reviewed and cannot be killed
```

Somer users want to keep their mutation coverage at 100%. But Stryker.Net may generates mutation that cannot be killed due to some external dependencies or (user project) internal design choices.
Users looking for 100% coverage want to be able to review those mutations and somehow mark them as to be ignored.
From a practical perspective, the best place to flag those mutations as to be ignored is within the source code itself, using 
single line comments to give instructions

**Describe the solution you'd like**
I also like the syntax. But it is somewhat different from the JS proposal (`next` which does not really makes to me).
Assuming this difference an issue, I think I could implement this, here are the envisioned specs:

- [ ] works on whole statements at minimum: this is due to how Roslyn attaches comments and code. Will do finer grained control (e.g. expression) if possible, but not a priority
- [ ] works on unitary/automatic scope as a priority, manually defined scope (i.e using `disable` and `restore`) is a secondary objective. This derives from the assumption this feature should be used sparingly and not in broad strokes.
- [ ] mutations will still be generated but will be flagged as ignored. First implementations may simply skip generating mutations. Remark: skipping mutating altogether may provide a workaround for when a mutator triggers an exception blocking all usage of Stryker  ⇒ we need to choose which behavior we want.
- [ ] works at statement, block or declaration level. The rule then apply to the attached scope. eg
```csharp
{
  // Stryker disable once all
  int x = 2+2;
  int y = x+3;
}
```
 ⇒ disable mutation for `int x = 2+2;` 
 ```csharp
// Stryker disable once all
{
  int x = 2+2;
  int y = x+3;
}
```
 ⇒ disable mutation for both lines
 - [ ] manual defined scope (i.e using `disable` and `restore`) does not work across scope (i.e. you cannot disable mutations in one method and reenable it somewhere in the next method). This is due to the tree parsing logic we use which makes this needlessly difficult to implement; plus it should be used sparingly
 - [ ] full comment syntax should be
    `// Stryker [disable[| once]|restore] [all|mutator1[|,mutator2...]][| : reason for disabling]`
[...]

PR #1583, opened 2021-05-30
Add fine grained control for mutators at source level.

dupdob:
Implements #1570.

Allow users to disable specific mutators for specific source lines/scopes thanks via source comments.
Impact mutants are still generated but flagged as ignored.
The general comment form is:
```// Stryker [disable|restore][once][all| mutator list][: reason for disabling]``` where
* **disable**: disable some or all mutators. This action is valid until:
  * end of scope (i.e. closing brackets
  * explicit _enable_comment
* **restore**: re enable some or all previously disabled mutators. Note: 
  * this does not override overall Stryker settings, one cannot enable a mutator disabled through general settings. 
  * no error is generated if one enables a mutator not disabled by an earlier comment
*  **once**: the action is limited to the following statement or expression. Valid both disable or enable
* **all**: action is for every mutator
* **mutator list**: comma separated list of mutators. you must use the mutator name, not is description. E.g.: ```arithmetic, statement```
*  **reason for disabling**: comment that will used as status detail for why the mutation has been ignored.

Also, this PR implement the needed refactoring for parsing classes: mutation scope and injection management is consistent with parsing. As such there is no longer the risk of having a mutation 'leaking' to next method/property as before. This allowed significant simplification of existing orchestrators and the removal of the `ForStatementOrchestrator`. As a side effect, mutation numbering may be altered.

### CODE REVIEW

File: src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs [hunk 1]
```diff
@@ -9,10 +9,14 @@ namespace Stryker.Core.Helpers
...
-        // gets the return type of a method (incl. constructor, destructor...)
-        public static TypeSyntax ReturnType(this BaseMethodDeclarationSyntax baseMethod)
-        {
-            return baseMethod switch
+
+        /// <summary>
+        /// gets the return type of a method (incl. constructor, destructor...)
+        /// </summary>
+        /// <param name="baseMethod">method to analyze</param>
+        /// <returns>method returns type. Null if this not rele</returns>
```
psfinaki, line 17:
```suggestion
        /// <returns>method returns type. Null if this not relevant</returns>
```

src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
```diff
@@ -9,10 +9,14 @@ namespace Stryker.Core.Helpers
 {
     internal static class RoslynHelper
     {
-        // gets the return type of a method (incl. constructor, destructor...)
-        public static TypeSyntax ReturnType(this BaseMethodDeclarationSyntax baseMethod)
-        {
-            return baseMethod switch
+
+        /// <summary>
+        /// gets the return type of a method (incl. constructor, destructor...)
```
psfinaki, line 14:
```suggestion
        /// Gets the return type of a method (incl. constructor, destructor...)
```

src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
```diff
@@ -21,31 +25,44 @@ public static TypeSyntax ReturnType(this BaseMethodDeclarationSyntax baseMethod)
...
-        }
+        /// <summary>
+        /// Checks if an accessor is a getter and needs to end with a return
+        /// </summary>
+        /// <param name="baseMethod"></param>
+        /// <returns>true if this is a getter</returns>
+        public static bool NeedsReturn(this AccessorDeclarationSyntax baseMethod) =>
+            baseMethod.Keyword.Text == "get";
 
+        /// <summary>
```
psfinaki, line 59: Not the most helpful doc

src/Stryker.Core/Stryker.Core/Mutants/NodeOrchestrators/ExpressionSpecificOrchestrator.cs
```diff
@@ -1,45 +1,27 @@
...
-using System.Linq;
+using System.Collections.Generic;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Stryker.Core.Helpers;
 
 namespace Stryker.Core.Mutants.NodeOrchestrators
 {
     /// <summary>
-    /// Handles expressions and sub expressions.
+    /// Orchestrate mutations for expressions (and sub expressions-.
```
psfinaki, line 8:
```suggestion
    /// Orchestrate mutations for expressions (and sub expressions).
```

psfinaki (commented):
A few small suggestions below.

Also, more general points:
1) So is it enable or restore? The PR desc mentions the former, the code uses the latter. I like `restore` more.
2) Do I get it right that `reason for disabling` is optional? If so, maybe it shouldn't be?
3) I feel the whole `once` thing is a bit superfluous - it adds unnecessary complexity here and AFAIU the same behavior can be easily achieved with `disable`+`restore` pairs. But if you really want to keep it for some reason, please add related tests, I was unable to understand `once` usage from going through existing ones.

src/Stryker.Core/Stryker.Core/Mutants/NodeOrchestrators/ExpressionSpecificOrchestrator.cs
dupdob: done, thanks

src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
dupdob: Thanks. Finished documenting all helpers.

dupdob:
Thanks for your review comments. Quick answers to your questions:
1. 'restore' is the proper term, as per discussion #1503. Just a slip here, as I ave a bias for enable/disable
2. as for now, it is optional. There is an automatic comment provided if none is given. There is no technical obstacle making it mandatory, I suggest we use the issues to this discuss this matter with other to reach a consensus.
3. To be honest, _once_ is what I need, as I plan to disable some specific mutations. Using disable/restore pair is cumbersome just for one line. Stryker-js proposed a similar approach using _next_. And the added complexity is handled by a single conditional operator, it was _diable/restore_ that required significant effort 😀.

psfinaki:
Alright, thanks for the explanations. Makes sense, though now I don't completely understand why we need all this disable/restore thing, AFAIU we could get by with `disable once` for specific lines and `ignore-methods` options for blocks. But I guess that was probably already discussed somewhere :)

dupdob:
Hi, as far as I understand the situation, there is nothing left to discuss on this PR, or is there any topics that blocks its merge

rouke-broersma:
Hi @dupdob I haven't had time to review so that's why I haven't merged

src/Stryker.Core/Stryker.Core/Mutants/CsharpMutantOrchestrator.cs
```diff
@@ -108,12 +97,14 @@ public IEnumerable<Mutant> GenerateMutationsForNode(SyntaxNode current, Mutation
                     var id = MutantCount;
                     Logger.LogDebug("Mutant {0} created {1} -> {2} using {3}", id, mutation.OriginalNode,
                         mutation.ReplacementNode, mutator.GetType());
+                    var mutantIgnored = context.FilteredMutators!= null &&  context.FilteredMutators.Contains(mutation.Type);
```
rouke-broersma, line 100:
```suggestion
                    var mutantIgnored = context.FilteredMutators?.Contains(mutation.Type) ?? false;
```

src/Stryker.Core/Stryker.Core/Mutants/MutationContext.cs
```diff
@@ -30,59 +22,76 @@ public MutationContext(CsharpMutantOrchestrator mutantOrchestrator)
...
-                foreach (var mutant in BlockLevelControlledMutations.Union(StatementLevelControlledMutations))
-                {
-                    mutant.ResultStatus = MutantStatus.CompileError;
-                    mutant.ResultStatusReason = "Stryker was not able to inject mutation in code.";
-                }
-                BlockLevelControlledMutations.Clear();
-                StatementLevelControlledMutations.Clear();
+                result.FilteredMutators = filteredMutators;
+            }
+            else if (result.FilteredMutators != null)
```
rouke-broersma, line 72:
```suggestion
            else if (result.FilteredMutators is not null)
```

All these fancy new csharp9 features but no is not null pattern? ;)

src/Stryker.Core/Stryker.Core/Mutants/MutationStore.cs
```diff
@@ -0,0 +1,125 @@
...
+                        StoreMutations(mutations, MutationControl.Block);
+                        return;
+                    }
+                    _statementMutants.Peek().AddRange(mutations);
+                    break;
+                case MutationControl.Block:
+                    if (_blockMutants.Count == 0)
+                    {
+                        // we have mutations that are going to be lost
+                        _logger.LogInformation($"{mutations.Count()} were generated but could not be injected as they cannot be controlled dynamically.");
```
rouke-broersma, line 59: I don't think the user will understand what's happening here based on the message. Informational logging should imo be obvious or explicit about why the user should care about this. Does it cause inconsistency? Should the user care? Can the user do something about this?

src/Stryker.Core/Stryker.Core/Mutants/NodeOrchestrators/NodeSpecificOrchestrator.cs
```diff
@@ -89,35 +94,94 @@ protected NodeSpecificOrchestrator(CsharpMutantOrchestrator mutantOrchestrator)
...
+        protected virtual MutationContext PrepareContext(TNode node, MutationContext context)
+        {
+            foreach (var commentTrivia in node.GetLeadingTrivia().Where(t => t.IsKind(SyntaxKind.SingleLineCommentTrivia) || t.IsKind(SyntaxKind.MultiLineCommentTrivia)).Select(t => t.ToString()))
+            {
+                // perform a quick pattern check to see if it is a 'Stryker comment'
+                if (!Pattern.Match(commentTrivia).Success)
+                {
+                    continue;
+                }
+                var match = Parser.Match(commentTrivia);
```
rouke-broersma, line 111: I'm not very familiar with Regex performance. Does it actually save anything to first do a quick match before the full match?

src/Stryker.Core/Stryker.Core/Mutants/NodeOrchestrators/NodeSpecificOrchestrator.cs
```diff
@@ -89,35 +94,94 @@ protected NodeSpecificOrchestrator(CsharpMutantOrchestrator mutantOrchestrator)
...
+                if (match.Success)
+                {
+                    // this is a Stryker comments, now we parse it
+                    bool mode;
```
rouke-broersma, line 115: Mode does not feel like a bool value to me. Perhaps call it 'disable' or 'restore' or make it an Enum (we might have more modes in the future?)

src/Stryker.Core/Stryker.Core/Mutants/NodeOrchestrators/NodeSpecificOrchestrator.cs
```diff
@@ -89,35 +94,94 @@ protected NodeSpecificOrchestrator(CsharpMutantOrchestrator mutantOrchestrator)
...
+                    // get the ignore comment
+                    var comment = match.Groups[4].Value?.Trim();
+                    if (string.IsNullOrEmpty(comment))
+                    {
+                        comment = "Ignored via Stryker comment.";
+                    }
+                    switch (match.Groups[1].Value.ToLower())
```
rouke-broersma, line 122: Perhaps it would be good to name these magic numbers

```suggestion
const int modeKey = 1;
const int mutatorsKey = 3;
                    switch (match.Groups[modeKey].Value.ToLower())
```

src/Stryker.Core/Stryker.Core/Mutants/NodeOrchestrators/NodeSpecificOrchestrator.cs
```diff
@@ -89,35 +94,94 @@ protected NodeSpecificOrchestrator(CsharpMutantOrchestrator mutantOrchestrator)
...
+                        filteredMutators = new Mutator[labels.Length];
+                        for (var i = 0; i < labels.Length; i++)
+                        {
+                            if (Enum.TryParse<Mutator>(labels[i], true, out var value))
+                            {
+                                filteredMutators[i] = value;
+                            }
+                            else
+                            {
+                                _logger.LogWarning(
```
rouke-broersma, line 150: I think this can be an error logging

src/Stryker.Core/Stryker.Core/Mutants/NodeOrchestrators/NodeSpecificOrchestrator.cs
```diff
@@ -89,35 +94,94 @@ protected NodeSpecificOrchestrator(CsharpMutantOrchestrator mutantOrchestrator)
...
+                            }
+                        }
+                    }
+
+                    context = context.FilterMutators(mode, filteredMutators, match.Groups[2].Value.ToLower() == "once", comment);
+                    break;
+                }
+
+                _logger.LogWarning($"Invalid Stryker comments at {node.GetLocation().GetMappedLineSpan().StartLinePosition}, {node.SyntaxTree.FilePath}.");
+            }return context;
```
rouke-broersma, line 161:
```suggestion
            }
            
            return context;
```

src/Stryker.Core/Stryker.Core/Mutants/NodeOrchestrators/NodeSpecificOrchestrator.cs
```diff
@@ -89,35 +94,94 @@ protected NodeSpecificOrchestrator(CsharpMutantOrchestrator mutantOrchestrator)
...
+                var match = Parser.Match(commentTrivia);
+                if (match.Success)
+                {
+                    // this is a Stryker comments, now we parse it
+                    bool mode;
+                    // get the ignore comment
+                    var comment = match.Groups[4].Value?.Trim();
+                    if (string.IsNullOrEmpty(comment))
+                    {
+                        comment = "Ignored via Stryker comment.";
```
rouke-broersma, line 120:
```suggestion
                        comment = "Ignored via source code comment.";
```

rouke-broersma (approved):
Looks very good overall! Tested all options and all working. Only some small comments here and there.

src/Stryker.Core/Stryker.Core/Mutants/MutationStore.cs
dupdob: Agreed, changed to LogDebug

src/Stryker.Core/Stryker.Core/Mutants/NodeOrchestrators/NodeSpecificOrchestrator.cs
dupdob: This is not related to performance. This first regex is used to recognize special comments so we can raise an error to the user if there is an error in the comment syntax. 
Check could be made the other way around : check the complex first, then the simple one, but it would mean testing twice each regular comment.

src/Stryker.Core/Stryker.Core/Mutants/NodeOrchestrators/NodeSpecificOrchestrator.cs
dupdob: agreed, switched to `disable`

src/Stryker.Core/Stryker.Core/Mutants/NodeOrchestrators/NodeSpecificOrchestrator.cs
dupdob: done. Completed the list with the `commentGroup`

src/Stryker.Core/Stryker.Core/Mutants/NodeOrchestrators/NodeSpecificOrchestrator.cs
dupdob: done

Issue #1612, opened 2021-06-30
NullReference in Stryker.Core.Instrumentation.EndingReturnEngine.InjectReturn


petertiedemann:
**Describe the bug**
NullReference in Stryker.Core.Instrumentation.EndingReturnEngine.InjectReturn when generating mutants.

**Logs**
[logs.zip](https://github.com/stryker-mutator/stryker-net/files/6741819/logs.zip)

**Expected behavior**
Not crashing while generating mutants.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Type of project: netstandard2.0
 - Framework Version core 3.1
 - Stryker Version 0.22.5

**Additional context**
I am not entirely sure how to read the logs, but the last created mutant `[DBG] Mutant 696 created "value.Sign < 0" -> "value.Sign <= 0" using "Stryker.Core.Mutators.BinaryExpressionMutator" (310d25d3)`  seems to be a mutation of this method ( that i can share because it is from an open sourced BigRational impl):

```
    public static explicit operator Double( BigRational value ) {
      // The Double value type represents a double-precision 64-bit number with
      // values ranging from -1.79769313486232e308 to +1.79769313486232e308
      // values that do not fit into this range are returned as +/-Infinity
      if ( value.Numerator.InDoubleRange() && value.Denominator.InDoubleRange() ) {
        return (double)value.Numerator / (double)value.Denominator;
      }

      // scale the numerator to preseve the fraction part through the integer division
      BigInteger denormalized = ( value.Numerator * s_bnDoublePrecision ) / value.Denominator;
      if ( denormalized.IsZero )
        **return ( value.Sign < 0 ) ? BitConverter.Int64BitsToDouble( unchecked ( (long)0x8000000000000000 ) )
          : 0d; // underflow to -+0**

      // Note that it does not handle subnormals.
      // and +/- infinity detection might be done with " simple" comparison with their BigRational representation ?
      double result = 0;
      bool isDouble = false;
      int scale = DoubleMaxScale;

      while ( scale > 0 ) {
        if ( !isDouble ) {
          if ( denormalized.InDoubleRange() ) {
            result = (double)denormalized;
            isDouble = true;
          }
          else {
            denormalized = denormalized / 10;
          }
        }

        result = result / 10;
        scale--;
      }

      if ( !isDouble )
        return ( value.Sign < 0 ) ? Double.NegativeInfinity : Double.PositiveInfinity;
      else
        return result;
    }
```

[...]

PR #1613, opened 2021-06-30
Support conversion operator

dupdob:
Add proper support to conversion operator and fix issue #1612


richardwerkman:
Tested locally and checks out. Thanks for the quick fix!

PR #1614, opened 2021-07-01
Mark noCoverage mutation as ignored when not changed

### LINKED ISSUE #1586: Coverage analysis filter overwrites Ignored mutant statuses while it should ignore them
**Describe the bug**
Let's say I have a project with two identical classes (attached to this ticket). 
Initially (_initial commit, hash d242de36_) I have poor test coverage:

`dotnet stryker`
image

This makes me sad, so I do some refactoring and add a bunch of tests for the first class (_added some tests, hash b71bedf0_). Now I have better score:

`dotnet stryker`
image

Numbers make sense. But when I run Stryker in diff mode against _initial commit_ (in reality, that's in PRs), the stats are odd:

` dotnet stryker -diff -gdt d242de36580bc0e13f8ba3bd00065fb8e2ef0ca6`
image

**Expected behavior**
There are a few things wrong:
1) Ideally, I wouldn't want to see `Class2.cs` here at all, because it was not touched since diff target and Stryker apparently somehow takes this into account, _ignoring_ some mutants in that class (and therefore making 0 score there). 
2) Even if that must be there, where does the number 66.67 come from? AFAIU it should be 100% there, but I would understand 50% (incorrect but at least clearly 4/8 = 50%).

 - OS: Windows
 - Type of project: .NET Core
 - Framework Version: 5
 - Stryker Version: beta 0.22.4

**Additional context**
The code: [StrykerDiff.zip](https://github.com/stryker-mutator/stryker-net/files/6616253/StrykerDiff.zip)

dupdob:
Make sure DiffMutantFilter marks NoCoverage mutations as ignored when there is no change.
Fix #1586 #1620

### DIFF
--- a/src/Stryker.Core/Stryker.Core/MutantFilters/DiffMutantFilter.cs
+++ b/src/Stryker.Core/Stryker.Core/MutantFilters/DiffMutantFilter.cs
@@ -97,7 +97,7 @@ namespace Stryker.Core.MutantFilters
 
         private IEnumerable<Mutant> SetMutantStatusForFileChanged(IEnumerable<Mutant> mutants)
         {
-            foreach (var mutant in mutants)
+            foreach (var mutant in mutants.Where(m => m.ResultStatus != MutantStatus.NoCoverage))
             {
                 mutant.ResultStatus = MutantStatus.NotRun;
                 mutant.ResultStatusReason = "Mutant changed compared to target commit";
@@ -108,7 +108,7 @@ namespace Stryker.Core.MutantFilters
 
         private IEnumerable<Mutant> SetMutantStatusForNonCSharpFileChanged(IEnumerable<Mutant> mutants)
         {
-            foreach (var mutant in mutants)
+            foreach (var mutant in mutants.Where(m => m.ResultStatus != MutantStatus.NoCoverage))
             {
                 mutant.ResultStatus = MutantStatus.NotRun;
                 mutant.ResultStatusReason = "Non-CSharp files in test project were changed";
--- a/src/Stryker.Core/Stryker.Core/MutantFilters/DiffMutantFilter.cs
+++ b/src/Stryker.Core/Stryker.Core/MutantFilters/DiffMutantFilter.cs
@@ -86,7 +86,7 @@ namespace Stryker.Core.MutantFilters
 
         private IEnumerable<Mutant> SetNotRunMutantsToIgnored(IEnumerable<Mutant> mutants)
         {
-            foreach (var mutant in mutants.Where(m => m.ResultStatus == MutantStatus.NotRun))
+            foreach (var mutant in mutants.Where(m => m.ResultStatus == MutantStatus.NotRun || m.ResultStatus == MutantStatus.NoCoverage))
             {
                 mutant.ResultStatus = MutantStatus.Ignored;
                 mutant.ResultStatusReason = "Mutant not changed compared to target commit";
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -23,8 +23,8 @@ pr:
     - '.github/*'
   
 variables:
-  PackageVersion: '0.22.5'
-  VersionBuildNumber: $[counter('0.22.5', 1)]
+  PackageVersion: '0.22.6'
+  VersionBuildNumber: $[counter('0.22.6', 1)]
   IntegrationTestVersion: $(PackageVersion)-alpha.$(VersionBuildNumber)
   ProjectsToPack: src/Stryker.Core/Stryker.Core/Stryker.Core.csproj;src/Stryker.CLI/Stryker.CLI/Stryker.CLI.csproj
   MygetFeed: https://www.myget.org/F/stryker-mutator-integrationtest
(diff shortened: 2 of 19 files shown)


psfinaki (approved):

richardwerkman:
@psfinaki Have you tested if this actually solves your issue? I think it should work but would be nice to know for sure before this is merged.

psfinaki:
@richardwerkman hmm I try but I cannot reproduce the issue using the local codebase. Like AFAIU this should be identical:
image

But the CLI project seems to ignore diff arguments:
image

src/Stryker.Core/Stryker.Core/MutantFilters/DiffMutantFilter.cs
```diff
@@ -86,7 +86,7 @@ public IEnumerable<Mutant> FilterMutants(IEnumerable<Mutant> mutants, ReadOnlyFi
 
         private IEnumerable<Mutant> SetNotRunMutantsToIgnored(IEnumerable<Mutant> mutants)
         {
-            foreach (var mutant in mutants.Where(m => m.ResultStatus == MutantStatus.NotRun))
+            foreach (var mutant in mutants.Where(m => m.ResultStatus == MutantStatus.NotRun || m.ResultStatus == MutantStatus.NoCoverage))
```
rouke-broersma, line 89: I think this breaks the baseline diff feature as now any nocoverage results from previous runs will be set to ignored. While that is semi-correct it is not fully correct. @Stijn-Rutten

dupdob: My understanding is that the purpose of this function is to mark mutations that are not changed. 
At this moment in Stryker test cycle, mutants are marked as not run, except if there is no coverage, in which case they have already been flagged as 'NoCoverage'.

A nicer approach may be to delay the 'NoCoverage' flagging to the test phase, but filtering them out earlier helps keep the test phase clean; and implies a significant refactoring effort

rouke-broersma: Mhm I see your point and also just now realize that I myself said it was a bug before. However we do need to take a close look at how this interacts with the baseline diff feature and I'm also realizing that some of the other helpers in this class might be wrong. Such as SetMutantStatusForNonCSharpFileChanged, which currently overwrites the calculated NoCoverage status which is not correct I think.

dupdob: It turns out I am not familiar with the diff workflows as I always test everything. Hence I used unit test to fix this, I think it would be faster if someone more familiar with the feature took his PR or the fix over.

I can still look into this, but I am busy with other stuff right now, so not sure when I will have a proper testing setup for this.

rouke-broersma: I think the pr is good as it is but I'll take a close look when I review 👌

richardwerkman:
@psfinaki I see your master is on an older version (0.21). We fixed a bug in 0.22 that diff could not be enabled. So you might be right that your diff is not working locally. Pulling your master branch should fix the issue :)

src/Stryker.Core/Stryker.Core/MutantFilters/DiffMutantFilter.cs
rouke-broersma: I have been testing this solution and this causes NoCoverage results to be set to Ignored in all cases. This does not change the mutation score but it does mean that if all your new code is uncovered you will not have any result on your new code. Everything will be marked as Ignored due to no change.

dupdob: so, if I understand you correctly, it sadly does not correct anything but it makes things a bit worse
am I right? 🤣

rouke-broersma: Well it fixes the reported problem but causes a problem for the other causes so I guess net change is 0 :)

dupdob: this is odd: I assume this function was used to flag as ignored mutants that belonged to non changed files.
it should change the status of other mutations.

What am I missing here?

rouke-broersma: You are correct, I am investigating

rouke-broersma: The problem was caused because other helper functions did not ignore NoCoverage results. The result is now correct.

My first comment stands though, this breaks NoCoverage results when combined with the Dashboard Compare feature. I think that feature is less important so I'll create a new bug report for it.

psfinaki:
A bit late to the party, but just wanted to confirm that this all works for me as I would imagine. Thanks!

psfinaki:
By the way, is this included in 0.22.7? I guess no?

rouke-broersma:
This should be included in 0.22.7 yes. It is not showing up in the changelog because I forgot to change the commit name to be prefixed with fix: 🙈

psfinaki:
Oh okay, yes I was misled a bit. Thanks!

PR #1643, opened 2021-07-29
Update documentation to include wrapping quotes

jayudey:
I was going through the docs and couldn't run the command as is, so I updated it to match some of the other commands that pass in an array and thought I'd propose the change.

Thanks again for stryker!


rouke-broersma:
Thank you!

PR #1645, opened 2021-07-30
docs: fix sidebar.json

hugo-vrijswijk:
See https://github.com/stryker-mutator/stryker-mutator.github.io/runs/3199886004?check_suite_focus=true#step:6:37


rouke-broersma:
Grrrr, thank you!

PR #1654, opened 2021-08-06
feat(Logging): Stryker.Core now accepts a logger from stryker runners

### LINKED ISSUE #1623: Logging in new inputs not working
**Describe the bug**
When using logging in the new inputs no message appears in the logs. We have a log buffer that tracks logging before the actual logger is setup. But the logger is set up after the inputs are validated. And the buffer is read before the inputs are validated. So the inputs have no way of logging right now.

**Potential fix**
Move the logger initialisation to CLI. That way we don't have to buffer logs in CLI and we can log anywhere.

richardwerkman:
closes #1623


src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
```diff
@@ -51,10 +49,11 @@ public int Run(string[] args)
             {
                 // app started
                 PrintStrykerASCIIName();
-                PrintStrykerVersionInformationAsync();
 
-                var inputs = InputBuilder.Build(args, app, cmdConfigHandler);
+                InputBuilder.Build(inputs, args, app, cmdConfigHandler);
```
rouke-broersma, line 53: It feels a bit weird that this is a static invocation that takes the inputs it created as params to it's own builder method. Perhaps we should make the InputBuilder non-static and save the inputs as a field.

src/Stryker.Core/Stryker.Core.UnitTest/Reporters/JsonReporterTests.cs
```diff
@@ -15,16 +15,11 @@
...
     {
         public JsonReporterTests()
         {
-            var logOptions = new LogOptions
-            {
-                LogLevel = Serilog.Events.LogEventLevel.Fatal,
-                LogToFile = false
-            };
-            ApplicationLogging.ConfigureLogger(logOptions, null);
+            ApplicationLogging.LoggerFactory = new LoggerFactory();
```
rouke-broersma, line 22:
```suggestion
            ApplicationLogging.LoggerFactory = new LoggerFactory();
```

This should be done by the TestBase right

src/Stryker.Core/Stryker.Core.UnitTest/Reporters/HtmlReporterTests.cs
```diff
@@ -10,12 +10,11 @@
...
 namespace Stryker.Core.UnitTest.Reporters
 {
-    public class HtmlReporterTests
+    public class HtmlReporterTests : TestBase
     {
         public HtmlReporterTests()
         {
-            var options = new LogOptions { LogLevel = Serilog.Events.LogEventLevel.Fatal, LogToFile = false };
-            ApplicationLogging.ConfigureLogger(options, null);
+            ApplicationLogging.LoggerFactory = new LoggerFactory();
```
rouke-broersma, line 17:
```suggestion
```

This should be done by `TestBase` right

src/Stryker.Core/Stryker.Core.UnitTest/Reporters/JsonReporterTests.cs
richardwerkman: Ah yes, I wrote this before I thought of the testbase... I'll remove it

src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
richardwerkman: Agreed. But I went another way to solve it. The strykerinputs are now created in the strykerinputs (where else right?). And I splisted the InputBuilder into a configreader and a loggingInitializer. Better like this?

src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
```diff
@@ -11,18 +11,24 @@
...
+        public int ExitCode { get; private set; } = 0;
 
-        public StrykerCLI(IStrykerRunner stryker = null, IStrykerNugetFeedClient nugetClient = null)
+        public StrykerCli(IStrykerRunner stryker = null,
+            IConfigReader inputBuilder = null,
+            ILoggingInitializer loggingInitializer = null,
+            IStrykerNugetFeedClient nugetClient = null)
         {
             _stryker = stryker ?? new StrykerRunner();
+            _inputBuilder = inputBuilder ?? new ConfigReader();
```
rouke-broersma, line 29: Field should also be renamed

src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
richardwerkman: @rouke-broersma if you agree I think this PR can be merged

rouke-broersma (approved):
Can be merged after fixing code smells

Issue #1656, opened 2021-08-10
Extend exclude mutations


richardwerkman:
**Is your feature request related to a problem? Please describe.**
With the current exclude mutations I can only filter all string mutations or all linq mutations. I'd rather have some more finegrained controll. 

**Describe the solution you'd like**
Extend the ignore-mutations option like this:

```json
{
  "stryker-config": {
    "ignore-mutations": [
      "string",
      "linq.FirstOrDefault"
    ]
}
```

This is not specific for linq mutations but could be used for every mutation. This means we have to give every mutation a name and give them a place in the docs. For linq mutations the names a pretty clear but for others we might have to think of some clever names.

PR #1660, opened 2021-08-14
feat(ignore-mutations): Allow ignoring specific linq mutations

leonardosimoura:
Add ignore options to linq expresssions. 
for #1546 |  #1656 

Feel free to make any suggestions


src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/ExcludeLinqExpressionFilterTests.cs
```diff
@@ -0,0 +1,168 @@
...
+        [InlineData(LinqExpression.OrderByDescending)]
+        [InlineData(LinqExpression.ThenBy)]
+        [InlineData(LinqExpression.ThenByDescending)]
+        [InlineData(LinqExpression.Reverse)]
+        [InlineData(LinqExpression.AsEnumerable)]
+        [InlineData(LinqExpression.Union)]
+        [InlineData(LinqExpression.Intersect)]
+        [InlineData(LinqExpression.Concat)]
+        [InlineData(LinqExpression.Except)]
+        public void ShouldRemoveLinqExpression(LinqExpression exp)
```
richardwerkman, line 58:
```suggestion
        public void ShouldRemoveLinqExpressionWhenFilterIsCorrect(LinqExpression exp)
```

src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/ExcludeLinqExpressionFilterTests.cs
```diff
@@ -0,0 +1,168 @@
...
+        [InlineData(LinqExpression.OrderByDescending, LinqExpression.OrderBy)]
+        [InlineData(LinqExpression.ThenBy, LinqExpression.ThenByDescending)]
+        [InlineData(LinqExpression.ThenByDescending, LinqExpression.ThenBy)]
+        [InlineData(LinqExpression.Reverse, LinqExpression.AsEnumerable)]
+        [InlineData(LinqExpression.AsEnumerable, LinqExpression.Reverse)]
+        [InlineData(LinqExpression.Union, LinqExpression.Intersect)]
+        [InlineData(LinqExpression.Intersect, LinqExpression.Union)]
+        [InlineData(LinqExpression.Concat, LinqExpression.Except)]
+        [InlineData(LinqExpression.Except, LinqExpression.Concat)]
+        public void ShouldNotRemoveLinqExpression(LinqExpression exp, LinqExpression excludedExp)
```
richardwerkman, line 111:
```suggestion
        public void ShouldNotRemoveLinqExpressionWhenFilterIsDifferentName(LinqExpression exp, LinqExpression excludedExp)
```

src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/ExcludeLinqExpressionFilterTests.cs
```diff
@@ -0,0 +1,168 @@
...
+        [InlineData(LinqExpression.Concat)]
+        [InlineData(LinqExpression.Except)]
+        public void ShouldRemoveLinqExpression(LinqExpression exp)
+        {
+            // Arrange
+            var target = new LinqMutator();
+
+            var expression = GenerateExpressions(exp.ToString());
+
+            var result = target.ApplyMutations(expression).ToList();
```
richardwerkman, line 65: I think it's confusing this var is called result. I expect a result after an Act step. Not an arrange step. Maybe name it `var mutations`?

src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/ExcludeLinqExpressionFilterTests.cs
```diff
@@ -0,0 +1,168 @@
...
+        [InlineData(LinqExpression.Concat, LinqExpression.Except)]
+        [InlineData(LinqExpression.Except, LinqExpression.Concat)]
+        public void ShouldNotRemoveLinqExpression(LinqExpression exp, LinqExpression excludedExp)
+        {
+            // Arrange
+            var target = new LinqMutator();
+
+            var expression = GenerateExpressions(exp.ToString());
+
+            var result = target.ApplyMutations(expression).ToList();
```
richardwerkman, line 118: Same for this. Name this `var mutations`

src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/ExcludeLinqExpressionFilterTests.cs
```diff
@@ -0,0 +1,168 @@
...
+                {
+                    excludedExp
+                }
+            });
+
+            // Assert
+            results.ShouldNotBeEmpty();
+        }
+
+
```
richardwerkman, line 137: I'm missing a test with an empty ExcludedLinqExpressions. It should not filter anything.

docs/configuration.md
```diff
@@ -267,6 +267,16 @@ Config file: `"ignore-mutations": ['string', 'logical']`
 
 Turn off mutations that are not currently relevant to your project. 
 
+#### `Linq expressions`
+
+Is possible to disable specifics linq expressions using:
```
richardwerkman, line 272:
```suggestion
It's possible to disable specific linq expressions using:
```

docs/configuration.md
```diff
@@ -267,6 +267,16 @@ Config file: `"ignore-mutations": ['string', 'logical']`
...
+```json
+"stryker-config": {
+    "ignore-mutations": ["linq.First", "linq.Sum"]
```
richardwerkman, line 275:
```suggestion
    "ignore-mutations": [
        "linq.First",
        "linq.Sum"
    ]
```

src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/MutantFilterFactoryTests.cs
```diff
@@ -80,6 +82,30 @@ public void MutantFilterFactory_Creates_DiffMutantFilter_When_Since_Enabled()
...
+
+            var diffProviderMock = new Mock<IDiffProvider>(MockBehavior.Loose);
+            var branchProviderMock = new Mock<IGitInfoProvider>(MockBehavior.Loose);
+            var baselineProvider = new Mock<IBaselineProvider>(MockBehavior.Loose);
+
+            // Act
+            var result = MutantFilterFactory.Create(options, null, diffProviderMock.Object, baselineProvider.Object, branchProviderMock.Object);
+
+            // Assert
+            var resultAsBroadcastFilter = result as BroadcastMutantFilter;
```
richardwerkman, line 102:
```suggestion
            var resultAsBroadcastFilter = result.ShouldBeOfType<BroadcastMutantFilter>();
```

src/Stryker.Core/Stryker.Core.UnitTest/Options/Inputs/IgnoreLinqExpressionInputTests.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            var target = new IgnoreLinqExpressionInput { SuppliedInput = new string[] { } };
+
+            var result = target.Validate();
+
+            result.ShouldBeEmpty();
+        }
+
+        [Fact]
+        public void ShouldReturnMultipleLinqExpressions()
```
richardwerkman, line 61: What if only one value starts with `linq.` and another is in the wrong format? We could use a test like that.

src/Stryker.Core/Stryker.Core/Options/Inputs/IgnoreLinqExpressionInput.cs
```diff
@@ -0,0 +1,51 @@
...
+                if (!linqExpressionsInput.Any())
+                    throw new InputException($"Invalid excluded linq expression. The excluded linq expression options are [{string.Join(", ", linqExpressions.Select(x => "linq."+ x.ToString()))}]");
+
+
+                foreach (var linqExpressionToExclude in linqExpressionsInput)
+                {
+                    // Find any LinqExpression that matches the name passed by the user
+                    var linqExpression = linqExpressions.FirstOrDefault(
+                        x => x.ToString().ToLower().Equals(linqExpressionToExclude[1].ToLower()));
+                    if (linqExpression != LinqExpression.None)
```
richardwerkman, line 40: This is not needed right? As LinqExpression.None is already filtered from `linqExpressions` at line 25

src/Stryker.Core/Stryker.Core/Options/Inputs/IgnoreLinqExpressionInput.cs
```diff
@@ -0,0 +1,51 @@
...
+                var excludedLinqExpressions = new List<LinqExpression>();
+
+                // Get all LinqExpression
+                var linqExpressions = Enum.GetValues(typeof(LinqExpression))
+                    .Cast<LinqExpression>()
+                    .Where(w => w != LinqExpression.None);
+
+                var linqExpressionsInput = SuppliedInput
+                                        .Select(s => s.Split("."))
+                                        .Where(s => s.Length == 2);
```
richardwerkman, line 29: Why should length be 2?

src/Stryker.Core/Stryker.Core/Options/Inputs/IgnoreMutationsInput.cs
```diff
@@ -24,7 +24,7 @@ public IEnumerable<Mutator> Validate()
                     .Cast<Mutator>()
                     .ToDictionary(x => x, x => x.GetDescription());
 
-                foreach (var mutatorToExclude in SuppliedInput)
+                foreach (var mutatorToExclude in SuppliedInput.Where(w => !w.ToLower().Contains(".")))
```
richardwerkman, line 27: Why use `ToLower()` here? I don't think that does affect the '.' character

src/Stryker.Core/Stryker.Core/Options/StrykerInputs.cs
```diff
@@ -86,6 +87,7 @@ public StrykerInputs(IFileSystem fileSystem = null)
         public MutateInput MutateInput { get; init; } = new();
         public IgnoreMethodsInput IgnoredMethodsInput { get; init; } = new();
         public IgnoreMutationsInput ExcludedMutationsInput { get; init; } = new();
+        public IgnoreLinqExpressionInput ExcludedLinqExpressionInput { get; init; } = new();
```
richardwerkman, line 90: The Input is never filled. Please see our CLI project. You need to add a line to https://github.com/stryker-mutator/stryker-net/blob/v1.0/src/Stryker.CLI/Stryker.CLI/JsonConfigHandler.cs 

That will fill the input and actually use user input. Please also add a test for that. I'd like to actually see this feature used in https://github.com/stryker-mutator/stryker-net/blob/v1.0/src/Stryker.CLI/Stryker.CLI.UnitTest/filled-stryker-config.json and test if the value is set in the input. That way we can be sure the value is read from the json

richardwerkman (changes requested):
Thank you for this quick pull request! The code is looking great but the feature won't work yet like this. I've commented how you can fix this. When the comments are resolved I think this can be merged :)

rouke-broersma (changes requested):

src/Stryker.Core/Stryker.Core/Options/Inputs/IgnoreLinqExpressionInput.cs
leonardosimoura: The ideia was to only use "linq.{expression} format,  but with this the user never will know he is passing a wrong format
will refactor this

src/Stryker.Core/Stryker.Core/Options/Inputs/IgnoreLinqExpressionInput.cs
leonardosimoura: Changing to use .Any to check if the input is a valid LinqExpression

src/Stryker.Core/Stryker.Core.UnitTest/Options/Inputs/IgnoreLinqExpressionInputTests.cs
leonardosimoura: added ShouldOnlyValidateCorrectFormat test

src/Stryker.Core/Stryker.Core/Options/StrykerInputs.cs
leonardosimoura: I changed the ConfigFileTests.WithConfigFile_ShouldStartStrykerWithConfigFileOptions  Tests
adding :
```csharp
actualInputs.ExcludedMutationsInput.SuppliedInput.ShouldContain("linq.FirstOrDefault");
````

File: src/Stryker.CLI/Stryker.CLI/JsonConfigHandler.cs [hunk 14]
```diff
@@ -47,6 +47,7 @@ public static void DeserializeConfig(string configFilePath, IStrykerInputs input
             inputs.LanguageVersionInput.SuppliedInput = jsonConfig.LanguageVersion;
             inputs.TestProjectsInput.SuppliedInput = jsonConfig.TestProjects;
             inputs.ExcludedMutationsInput.SuppliedInput = jsonConfig.IgnoreMutations;
+            inputs.ExcludedLinqExpressionInput.SuppliedInput = jsonConfig.IgnoreMutations;
```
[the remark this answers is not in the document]
leonardosimoura: I changed to be only one input,  but keeped the separeted filter,  because  we starting discussing new rules with mutation level about which linq expressions should run, and having this filter separeted will help to implement this feature.

#1546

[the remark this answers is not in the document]
rouke-broersma: That's a good thought but:

If you use a filter that means the mutation is still placed. The existing mutation level mechanism skips placing the mutation. See: https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/Mutators/MutatorBase.cs

I think it's better to use the existing mutation level mechanism for linq mutation levels and keep that separate from the filter.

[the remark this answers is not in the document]
leonardosimoura: Hmm yeah the LinqMutator is the best place to make the logic for which expressions will run based on mutation level. ( I will create another issue for this one, to focus on this explicit configuration for the linq mutations)

So we keep the separated filter for the "ignore-mutations" configuration?

src/Stryker.Core/Stryker.Core/MutantFilters/MutantFilterFactory.cs
```diff
@@ -46,6 +47,8 @@ private static IEnumerable<IMutantFilter> DetermineEnabledMutantFilters(StrykerO
             {
                 enabledFilters.Add(new DiffMutantFilter(_diffProvider));
             }
+            if (options.ExcludedLinqExpressions.Any())
+                enabledFilters.Add(new ExcludeLinqExpressionFilter());
```
rouke-broersma, line 51:
```suggestion
            if (options.ExcludedLinqExpressions.Any())
            {
                enabledFilters.Add(new ExcludeLinqExpressionFilter());
            }
```

src/Stryker.Core/Stryker.Core/Options/Inputs/IgnoreMutationsInput.cs
```diff
@@ -43,5 +43,39 @@ public IEnumerable<Mutator> Validate()
...
+
+                //Remove the 'linq.'
+                var linqExpressionsInput = linqSuppliedInput
+                                        .Select(s => s.Substring(5));
+
+                foreach (var linqExpressionToExclude in linqExpressionsInput)
+                {
+                    // Validate if the input is a valid LinqExpression
+                    if (!linqExpressions.Any(x => x.ToString().ToLower().Equals(linqExpressionToExclude.ToLower())))
+                        throw new InputException($"Invalid excluded linq expression ({linqExpressionToExclude}). The excluded linq expression options are [{string.Join(", ", linqExpressions.Select(x => x.ToString()))}]");
```
rouke-broersma, line 69:
```suggestion
                    if (!linqExpressions.Any(x => x.ToString().ToLower().Equals(linqExpressionToExclude.ToLower())))
                    {
                        throw new InputException($"Invalid excluded linq expression ({linqExpressionToExclude}). The excluded linq expression options are [{string.Join(", ", linqExpressions.Select(x => x.ToString()))}]");
                    }
```

src/Stryker.CLI/Stryker.CLI/JsonConfigHandler.cs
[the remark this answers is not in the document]
rouke-broersma: ~Well I don't think the filter really has anything linq specific so to me it makes the most sense to just fold it into the excluded mutations completely.~

I only just now realized linq methods have their own syntaxkind which makes this a bit different from exclude methods and I now see why you made it it's own filter.

I guess this is fine for now, though I think we should find a more generic way to filter specific mutations in the future so we can support other mutations.

src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/ExcludeLinqExpressionFilterTests.cs
leonardosimoura: Added ShouldNotFilterMutationsWhenFilterIsEmpty test

Issue #1608, opened 2021-06-28
Error During Analysis


JasonBock:
**Describe the bug**
I ran Stryker against [this code base](https://github.com/JasonBock/wsharp), and during the analysis, this showed up in the logs:
```
[08:48:59 WRN] Stryker.NET encountered an compile error in C:\Users\jason\source\repos\WSharp\WSharp.Compiler\Syntax\SyntaxTree.cs (at 87:72) with message: 'ParseTokens(SyntaxTree)': not all code paths return a value (Source code: (CompilationUnitSyntax root, ImmutableArray<Diagnostic> diagnostics) ParseTokens(SyntaxTree tree)
                        {
                                var lexer = new Lexer(tree);

                                while ((Stryker75poqjYc5xaE43c.MutantControl.IsActive(696)?!(true):(Stryker75poqjYc5xaE43c.MutantControl.IsActive(695)?false:true)))
                                {
                                        var token = lexer.Lex();

                                        if ((Stryker75poqjYc5xaE43c.MutantControl.IsActive(698)?!(token.Kind != SyntaxKind.EndOfFileToken || includeEndOfFile):(Stryker75poqjYc5xaE43c.MutantControl.IsActive(697)?token.Kind != SyntaxKind.EndOfFileToken && includeEndOfFile:(Stryker75poqjYc5xaE43c.MutantControl.IsActive(699)?token.Kind == SyntaxKind.EndOfFileToken :token.Kind != SyntaxKind.EndOfFileToken )|| includeEndOfFile)))
                                        {
                                                tokens.Add(token);
                                        }

                                        if ((Stryker75poqjYc5xaE43c.MutantControl.IsActive(700)?token.Kind != SyntaxKind.EndOfFileToken:token.Kind == SyntaxKind.EndOfFileToken))
                                        {
                                                return (new CompilationUnitSyntax(tree, new LineStatementsSyntax(tree, new List<LineStatementSyntax>()), token),
                                                        lexer.Diagnostics.ToImmutableArray());
                                        }
                                }
                        })
[08:48:59 WRN] Safe Mode! Stryker will try to continue by rolling back all mutations in method. This should not happen, please report this as an issue on github with the previous error message.
```
I'm reporting it because it asked me to :)

**Desktop (please complete the following information):**
 - OS: Windows
 - Framework Version: .NET 5
 - Stryker Version [e.g. beta 0.22.5]

dupdob:
Thanks for reporting this! 
[...]

PR #1664, opened 2021-08-17
Add full support for local functions

dupdob:
Add support for local functions (via dedicated orchestrator class and needed helper engine).
This will fix safemode triggering on local functions due to `out` parameters our `while(true)` loop. Also:

- Fixes issue #1608.

- Some refactoring to limit code duplication due to the lack of  inheritance between `BaseMethodDeclarationSyntax` and `LocalFunctionDeclarationSyntax`.

- Split `MutantOrchestratorTest` class to increase readability via the addition of `MutationSpecificSyntaxTests` class.

- Align version number between assemblies


dupdob:
Switched PR to draft to give me an opportunity to simplify design further

src/Stryker.DataCollector/Stryker.DataCollector/Stryker.DataCollector.csproj
```diff
@@ -2,9 +2,9 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <AssemblyVersion>0.16.1.0</AssemblyVersion>
-    <FileVersion>0.16.1.0</FileVersion>
-    <Version>0.16.1</Version>
+    <AssemblyVersion>0.22.9.0</AssemblyVersion>
+    <FileVersion>0.22.9.0</FileVersion>
+    <Version>0.22.9</Version>
```
rouke-broersma, line 7: We should probably add the updating of this version number to the prepare-release.js script we use to bump the version number in the other assemblies.

rouke-broersma:
I really like the cleaned up MutationContext usage! Very nice!

src/Stryker.DataCollector/Stryker.DataCollector/Stryker.DataCollector.csproj
dupdob: I had a look at the script file, but there is an assumption that each assembly is a project in itself.

dupdob:
Thanks for the kind word. I had a similar design in an early attempt of my previous refactor effort, but it did not work at the time.

src/Stryker.DataCollector/Stryker.DataCollector/Stryker.DataCollector.csproj
rouke-broersma: I'll create an issue to fix this properly 👍👍

src/Stryker.DataCollector/Stryker.DataCollector/Stryker.DataCollector.csproj
```diff
@@ -2,9 +2,9 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <AssemblyVersion>0.16.1.0</AssemblyVersion>
-    <FileVersion>0.16.1.0</FileVersion>
-    <Version>0.16.1</Version>
+    <AssemblyVersion>0.22.9.0</AssemblyVersion>
```
rouke-broersma, line 5: In #1681 @richardwerkman proposes to simply remove all these version numbers since they are not used anywhere anyway. They would default and remain static but this should not really be a problem afaik.

src/Stryker.RegexMutators/Stryker.RegexMutators/Stryker.RegexMutators.csproj
```diff
@@ -2,6 +2,8 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
+    <AssemblyVersion>0.22.9.0</AssemblyVersion>
```
rouke-broersma, line 5: In #1681 @richardwerkman proposes to simply remove all these version numbers since they are not used anywhere anyway. They would default and remain static but this should not really be a problem afaik.

src/Stryker.DataCollector/Stryker.DataCollector/Stryker.DataCollector.csproj
dupdob: done
I rebased the branch to master head as it was behind

src/Stryker.RegexMutators/Stryker.RegexMutators/Stryker.RegexMutators.csproj
dupdob: done
I rebased the branch to master head as it was behind

PR #1669, opened 2021-08-21
feat(testrunner): Filter which unit tests are used during mutation test

### LINKED ISSUE #420: Feature request: Support for test filters

corentinaltepe:
Replaces #1557 and fixes #420.

- Added the option `--filter` to specify a [TestCaseFilter](https://docs.microsoft.com/en-us/dotnet/core/testing/selective-unit-tests).
- Passed the option to VsTestRunner in two forms: one with the XML RunSettings and one as `TestPlatformOptions` parameter when calling `_vsTestConsole.RunTestsWithCustomTestHost()`. Previous work in #1557 had shown that both were required, since one is used during test discovery and the other is used during test execution.
- Tested the option with a very basic test of two xUnit tests. One test is attributed `[Trait("Category", "unit")]` and the other one `[Trait("Category", "integration")]`. When running Stryker.CLI without `--filter` option, the console outputs `Total number of tests found: 2.` with a correct HTML report. When using either `--filter "Category=unit"` or `--filter "Category=integration"`, then Stryker only analyses the targetted test `Total number of tests found: 1.`

I'm not sure what to do next to go forward with this. Would you require unit or integration tests on this new option? Any chance you might help on that?

### DIFF
--- a/src/Stryker.CLI/Stryker.CLI/CommandLineConfigHandler.cs
+++ b/src/Stryker.CLI/Stryker.CLI/CommandLineConfigHandler.cs
@@ -126,8 +126,6 @@ namespace Stryker.CLI
             AddCliInput(inputs.ProjectUnderTestNameInput, "project", "p", argumentHint: "project-name.csproj");
             AddCliInput(inputs.MutationLevelInput, "mutation-level", "l");
 
-            AddCliInput(inputs.TestCaseFilterInput, "filter", "");
-
             AddCliInput(inputs.LogToFileInput, "log-to-file", "L", optionType: CommandOptionType.NoValue);
             AddCliInput(inputs.VerbosityInput, "verbosity", "V");
 
--- a/src/Stryker.CLI/Stryker.CLI/CommandLineConfigHandler.cs
+++ b/src/Stryker.CLI/Stryker.CLI/CommandLineConfigHandler.cs
@@ -40,7 +40,7 @@ namespace Stryker.CLI
             {
                 var strykerInput = GetStrykerInput(cliInput);
 
-                switch(cliInput.OptionType)
+                switch (cliInput.OptionType)
                 {
                     case CommandOptionType.NoValue:
                         HandleNoValue((IInput<bool?>)strykerInput);
@@ -53,7 +53,7 @@ namespace Stryker.CLI
                         break;
                 }
 
-                switch(strykerInput)
+                switch (strykerInput)
                 {
                     case IInput<string> stringInput:
                         HandleSingleStringValue(cliInput, stringInput);
@@ -83,7 +83,8 @@ namespace Stryker.CLI
             if (int.TryParse(cliInput.Value(), out var value))
             {
                 strykerInput.SuppliedInput = value;
-            } else
+            }
+            else
             {
                 throw new InputException($"Unexpected value for argument {cliInput.LongName}:{cliInput.Value()}. Expected type to be integer");
             }
@@ -124,6 +125,8 @@ namespace Stryker.CLI
 
             AddCliInput(inputs.ProjectUnderTestNameInput, "project", "p", argumentHint: "project-name.csproj");
             AddCliInput(inputs.MutationLevelInput, "mutation-level", "l");
+
+            AddCliInput(inputs.TestCaseFilterInput, "filter", "f");
 
             AddCliInput(inputs.LogToFileInput, "log-to-file", "L", optionType: CommandOptionType.NoValue);
             AddCliInput(inputs.VerbosityInput, "verbosity", "V");
--- a/src/Stryker.CLI/Stryker.CLI/CommandLineConfigHandler.cs
+++ b/src/Stryker.CLI/Stryker.CLI/CommandLineConfigHandler.cs
@@ -126,7 +126,7 @@ namespace Stryker.CLI
             AddCliInput(inputs.ProjectUnderTestNameInput, "project", "p", argumentHint: "project-name.csproj");
             AddCliInput(inputs.MutationLevelInput, "mutation-level", "l");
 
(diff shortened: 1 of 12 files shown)


src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
```diff
@@ -507,7 +509,11 @@ private string GenerateRunSettings(int? timeout, bool forMutantTesting, bool for
                 settingsForCoverage += "<DisableParallelization>true</DisableParallelization>";
             }
             // TODO: get proper timeout
-            var timeoutSettings = timeout!=null ? $"<TestSessionTimeout>{timeout}</TestSessionTimeout>" + Environment.NewLine : string.Empty;
+            var timeoutSettings = timeout != null ? $"<TestSessionTimeout>{timeout}</TestSessionTimeout>" + Environment.NewLine : string.Empty;
+
+            var testCaseFilter = string.IsNullOrWhiteSpace(_options.TestCaseFilter) ?
+                string.Empty : $"<TestCaseFilter>{_options.TestCaseFilter}</TestCaseFilter>" + Environment.NewLine;
```
corentinaltepe, line 515: I see potential issues here:
1. Should the test case filter syntax be escaped to fit into the XML? I don't know the complete specs for the option, but could complex expression break the XML file, such as: `(FullyQualifiedName~UnitTest1&TestCategory=CategoryA)|Priority=1`?
2. Could there an XML injection attack (or by mistake) if the user typed options like `--filter </RunSetting><AnotherOption />`? Should this be protected against in the user input `Validate()` call? How would you go about this?

corentinaltepe:
It seems the changes have broken 17 unit tests. I will look into it and try and repair them.

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
rouke-broersma: 1. I don't know either. I guess we'll find out ;)
2. Since we don't process the xml there is no xml injection attack from our point of view. However I don't know if vstest protects against this. IMO it is not our job to protect vstest against injection attacks. If it is a risk, vstest needs to protect against this.

src/Stryker.CLI/Stryker.CLI/CommandLineConfigHandler.cs
```diff
@@ -125,6 +126,8 @@ private void PrepareCliOptions(IStrykerInputs inputs)
             AddCliInput(inputs.ProjectUnderTestNameInput, "project", "p", argumentHint: "project-name.csproj");
             AddCliInput(inputs.MutationLevelInput, "mutation-level", "l");
 
+            AddCliInput(inputs.TestCaseFilterInput, "filter", "");
```
rouke-broersma, line 129: I think filter is too generic. How about testCaseFilter. Also I'm not sure this needs to be a command line option and it probably should be a file based option. See https://github.com/stryker-mutator/stryker-net/blob/v1.0/src/Stryker.CLI/Stryker.CLI/FileBasedInput.cs and https://github.com/stryker-mutator/stryker-net/blob/v1.0/src/Stryker.CLI/Stryker.CLI/JsonConfigHandler.cs

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
corentinaltepe: Fair enough. I just wanted to make sure I wasn't missing something.

src/Stryker.CLI/Stryker.CLI/CommandLineConfigHandler.cs
corentinaltepe: Thank you for your feedback.

Agreed to renaming `filter` option to `testCaseFilter`.

I'll remove the command line option and place it to the file based options.

corentinaltepe: The option has been moved from command line to file, and renamed to `test-case-filter`. I used the dashes instead of camel case to be consistent with other options naming style.

rouke-broersma (changes requested):
Changes look good overall.

As for the tests, if you could add this option to this test stryker config: https://github.com/stryker-mutator/stryker-net/blob/v1.0/src/Stryker.CLI/Stryker.CLI.UnitTest/filled-stryker-config.json

And then test the property gets set properly in this test: https://github.com/stryker-mutator/stryker-net/blob/v1.0/src/Stryker.CLI/Stryker.CLI.UnitTest/ConfigFileTests.cs#L44


And for the input you can add a test class here: https://github.com/stryker-mutator/stryker-net/tree/v1.0/src/Stryker.Core/Stryker.Core.UnitTest/Options/Inputs

If you could add a test that tests the helptext and a test that tests the supplied value is used if given, and otherwise the default that would be great and I'll merge this after!

corentinaltepe:
Thank you for your guidance. I will add the tests you've listed.

corentinaltepe:
Hello. Sorry it took so long. I've added the tests as you described.
I truly appreciate your guidance on this PR. I hope the changes fit in with the rest of the code base, and the feature turns useful (I'll be the first to use it).

Please let me know if there's anything to be added before merging.

rouke-broersma:
All good! Awesome! We had hoped to release rc1 of v1 on Friday but couldn't due to pipeline issues. There should soon be a rc and it will include your new feature!

Issue #1659, opened 2021-08-12
Random crashes with FileNotFoundException


psfinaki:
**Describe the bug**

Sometimes Stryker crashes for random projects with this error
<img width="323" alt="error" src="https://user-images.githubusercontent.com/5451366/129216325-ff02dd81-8968-4fef-b600-7f67aa23f175.png">

This happens after throwing a bunch of exceptions with the `FileNotFoundException` on the top, see logs attached.
The process exit code is `-532462766`.

It's important to say this is by far not deterministic. I get a feeling we have this once every 10 times or so. I have just run Stryker against the same project 20 times and it failed 3 times with this thing.

**Logs**

Here are the logs:
[logs.txt](https://github.com/stryker-mutator/stryker-net/files/6976277/logs.txt)

They are pretty similar for these crashes. Once the log also had this attached at the end of the call stack (just before Time Elapsed TIME):
```
   at Stryker.CLI.StrykerCLI.RunStryker(StrykerOptions options) in D:\a\1\s\src\Stryker.CLI\Stryker.CLI\StrykerCLI.cs:line 136
   at Stryker.CLI.StrykerCLI.<>c__DisplayClass7_0.<Run>b__0() in D:\a\1\s\src\Stryker.CLI\Stryker.CLI\StrykerCLI.cs:line 125
   at McMaster.Extensions.CommandLineUtils.CommandLineApplication.<>c__DisplayClass146_0.<OnExecute>b__0(CancellationToken _)
   at McMaster.Extensions.CommandLineUtils.CommandLineApplication.ExecuteAsync(String[] args, CancellationToken cancellationToken)
   at McMaster.Extensions.CommandLineUtils.CommandLineApplication.Execute(String[] args)
   at Stryker.CLI.StrykerCLI.Run(String[] args) in D:\a\1\s\src\Stryker.CLI\Stryker.CLI\StrykerCLI.cs:line 128
   at Stryker.CLI.Program.Main(String[] args) in D:\a\1\s\src\Stryker.CLI\Stryker.CLI\Program.cs:line 16
```

**Expected behavior**

This shouldn't happen.

**Desktop:**
 - OS: Windows
 - Type of project: observed both for `net5` and `netcore3.1`
 - Stryker Version [e.g. beta 0.22.9]

richardwerkman:
[...]

PR #1673, opened 2021-08-24
fix: Return exit code 2 for threshold break violations

### LINKED ISSUE #1672: Cannot distinguish some errors by exit code
**Describe the bug**

Given that Stryker is still not entirely stable, our mutation testing script reruns the execution in case of errors (I should actually publish our script one day here, it is now quite elaborate and can be useful for others). But we actually want to rerun it differently depending on what exactly happened and I haven't found a better way to make a decision rather than looking at the error code.

In particular, I have now two things I want to distinguish:
1) Stryker tries to run against a wrong framework ([this thing](https://github.com/stryker-mutator/stryker-net/issues/1487)). In that case I just retry 8 times (lucky number in China), it helps in 99% cases :D
2) Stryker fails because mutation score is lower than the break threshold (in that case I don't want to rerun or maybe max 1 time haha)

However in both cases the exit code is 1. AFAIU Stryker doesn't control the code in the first case, but explicitly set 1 in the second case. Is there a reason for that particular code? If no, can it be set to something else, e.g. 42?

psfinaki:
closes #1672 

Adding a proper static class for exit codes removing some magic numbers from the source.

### DIFF
--- a/src/Stryker.CLI/Stryker.CLI/ExitCodes.cs
+++ b/src/Stryker.CLI/Stryker.CLI/ExitCodes.cs
@@ -1,0 +1,18 @@
+namespace Stryker.CLI
+{
+    /// <summary>
+    /// Known Stryker exit codes.
+    /// </summary>
+    public static class ExitCodes
+    {
+        /// <summary>
+        /// Stryker is executed against .NET Framework but no solution file is given.
+        /// </summary>
+        public static int NoSolutionFileSpecified = 1;
+
+        /// <summary>
+        /// Mutation score is below the break threshold.
+        /// </summary>
+        public static int BreakThresholdViolated = 2;
+    }
+}
--- a/src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
+++ b/src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
@@ -154,7 +154,7 @@ namespace Stryker.CLI
 
                 Console.WriteLine(Output.Green(" Looks like you've got some work to do :)"));
 
-                ExitCode = ExitCodes.BreakThresholdViolated;
+                ExitCode = ErrorCodes.BreakThresholdViolated;
             }
         }
 
--- a/src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
+++ b/src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
@@ -24,7 +24,7 @@ namespace Stryker.CLI
             _stryker = stryker;
             // Create a log buffer to buffer log messages until the logging is configured.
             _logBuffer = new LogBuffer();
-            ExitCode = 0;
+            ExitCode = ExitCodes.Success;
         }
 
         /// <summary>
@@ -154,7 +154,7 @@ namespace Stryker.CLI
 
                 Console.WriteLine(Output.Green(" Looks like you've got some work to do :)"));
 
-                ExitCode = ErrorCodes.BreakThresholdViolated;
+                ExitCode = ExitCodes.BreakThresholdViolated;
             }
         }
 
--- a/src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
+++ b/src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
@@ -154,7 +154,7 @@ namespace Stryker.CLI
 
                 Console.WriteLine(Output.Green(" Looks like you've got some work to do :)"));
 
-                ExitCode = 42;
+                ExitCode = ExitCodes.BreakThresholdViolated;
             }
         }
 
--- a/src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
+++ b/src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
@@ -154,7 +154,7 @@ namespace Stryker.CLI
 
                 Console.WriteLine(Output.Green(" Looks like you've got some work to do :)"));
 
-                ExitCode = 1;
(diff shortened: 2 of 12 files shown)


src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
```diff
@@ -154,7 +154,7 @@ private void HandleStrykerRunResult(StrykerOptions options, StrykerRunResult res
 
                 Console.WriteLine(Output.Green(" Looks like you've got some work to do :)"));
 
-                ExitCode = 1;
+                ExitCode = 42;
```
rouke-broersma, line 157: I don't think a breaking mutation score is the answer the everything, so perhaps we should pick something else :p 

```suggestion
                ExitCode = 2;
```

I would also like to make this less magic. Perhaps a static class with exit codes with clear names describing what they are used for?

psfinaki: Actually, yes. Code is the best docs. Hang on!

psfinaki: I've updated this. Sorry I didn't mean to be too funny. It's broken window theory - you see one magic number and create another one :facepalm:

rouke-broersma: Haha no problem I can appreciate the humor, just couldn't let it actually get merged like that ;) I was tempted though :)

psfinaki: Sure :D

src/Stryker.CLI/Stryker.CLI/ExitCodes.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    /// <summary>
+    /// Known Stryker exit codes.
+    /// </summary>
+    public static class ExitCodes
+    {
+        /// <summary>
+        /// Stryker is executed against .NET Framework but no solution file is given.
+        /// </summary>
+        public static int NoSolutionFileSpecified = 1;
```
rouke-broersma, line 11: Is this actually what we use exit code 1 for? I would think exit code 1 is used for anything that throws an exception?

psfinaki: Well not sure actually, empirically that's the only case where I've seen 1 (besides violating threshold break which is set explicitly). For example, [here](https://github.com/stryker-mutator/stryker-net/issues/1659) it's `-532462766` (I plan to add it when addressing that issue).

rouke-broersma: Afaik we do not set any specific exit code for no solution specified, we simply throw a stryker input exception: https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/Initialisation/InitialBuildProcess.cs#L36 which we eventually catch here: https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.CLI/Stryker.CLI/Program.cs#L22 where we return exit code 1.

psfinaki: Oh right thaaaat's why it's 1 there. Makes sense. 

I reshuffled this a bit then. Maybe let it be `ErrorCodes` not `ExitCodes` for now as 0 is also technically an exit code but codifying that would require changing a lot things in the codebase and I didn't want to mix refactoring with logical changes in the PR.

rouke-broersma: I would love to know what changes would be needed to support ExitCode 0, because the only place I know about is right here: https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs#L27

If it required a lot of effort to incorporate exit 0 then I agree that ErrorCodes is better for now.

psfinaki: Well technically I would imagine this to be replaced everywhere we test it, so 6 more places as per this:
image

So yeah depends on how clean you want to have your git history. I would be happy to make this as a follow up, but I can do it in this PR as well.

rouke-broersma: Ah right, didn't think of those. Yea if you want make this change go ahead I don't mind!

psfinaki: Alright, hang on then :)

psfinaki: Alright, so I think I've updated all the places in the source. I moved ExitCodes from the CLI to Core to make this possible, hope that's okay.

rouke-broersma: The move feels like we're combining two concepts into one here, because one is the stryker exit codes and the other is exit codes of other apps. If we don't combine these concepts then the stryker exit codes could remain with stryker.cli, since stryker.core does not really 'exit' that makes sense to me. However it is also nice to give a name to the magic numbers used in stryker.core. I think it's fine now, there's little benefit other than clear split concerns by splitting them again. Looks good to me 👍

psfinaki: Yeah we can later reshuffle things around again. Thanks!

psfinaki:
@rouke-broersma will it be possible to make a release with this? This will empower us to start failing PR builds for bad mutation scores :D

Issue #1413, opened 2021-02-16
Stryker cannot handle C# 9 Source Generators


psfinaki:
**Describe the bug**
In our project, we rely on [source generators](https://docs.microsoft.com/en-us/dotnet/core/dotnet-five#source-generators), one of the latest and greatest C# 9 features. Looks like their presence in code breaks Stryker's efforts.

**Logs**
[log-20210216.txt](https://github.com/stryker-mutator/stryker-net/files/5990683/log-20210216.txt)

**Expected behavior**
I don't know if Stryker can theoretically handle all the magic behind source generators. If not, it would be probably enough to just not touch generated stuff at all.

**Environment**
 - OS: Windows
 - Type of project: core
 - Framework Version: .net 5
 - Stryker Version: 0.21.1
 
**Additional context**
I also attach the code to reproduce the issue.
[SourceGeneratorExample.zip](https://github.com/stryker-mutator/stryker-net/files/5990726/SourceGeneratorExample.zip)

I just run `dotnet stryker` from the test folder. If I comment generated code (`//LogGenerator.Generated.Logger.Log();`), everything works.

Mobrockers:
Could you perhaps complain how a 'source generation' is executed? We compile sources manually using Roslyn so any processes built into msbuild instead of Roslyn will have to be recreated by us.

psfinaki:
Well regretfully I don't know how this works in depth, but our usage is only a bit more complex than in the example I posted (we also have some parameters for the logs). [Here](https://www.youtube.com/watch?v=3YwwdoRg2F4) the guys explain the magic a little bit.

But if you have some specific questions, I can ask around in the company, I have contacts to some gurus here :)

psfinaki:
Also, we actually have our logging strategy open sourced [here](https://github.com/geeknoid/LoggingGenerator), that's what eventually we want to get working :)

JohnMcGlynnMSFT:
I've created a [Pull Request](https://github.com/stryker-mutator/stryker-net/pull/1617) that demonstrates how to integrate source generators.

rouke-broersma:
That's amazing thank you!

Issue #1650, opened 2021-08-05
Stryker Logs Warnings For .NET 5.0 Analyzers


pinkfloydx33:
**Describe the bug**
When running Stryker against a .NET 5.0 project, errors/`FileNotFoundException`s are logged to the console as warnings. These appear to be related to built-in analyzers automatically added when targeting .NET 5.0.

This behavior is exhibited in 0.22.9 but not in 0.22.6

**Logs**

> 2021-08-05T09:02:44.0586863-04:00  [DBG] 206 mutants created (edec1dee)
> 2021-08-05T09:02:44.1083293-04:00  [WRN] Analyzer assembly /usr/share/dotnet/sdk/5.0.300/Sdks/Microsoft.NET.Sdk/targets/../analyzers/Microsoft.CodeAnalysis.CSharp.NetAnalyzers.dll could not be loaded. 
> Generated source code may be missing.
> System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.CodeAnalysis.NetAnalyzers, Version=5.0.4.22903, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. The system cannot find the file specified.
> 
> File name: 'Microsoft.CodeAnalysis.NetAnalyzers, Version=5.0.4.22903, Culture=neutral, PublicKeyToken=31bf3856ad364e35'
>    at System.Reflection.RuntimeAssembly.GetExportedTypes()
>    at System.Reflection.Assembly.get_ExportedTypes()
>    at Stryker.Core.Initialisation.Buildalyzer.IAnalyzerResultExtensions.GetSourceGenerators(IAnalyzerResult analyzerResult, ILogger logger)
> 
> [10:31:46 WRN] Analyzer assembly /usr/share/dotnet/sdk/5.0.300/Sdks/Microsoft.NET.Sdk/targets/../analyzers/Microsoft.CodeAnalysis.NetAnalyzers.dll could not be loaded. 
> Generated source code may be missing.
> System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.CodeAnalysis.Workspaces, Version=3.3.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. The system cannot find the file specified.
> 
> File name: 'Microsoft.CodeAnalysis.Workspaces, Version=3.3.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'
>    at System.Reflection.RuntimeAssembly.GetExportedTypes()
[...]

Issue #1534, opened 2021-04-25
Stryker Does Not Honour Static Analysis


MelbourneDeveloper:
**Describe the bug**
Static analysis disallows the mutation, but Stryker ignores this.

**Expected behavior**
Stryker should not add mutations that don't compile due to static analysis via Roslyn code rules

**Desktop (please complete the following information):**
 - OS: Windows
 - Type of project .NET Core
 - Framework Version 5
 - Stryker Version  0.22.3 (beta)

**Additional context**

Here is the mutation:
![image](https://user-images.githubusercontent.com/16697547/115982847-5eb3d300-a5e1-11eb-9027-28652d8a3fae.png)

But, I cannot compile this because of Roslyn's code rules. Roslyn should have squashed the mutation. But, it seems as though Stryker is not honouring the rules.

![image](https://user-images.githubusercontent.com/16697547/115982881-85720980-a5e1-11eb-9760-f05d854eac2d.png)

You can recreate the issue with this repo:
https://github.com/MelbourneDeveloper/RestClient.Net/tree/5/develop

Commit: 4c14356e80906b8017e06c422db814109635dd6d

You can run the batch files in src\RestClient.Net.UnitTests\

rouke-broersma:
To implement analysers we need to figure out how msbuild passes them to roslyn and recreate this. Seeing as analysers are simple csharp classes or dlls I think we can just pass them straight into the compilation either as syntax yree or as assembly reference. Hopefully the analysers are exposed as msbuild properties in which case buildalyzer can probably pass them to us.

MelbourneDeveloper:
@rouke-broersma it seems odd that you would need to do anything at all. If you depend on Roslyn for compilation, it should fail when code rules are violated. The important thing is the analysers Nuget package.

MelbourneDeveloper:
Philosophical point: 

What is the point of mutation testing? 

It's to eliminate the possibility of  potential code entering the system that would destroy the fundamental correctness of the system. 

Static analysis has the same end goal but achieves it through a different means. It reduces possible code paths where it is clear that the system will behave incorrectly. 

For example, static analysis can determine if there is code without a null check. If static analysis catches this, there is no need for mutations because static analysis already prevents them.

rouke-broersma:
I agree that so long as the rule is active and causes compilation errors that mutation testing does not need to point out the possible mutations as well.

rouke-broersma:
[...]

PR #1675, opened 2021-08-24
Calming down source generation messages

psfinaki:
This is related to the discussion in #1413.

We've been enjoying source generation working for a while and haven't encountered any other related issues but yeah this creates a lot of noise which after this "probation period" can be probably silenced a bit. 

Feel free to discard this if you want - it was not a big effort to create :D

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultExtensions.cs
+++ b/src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultExtensions.cs
@@ -122,7 +122,7 @@ namespace Stryker.Core.Initialisation.Buildalyzer
                 }
                 catch (Exception e)
                 {
-                    logger?.LogWarning(e,
+                    logger?.LogDebug(e,
                         $"Analyzer assembly {analyzer} could not be loaded. {Environment.NewLine}" +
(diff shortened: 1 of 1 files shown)


rouke-broersma:
I don't think this should be downgraded, rather the situations where it shows up should be fixed so it is no longer an issue. If this happens your mutation result might not be completely accurate which we should report to the user.

@dupdob what do you think

psfinaki:
Fair, as I say, if you feel there can be a scenario when this message can actually be helpful rather than scaring, we don't need to make it quieter. My idea was to calm it down until we actually see such a scenario.

dupdob:
Issue #1650 tracks this problem and related discussions. I suggest further discussions to happen there and I close this PR as it is too radical a solution

To summarize the problem:
1. Code generators support are a must have for proper compilation moving forward.
2. Code generators are piggy packed over Code Analyzers support infrastructure.
3. Analyzers prove to be difficult to load at compile time while this is a needed feature (as per issue #1534).
4. It is crucial to log when Stryker fails to load a Code Generator as it will result in a failed compilation
5. We cannot distinguish between an analyzer or a code generator without loading them

psfinaki:
@dupdob thanks for the prompt reaction :) 
No problem, I hope we can address this properly then, I will create an issue in our internal backlog for this, in the end it's an interesting nut to crack.

Issue #1642, opened 2021-07-28
Release RC's for V1


rouke-broersma:
V1 contains breaking changes. We should release RC builds so users can test V1 before release.

PR #1681, opened 2021-09-03
chore(Pre releases): prepare scripts for pre releases

### LINKED ISSUE #1670: Update all assembly versions when releasing new Stryker package
We should probably add the updating of this version number to the prepare-release.js script we use to bump the version number in the other assemblies.

_Originally posted by @rouke-broersma in https://github.com/stryker-mutator/stryker-net/pull/1664#discussion_r693360626_

richardwerkman:
Fix #1670 
Is part of #1642

### DIFF
--- a/prepare-release.js
+++ b/prepare-release.js
@@ -20,8 +20,9 @@ const packages = [
     { name: 'dotnet-stryker', path: './src/Stryker.CLI', csproj: './src/Stryker.CLI/Stryker.CLI/Stryker.CLI.csproj' }
 ];
 
-const oldVersionNumber = packagejson.version;
-console.log(`Current package version is ${oldVersionNumber}`);
+const oldVersionPrefix = packagejson.versionPrefix;
+const oldVersionSuffix = packagejson.versionSuffix;
+console.log(`Current package version is ${oldVersionPrefix}${oldVersionSuffix?'-':''}${oldVersionSuffix}`);
 const rl = readline.createInterface({
     input: process.stdin,
     output: process.stdout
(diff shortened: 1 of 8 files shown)


prepare-release.js
```diff
@@ -20,39 +20,51 @@ const packages = [
...
+
         console.log(`Updating changelog for ${package.name}`);
         exec(`npx conventional-changelog-cli -p angular --infile "${package.path}/CHANGELOG.md" --same-file --commit-path ${package.path} --tag-prefix "${package.name}@"`);
         commitMessageLines.push(`- ${package.name}@${newVersionNumber}`);
     });
 
-    console.log('Updating azure-pipelines.yml');
-    replaceVersionNumber('./azure-pipelines.yml', `VersionBuildNumber: $[counter('${oldVersionNumber}', 1)]`, `VersionBuildNumber: $[counter('${newVersionNumber}', 1)]`);
-    replaceVersionNumber('./azure-pipelines.yml', `PackageVersion: '${oldVersionNumber}'`, `PackageVersion: '${newVersionNumber}'`);
+    // console.log('Updating azure-pipelines.yml');
```
rouke-broersma, line 55: Should this be commented out

richardwerkman: Nope, fixed

rouke-broersma:
It's probably a good idea to change all the project versions to v1.0.0 (for example regex mutator project)

richardwerkman:
@rouke-broersma the regex mutator project doesn't have a version. It's just included in stryker.core and will get it's version from there. If we have plans to release it as a seperate package I see the benefit but for now I don't think it needs a version.

rouke-broersma:
> @rouke-broersma the regex mutator project doesn't have a version. It's just included in stryker.core and will get it's version from there. If we have plans to release it as a seperate package I see the benefit but for now I don't think it needs a version.

My bad I thought versions were mandatory. the DataCollector does have a version, I think it probably does need a version there: https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.DataCollector/Stryker.DataCollector/Stryker.DataCollector.csproj

Issue #1694, opened 2021-09-17
Stryker 1.0.0-rc2 fails with error about missing log-level property


mapspiral:
**Describe the bug**
Running `dotnet stryker` on my machine fails with the following error after upgrading from 0.22.7 to 1.0.0-rc2:

```
There was a problem with one of the json properties in your stryker config. Path "stryker-config.log-level", message: "Could not find member 'log-level' on object of type 'FileBasedInput'. Path 'stryker-config.log-level', line 8, position 20."
```
The stryker-config.json looks like this:

```
{
    "stryker-config":
    {
        "reporters": [
            "progress",
            "html"
        ],
        "log-level": "info",
        "log-file":true,
        "timeout-ms": 10000,
        "project-file": "DataServices.Worker.JoinData.UBL.csproj",
        "max-concurrent-test-runners": 2,
        "threshold-high": 80,
        "threshold-low": 70,
        "threshold-break": 60,
        "mutation-level": "Standard",
        "mutate": [
        ],
        "excluded-mutations": [
            "string",
            "Logical operators"
        ],
        "ignore-methods": [
            "*Log*",
            "ToString",
            "*HashCode*"
        ]
    }
}
```
It looks as if the log-level is present in the file.

**Logs**
No logs are generated using the '--log-to-file' option.

**Expected behavior**
I expect Stryker to run and complete as it did in my original version (0.22.7).

**Desktop (please complete the following information):**
 - OS: Linux
 - Type of project Core
 - Framework Version core 5.0.401
 - Stryker Version 1.0.0-rc2

**Additional context**
Nothing additional to add.

rouke-broersma:
Hi @mapspiral 

log-level has been renamed to verbosity in v1. We will publish a migration guide for v0 to v1 when we release a final version for v1. You can downgrade to v0.22.10 until v1 is released to get back the old configuration options.

I will fix the upgrade notice today, it should not suggest to install preview versions, sorry about that!

PR #1695, opened 2021-09-17
fix(update check): Preview features should not trigger new version message

rouke-broersma:
Related to #1694


src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs
```diff
@@ -173,21 +172,31 @@ private void PrintStrykerASCIIName()
...
+            var latestVersion = await client.GetLatestVersionAsync();
+            if (latestVersion > currentVersion)
             {
-                Console.WriteLine(Output.Yellow($@" A new version of Stryker.NET ({latestVersion}) is available. Please consider upgrading using `dotnet tool update -g dotnet-stryker`"));
+                Console.WriteLine(Output.Yellow($@"A new version of Stryker.NET ({latestVersion}) is available. Please consider upgrading!"));
+                Console.WriteLine();
+            }
+
+            var previewVersion = await client.GetPreviewVersionAsync();
+            if(previewVersion > currentVersion)
```
richardwerkman, line 195: I think this will always print when running a preview version. But thats ok for me

rouke-broersma: That's correct. Unfortunately AssemblyVersion does not contain preview version information..

PR #1697, opened 2021-09-18
Fixed some of the sonarqube code smells for v1.0.

### LINKED ISSUE #1607: Clean up sonarqube issues
**Describe the bug**
We've let our code quality slip

**Logs**
https://github.com/stryker-mutator/stryker-net/runs/2882596005

**Expected behavior**
Clean it all up

corentinaltepe:
Partially fixes #1607.

I offer to address some of the [Sonarqube code smells on v1.0](https://sonarcloud.io/dashboard?branch=v1.0&id=stryker-net).

A couple comments:
- I use the plugin Code Cleanup On Save, which fixed some of the default code styling. It fixed spacing, and re-ordered imports in the files I modified. I would recommend setting up a [.editorconfig](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/code-style-rule-options) file in the future. That would make future PR reviews easier and the style more consistent and easier to enforce.
- I fixed the `IDisposable` implementations the lazy way, by setting the classes `sealed`. I wasn't sure how to properly implement IDisposable otherwise (I'd need to be more familiar with the code base). I'd understand if you wish to revert this.

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/VsTestRunnersTest.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/VsTestRunnersTest.cs
@@ -111,7 +111,6 @@ namespace Stryker.Core.UnitTest.TestRunners
                     }).Object,
                 ProjectContents = _projectContents
             };
-            //CodeInjection.HelperNamespace = "Stryker.Core.UnitTest.TestRunners";
 
             _testCases = new List<TestCase> { firstTest, secondTest };
         }
@@ -123,7 +122,7 @@ namespace Stryker.Core.UnitTest.TestRunners
                 ContinueWith((_, u) => discoveryEventsHandler.HandleDiscoveryComplete((int)u, null, aborted), tests.Count);
         }
 
-        private TestCase BuildCase(string name) => new TestCase(name, _executorUri, _testAssemblyPath) { Id = new Guid() };
+        private TestCase BuildCase(string name) => new TestCase(name, _executorUri, _testAssemblyPath) { Id = Guid.NewGuid() };
 
         private TestCase FindOrBuildCase(string testResultId) => _testCases.FirstOrDefault(@t => t.FullyQualifiedName == testResultId) ?? BuildCase(testResultId);
 
@@ -384,7 +383,6 @@ namespace Stryker.Core.UnitTest.TestRunners
         {
             using var endProcess = new EventWaitHandle(true, EventResetMode.ManualReset);
             BuildVsTestRunner(new StrykerOptions(), endProcess, out var runner);
-            var tests = runner.DiscoverTests();
             // make sure we have discovered first and second tests
             runner.DiscoverTests().Count.ShouldBe(_testCases.Count);
         }
--- a/src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/VsTestRunnersTest.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/VsTestRunnersTest.cs
@@ -122,7 +122,7 @@ namespace Stryker.Core.UnitTest.TestRunners
                 ContinueWith((_, u) => discoveryEventsHandler.HandleDiscoveryComplete((int)u, null, aborted), tests.Count);
         }
 
-        private TestCase BuildCase(string name) => new TestCase(name, _executorUri, _testAssemblyPath) { Id = Guid.NewGuid() };
+        private TestCase BuildCase(string name) => new TestCase(name, _executorUri, _testAssemblyPath) { Id = new Guid() };
 
         private TestCase FindOrBuildCase(string testResultId) => _testCases.FirstOrDefault(@t => t.FullyQualifiedName == testResultId) ?? BuildCase(testResultId);
 
--- a/src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
+++ b/src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
@@ -41,6 +41,9 @@ namespace Stryker.Core.Helpers
                 _ => false
             };
 
+        public static bool NeedsReturn(this LocalFunctionStatementSyntax localFunction) =>
+            !localFunction.ReturnType.IsVoid();
+
         /// <summary>
         /// Checks if a member is static
         /// </summary>
@@ -55,6 +58,15 @@ namespace Stryker.Core.Helpers
         /// <returns>true if this is a getter</returns>
         public static bool NeedsReturn(this AccessorDeclarationSyntax baseMethod) =>
             baseMethod.Keyword.Text == "get";
+
+
+        /// <summary>
+        /// Returns true if the given type is 'void'.
+        /// </summary>
+        /// <param name="type">type syntax</param>
+        /// <returns>type == typeof(void)</returns>
+        public static bool IsVoid(this TypeSyntax type) => type is PredefinedTypeSyntax predefinedType &&
+                                                           predefinedType.Keyword.IsKind(SyntaxKind.VoidKeyword);
 
         /// <summary>
         /// Build a mutated version of a <see cref="SyntaxNode"/>.
--- a/src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
+++ b/src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
@@ -42,19 +42,19 @@ namespace Stryker.Core.Helpers
             };
 
         /// <summary>
-        /// Checks if a member is static
-        /// </summary>
-        /// <param name="node">Member to analyze</param>
-        /// <returns>true if it is a static method, properties, fields...</returns>
-        public static bool IsStatic(this MemberDeclarationSyntax node) => node.Modifiers.Any(x => x.Kind() == SyntaxKind.StaticKeyword);
-
-        /// <summary>
         /// Checks if an accessor is a getter and needs to end with a return
         /// </summary>
         /// <param name="baseMethod"></param>
         /// <returns>true if this is a getter</returns>
         public static bool NeedsReturn(this AccessorDeclarationSyntax baseMethod) =>
             baseMethod.Keyword.Text == "get";
+
+        /// <summary>
+        /// Checks if a member is static
+        /// </summary>
+        /// <param name="node">Member to analyze</param>
+        /// <returns>true if it is a static method, properties, fields...</returns>
+        public static bool IsStatic(this MemberDeclarationSyntax node) => node.Modifiers.Any(x => x.Kind() == SyntaxKind.StaticKeyword);
 
         /// <summary>
         /// Build a mutated version of a <see cref="SyntaxNode"/>.
@@ -65,7 +65,7 @@ namespace Stryker.Core.Helpers
         /// <returns>A copy of <see cref="original"/> with the mutation applied.</returns>
         /// <exception cref="InvalidOperationException">when mutation does not belong to this node.</exception>
         /// <remarks><paramref name="original"/> can be any node that includes the original, non mutated node described in the mutation.</remarks>
-        public static T InjectMutation<T>(this T original, Mutation mutation) where T:SyntaxNode
+        public static T InjectMutation<T>(this T original, Mutation mutation) where T : SyntaxNode
         {
             if (!original.Contains(mutation.OriginalNode))
             {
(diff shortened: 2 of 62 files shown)


src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
```diff
@@ -35,7 +35,7 @@ public class ProjectFileReader : IProjectFileReader
         public IAnalyzerResult AnalyzeProject(
             string projectFilePath,
             string solutionFilePath,
-            string targetFramework)
+            string targetFramework = null)
```
psfinaki: Why this change?

src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/VsTestRunnersTest.cs
```diff
@@ -123,7 +122,7 @@ private static void DiscoverTests(ITestDiscoveryEventsHandler discoveryEventsHan
                 ContinueWith((_, u) => discoveryEventsHandler.HandleDiscoveryComplete((int)u, null, aborted), tests.Count);
         }
 
-        private TestCase BuildCase(string name) => new TestCase(name, _executorUri, _testAssemblyPath) { Id = new Guid() };
+        private TestCase BuildCase(string name) => new TestCase(name, _executorUri, _testAssemblyPath) { Id = Guid.NewGuid() };
```
psfinaki, line 125: Just passing by here but still - this produces a random guid instead of the default one, is this change on purpose?

rouke-broersma:
There is an editorconfig! :) But some of these code smells are from before we had it, or some contributors follow the editorconfig recommendation.

src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
```diff
@@ -56,6 +49,13 @@ baseMethod switch
         public static bool NeedsReturn(this AccessorDeclarationSyntax baseMethod) =>
             baseMethod.Keyword.Text == "get";
 
+        /// <summary>
+        /// Checks if a member is static
+        /// </summary>
+        /// <param name="node">Member to analyze</param>
+        /// <returns>true if it is a static method, properties, fields...</returns>
+        public static bool IsStatic(this MemberDeclarationSyntax node) => node.Modifiers.Any(x => x.Kind() == SyntaxKind.StaticKeyword);
```
rouke-broersma, line 57: This file is part of a difficult merge conflict between master and V1. It is currently being solved by @dupdob and I think also includes this change but to be safe I would prefer if you could revert the changes in this file.

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/RunEventHandler.cs
```diff
@@ -40,38 +40,38 @@ public bool IsComplete()
...
-
-                foreach (var message in next.Messages)
-                {
-                    acc.Messages.Add(message);
-                }
-
-                acc.Duration = acc.EndTime - acc.StartTime;
-                return acc;
-            });
+            var result = _results.Aggregate((TestResult)null, (acc, next) =>
```
rouke-broersma, line 43:
```suggestion
           var result = _results.Aggregate((TestResult)null, (acc, next) =>
```

rouke-broersma, line 43: Something is off about the indenting change here

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
```diff
@@ -100,12 +100,12 @@ public TestRunResult RunAll(ITimeoutValueCalculator timeoutMs, Mutant activeMuta
             return TestMultipleMutants(timeoutMs, activeMutant == null ? null : new List<Mutant> { activeMutant }, update);
         }
 
-        public TestRunResult TestMultipleMutants(ITimeoutValueCalculator timeoutCalc, IReadOnlyList<Mutant> mutants, TestUpdateHandler update)
+        public TestRunResult TestMultipleMutants(ITimeoutValueCalculator timeoutMs, IReadOnlyList<Mutant> mutants, TestUpdateHandler update)
```
rouke-broersma, line 103: In this case the interface probably needs to be changed as the parameter type and name have been changed in the implementation but the interface param name was forgotten.

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
rouke-broersma: This was done to conform with the interface. But in this case I think the interface should be changed rather than the implementation.

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
```diff
@@ -416,7 +417,7 @@ public void CoverageForOneTest(Guid test, IEnumerable<Mutant> mutants)
             Action<RunEventHandler> updateHandler = null,
             int retries = 0)
         {
-            using var eventHandler = new RunEventHandler(_vsTests, _logger, RunnerId);
+            var eventHandler = new RunEventHandler(_vsTests, _logger, RunnerId);
```
rouke-broersma, line 420: Mhm is this done because you're worried there's a 'leak' here due to recursion?

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
```diff
@@ -159,7 +159,8 @@ public TestRunResult TestMultipleMutants(ITimeoutValueCalculator timeoutCalc, IR
                 testCases = null;
             }
 
-            var expectedTests = needAll ? DiscoverTests().Count : testCases.Count;
+            var numberTestCases = testCases == null ? 0 : testCases.Count;
```
rouke-broersma, line 162: Why was this done?

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
corentinaltepe: Hello. That's it. I tried to fix [this code smell](https://sonarcloud.io/code?branch=v1.0&id=stryker-net&selected=stryker-net%3Asrc%2FStryker.Core%2FStryker.Core%2FInitialisation%2FProjectFileReader.cs) (S1006: Method overrides should not change parameter defaults). But it didn't make much sense so I ended up reverting the change.

I'll try changing the interface.

src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/VsTestRunnersTest.cs
corentinaltepe: It wasn't my intention to break current behavior. I only tried quality refactoring.

My IDE's Sonarlint extention flags `new Guid()` with [the rule S4581](https://rules.sonarsource.com/csharp/RSPEC-4581): _"new Guid()" should not be used_.

However, [it doesn't seem to be reported by Sonarcloud](https://sonarcloud.io/code?branch=v1.0&id=stryker-net&selected=stryker-net%3Asrc%2FStryker.Core%2FStryker.Core.UnitTest%2FTestRunners%2FVsTestRunnersTest.cs). Maybe that rule was disabled on purpose?

Anyway, the tests seem to still pass with the change. Should we keep it or revert back to `new Guid()`?

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
corentinaltepe: The line was [reported as a bug](https://sonarcloud.io/code?branch=v1.0&id=stryker-net&selected=stryker-net%3Asrc%2FStryker.Core%2FStryker.Core%2FTestRunners%2FVsTest%2FVsTestRunner.cs) by Sonar, rule S2259: _'testCases' is null on at least one execution path_.

I had to change the business logic a little, by defining that the number of test cases will be 0 if `testCases` is `null`.

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
corentinaltepe: Yes, rule S2997: _Remove the 'using' statement; it will cause automatic disposal of 'eventHandler'._ 

I'm not totally sure what I'm doing here. I see the `eventHandler` is returned by the method when `!_vsTestFailed || retries > 10`. I wasn't sure of the meaning of returning a disposed event handler. Wouldn't it be a bug?

I offer to : 
- not dispose the event handler when `!_vsTestFailed || retries > 10` and return it.
- otherwise dispose it, then `return RunTestSession(mutantTestsMap, true, runSettings, updateHandler, ++retries);`

Should I revert the change?

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/RunEventHandler.cs
corentinaltepe: Yes, something was off. It's fixed now.

src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
corentinaltepe: I've reverted all changes on this file to avoid merge conflicts.

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
corentinaltepe: Understood. The interface's method parameter was renamed to `timeoutCalc` and both implementations (`VsTestRunner` and `VsTestRunnerPool`) were updated accordingly.

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
corentinaltepe: Oh, actually it was already done 3 days ago in #1684. Sonarqube analysis was from 16 days ago, that's what got me confused.

src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/VsTestRunnersTest.cs
psfinaki: Well I looked at the link, it recommends
```
An empty GUID, in which case Guid.Empty is clearer.
A randomly-generated GUID, in which case Guid.NewGuid() should be used.
A new GUID with a specific initialization, in which case the initialization parameter is missing.
```
So if you want to make an identical change use `Guid.Empty`. At the same time, as long as tests pass, you can probably use whatever :)

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
rouke-broersma: I don't think we can ever actually end up in this situation but because if we have no test cases there's nothing we can actually test anyway, but it is technically better 👌

src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/VsTestRunnersTest.cs
rouke-broersma: Looks like it was reverted already. I don't think it matters in any case. We can fix this up again some other time 👌

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
rouke-broersma: Cleaning up the VsTestRunner is probably a job in and of itself, let's revert the change and let some other unfortunate person deal with it another time ;)

dupdob: There is a design issue for sure, but the event handler needs to be disposed here.
The design problem is that `RunEventHandler` aggregates event handling ressources that require to be disposed as well as plain data.
In practice, the eventhandling part is only used within the testrunner logic while the data can be used later.
There should be two classes for that, I think this is a legacy decision that never was improved.

corentinaltepe: The changes around disposing the event handler have been reverted.

PR #1698, opened 2021-09-19
chore: Update v1 with master

dupdob:
This PR only contains  a V1 branched merged with master. No fix nor new features.
Refactoring is due after this merge as there are still some design inconsistencies, but **this version works**


rouke-broersma:
Let's use this pr to validate the build and then merge this with a normal merge locally by hand. Merging from github will be a squash merge which we don't want in this specific case.

PR #1702, opened 2021-09-25
Assembly Loader For Source Generator References

JohnMcGlynnMSFT:
Copied the logic from [Roslyn source](https://github.com/dotnet/roslyn/blob/315c2e149ba7889b0937d872274c33fcbfe9af5f/src/CodeStyle/Tools/Program.cs#L55) and it did not throw any exceptions loading the source generators.

https://github.com/stryker-mutator/stryker-net/issues/1650

Please try this out in your projects and comment, and I will upgrade this to a full PR if it is generally valuable.


rouke-broersma:
@JohnMcGlynnMSFT we would prefer to merge this on v1 when it's ready, could you rebase this on the v1 branch?

rouke-broersma:
> > @JohnMcGlynnMSFT we would prefer to merge this on v1 when it's ready, could you rebase this on the v1 branch?
> 
> Breaking change: v1.0 is missing StrykerCompilationException. Do you have a plan to catch up that branch first?

We removed Stryker from all exception names. See https://github.com/stryker-mutator/stryker-net/blob/v1.0/src/Stryker.Core/Stryker.Core/Exceptions/CompilationException.cs

JohnMcGlynnMSFT:
> > > @JohnMcGlynnMSFT we would prefer to merge this on v1 when it's ready, could you rebase this on the v1 branch?
> > 
> > 
> > Breaking change: v1.0 is missing StrykerCompilationException. Do you have a plan to catch up that branch first?
> 
> We removed Stryker from all exception names. See https://github.com/stryker-mutator/stryker-net/blob/v1.0/src/Stryker.Core/Stryker.Core/Exceptions/CompilationException.cs
> 
> V1 is up to date with master

I didn't notice that change. Thanks!

PR #1707, opened 2021-10-01
Remove Newtonsoft.Json dependency

0xced:
Instead, use System.Text.Json everywhere.

### JSON Configuration File

Since there's [no equivalent][1] to `MissingMemberHandling` with `System.Text.Json`, the error messages in case of erroneous keys in the JSON configuration file have been improved. For example, if you have a typo in the `enabled` key you would get this error:

> The allowed keys for the "stryker-config.since" object are { "enabled", "ignore-changes-in", "target" } but "enable" was found in the config file at "[…]/tests/stryker-config.json"

Those better error messages also required to have a `JsonPropertyName` attribute on _all_ the properties. This is probably a good thing anyway since it makes the code more _greppable_.

### Newtonsoft.Json transitive dependency

Unfortunately, `Newtonsoft.Json` is still there transitively (through Buildalyzer/3.2.2 → Microsoft.Extensions.DependencyModel/2.1.0 → Newtonsoft.Json/9.0.1) but there's nothing we can do about it for now, see https://github.com/microsoft/vstest/issues/2488#issuecomment-932036883 for a full explanation.

[1]: https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to#missingmemberhandling

### DIFF
--- a/src/Stryker.CLI/Stryker.CLI.UnitTest/ConfigFileTests.cs
+++ b/src/Stryker.CLI/Stryker.CLI.UnitTest/ConfigFileTests.cs
@@ -18,7 +18,7 @@ namespace Stryker.CLI.UnitTest
             var mock = new Mock<IStrykerRunner>(MockBehavior.Strict);
             var options = new StrykerOptions()
             {
-                Thresholds = new Thresholds()
+                Thresholds = new Stryker.Core.Options.Thresholds()
                 {
                     High = 80,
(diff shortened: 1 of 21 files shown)


rouke-broersma:
Apologies on the delay. we'll look into this pr tomorrow.

PR #1708, opened 2021-10-02
Fix finding the project under test

0xced:
If the test projects are defined in the stryker-config.json file (under stryker-config.test-projects), it becomes impossible to find the project under test unless it is in the current working directory since `FindProjectFile` is used instead of `FindProjectUnderTest`.

It used to work before #1039 where a behavioural change was introduced. @richardwerkman do you remember what was the intention behind that change?

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
+++ b/src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
@@ -48,7 +48,6 @@ namespace Stryker.Core.Initialisation
             var projectInfo = new ProjectInfo(_fileSystem);
             // Determine test projects
             var testProjectFiles = new List<string>();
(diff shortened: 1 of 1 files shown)


rouke-broersma:
Iirc this is intended behavior. If you set the test projects to use then you should run Stryker from the project under test. If you have one test project then you should run Stryker from the test project. This doesn't necessarily need to be a requirement but it makes it easier for us to reason about the project structure under test. I am not 100 percent sure stryker in its current state could run with all features from just any working directory.

0xced:
OK I understand. The error message could be probably improved, though. I had to dig through the source code to understand this:
> No .csproj file found, please check your project directory at /Users/0xced/Projects/nodatime

Something along those lines could be better:

> Since you specified the test projects in the stryker-config.json file you must run Stryker inside the directory containing the project to mutate.

PR #1663, opened 2021-08-17
feat(dashboard reporter): Retrieve the project name and version from the project under test using source link

0xced:
Thanks to [SourceLink][1], the repository URL and the full version (including the git SHA1) of a project can be included in the produced assembly.

This commit uses the information computed by SourceLink to automatically retrieve the project name (github.com/organization/project) and project version which are required for the dashboard reporter.

With this technique, the table on the [Stryker dashboard documentation][2] can be updated with ✅ everywhere for the Stryker.NET column since it doesn't depend on a specific CI provider to provide the repository URL and version.

Note that I have run `Stryker.CLI --reporter progress --reporter html --reporter dashboard` on my local checkout of https://github.com/0xced/serilog-formatting-log4net/tree/20e86d55b4bd9e3682c3608530398e794df12481 and it automatically retrieved the project name (`github.com/0xced/serilog-formatting-log4net`) and project version (`1.0.0-rc.2+20e86d55b4bd9e3682c3608530398e794df12481`). The report was successfully uploaded to the [Stryker dashboard][3]

[1]: https://github.com/dotnet/sourcelink#readme
[2]: https://stryker-mutator.io/docs/General/dashboard/#retrieved-from-environment
[3]: https://dashboard.stryker-mutator.io/reports/github.com/0xced/serilog-formatting-log4net/1.0.0-rc.2%2B20e86d55b4bd9e3682c3608530398e794df12481

PR #1532, opened 2021-04-24
Retrieve the project name and version from the project under test

0xced:
Thanks to [SourceLink][1], the repository URL and the full version (including the git SHA1) of a project can be included in the produced assembly.

This commit uses the information computed by SourceLink to automatically retrieve the project name (github.com/organization/project) and project version which are required for the dashboard reporter.

With this technique, the table on the [Stryker dashboard documentation][2] can be updated with ✅ everywhere for the Stryker.NET column since it doesn't depend on a specific CI provider to provide the repository URL and version.

Note that I have run `Stryker.CLI --reporters ["progress","html","dashboard"]` on my local checkout of https://github.com/0xced/serilog-formatting-log4net/tree/20e86d55b4bd9e3682c3608530398e794df12481 and it automatically retrieved the project name (`github.com/0xced/serilog-formatting-log4net`) and project version (`1.0.0-rc.2+20e86d55b4bd9e3682c3608530398e794df12481`). The report was sucessfully uploaded to the [Stryker dashboard][3]

[1]: https://github.com/dotnet/sourcelink#readme
[2]: https://stryker-mutator.io/docs/General/dashboard/#retrieved-from-environment
[3]: https://dashboard.stryker-mutator.io/reports/github.com/0xced/serilog-formatting-log4net/1.0.0-rc.2%2B20e86d55b4bd9e3682c3608530398e794df12481

PR #1710, opened 2021-10-02
fix: Dashboard baseline project name and version are not set

0xced:
The automatic detection of project name and version introduced in https://github.com/stryker-mutator/stryker-net/pull/1663 broke the DashboardBaselineProvider because the project name would not be set and thus the report would be sent to an invalid URL: `/api/reports//version`

This commit addresses this issue by setting the project name and version on the StrykerOptions rather than on the DashboardClient and DashboardReporter (like it was originally done in https://github.com/stryker-mutator/stryker-net/pull/1532).

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Reporters/JsonReporter/JsonReportFileComponent.cs
+++ b/src/Stryker.Core/Stryker.Core/Reporters/JsonReporter/JsonReportFileComponent.cs
@@ -33,11 +33,11 @@ namespace Stryker.Core.Reporters.Json
                 var jsonMutant = new JsonMutant
                 {
                     Id = mutant.Id.ToString(),
-                    MutatorName = mutant.Mutation.DisplayName,
+                    MutatorName = mutant.Mutation.DisplayName ?? "",
                     Replacement = mutant.Mutation.ReplacementNode.ToFullString(),
                     Location = new JsonMutantLocation(mutant.Mutation.OriginalNode.GetLocation().GetMappedLineSpan()),
                     Status = mutant.ResultStatus.ToString(),
-                    Description = mutant.Mutation.Description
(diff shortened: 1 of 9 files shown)


src/Stryker.Core/Stryker.Core/Reporters/JsonReporter/JsonReportFileComponent.cs
```diff
@@ -33,11 +33,11 @@ public JsonReportFileComponent(ReadOnlyFileLeaf file, ILogger logger = null)
...
                 {
                     Id = mutant.Id.ToString(),
-                    MutatorName = mutant.Mutation.DisplayName,
+                    MutatorName = mutant.Mutation.DisplayName ?? "",
                     Replacement = mutant.Mutation.ReplacementNode.ToFullString(),
                     Location = new JsonMutantLocation(mutant.Mutation.OriginalNode.GetLocation().GetMappedLineSpan()),
                     Status = mutant.ResultStatus.ToString(),
-                    Description = mutant.Mutation.Description
+                    Description = mutant.Mutation.Description ?? ""
                 };
```
rouke-broersma, line 41: Did you find an issue with the DisplayName or Description being null?

rouke-broersma (changes requested):
You have some unit test failures: https://dev.azure.com/stryker-mutator/Stryker/_build/results?buildId=8239&view=ms.vss-test-web.build-test-results-tab

0xced:
I just addressed the `NullReferenceException` in commit 1e106bdf9a332e20f904f55e08c1b23d8b3fd093 that I pushed to the my `Fix-DashboardBaselineProvider` branch but for some reason, this pull request does not see [this commit](https://github.com/0xced/stryker-net/commit/1e106bdf9a332e20f904f55e08c1b23d8b3fd093).

PR #1712, opened 2021-10-04
should dispose of process

SimonCropp:


rouke-broersma:
Could you target this to the v1 branch? We're not currently accepting prs to master.

SimonCropp:
done

richardwerkman:
Thanks for this fix! It seems we overlooked this one. I have one question? Did you find this after you encountered a bug? We have not noticed anything before but maybe we missed something. In any case, thanks for the contribution :)

SimonCropp:
@richardwerkman i was just reading the code and noticed. I am not currently using stryker

PR #1746, opened 2021-10-29
feat(mutating): Don't mutate strings as string literal when they contain regex

### LINKED ISSUE #1132: Disable string mutation on regexes when regex mutator is enabled
The string mutation on regex strings is not needed when placing regex mutations. 

It would be nice to skip the string mutation when the regex mutator is enabled. The string mutation should still be placed on other strings when the regex mutator is enabled.

**Additional context**
We should wait for #1123 before we start working on this.

tom171296:
closes #1132


src/Stryker.Core/Stryker.Core/Mutators/StringMutator.cs
```diff
@@ -12,7 +14,12 @@ public class StringMutator : MutatorBase<LiteralExpressionSyntax>, IMutator
         public override IEnumerable<Mutation> ApplyMutations(LiteralExpressionSyntax node)
         {
             var kind = node.Kind();
-            if (kind == SyntaxKind.StringLiteralExpression && !(node.Parent is ConstantPatternSyntax))
+            // Get objectCreationSyntax to check if it contains a regex type.
+            var root = node.Parent?.Parent?.Parent;
```
rouke-broersma, line 18: @richardwerkman Is there no better way to do this?

src/Stryker.Core/Stryker.Core/Mutators/StringMutator.cs
```diff
@@ -12,7 +14,12 @@ public class StringMutator : MutatorBase<LiteralExpressionSyntax>, IMutator
...
+
+            if (!(root is ObjectCreationExpressionSyntax parsedRoot && IsRegexType(parsedRoot))
```
rouke-broersma, line 20:
```suggestion
            if (!IsRegexType(parsedRoot)
```

And move the `root is ObjectCreationExpressionSyntax` to IsRegexType?

src/Stryker.Core/Stryker.Core/Mutators/StringMutator.cs
```diff
@@ -12,7 +14,12 @@ public class StringMutator : MutatorBase<LiteralExpressionSyntax>, IMutator
...
+                && kind == SyntaxKind.StringLiteralExpression
+                && !(node.Parent is ConstantPatternSyntax))
```
rouke-broersma, line 22: Could you perhaps also extract these two conditions to private helper functions with clear names similar to IsRegexType so the condition becomes easier to read?

src/Stryker.Core/Stryker.Core/Mutators/StringMutator.cs
richardwerkman: I'm afraid not. A constructor always contains a argumentlist which contains a first argument which contains the string literal. So we always have to look 3 nodes up to see if the literalexpression is in a regex constructor. We also can't loop upwards as it will be less performant and could be error prone.

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/StringMutatorTests.cs
```diff
@@ -32,5 +32,36 @@ public void ShouldMutate(string original, string expected)
                 .Token.Value.ShouldBe(expected);
             mutation.DisplayName.ShouldBe("String mutation");
         }
+
+        [Fact]
+        public void ShouldNotMutateOnRegexExpression()
+        {
+            var expressionSyntax = SyntaxFactory.ParseExpression("new Regex(\"myregex\")");
```
richardwerkman, line 39: Maybe also test if this works when fully specifying the namespace like: 
```suggestion
            var expressionSyntax = SyntaxFactory.ParseExpression("new System.Text.RegularExpressions.Regex(\"myregex\")");
```

richardwerkman:
Thanks for your first contribution 😄 👏

PR #1759, opened 2021-10-31
fix(Block mutator): Prevent false positives from block mutations

sbergen:
* Don't mutate empty blocks
* Don't mutate blocks in struct constructors containing assignments
* Don't mutate infinite while loops (thanks for the suggestion @Pentiva)
* Correctly deduce the first return-typed ancestor
* Only return `default` if the block contains a return statement
* Add tests

Only thing I'm not happy with is the struct constructor logic, which treats *any* assignment as a potential member assignment. Also, if some of the helper functions I wrote already exist, or should be elsewhere, please comment on that, I'm not too familiar with the code base.

Fixes issues (false positives) and warnings I faced while [updating my project to 1.0](https://github.com/sbergen/Responsible/runs/4059889905?check_suite_focus=true).

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutators/BlockMutatorTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutators/BlockMutatorTests.cs
@@ -1,0 +1,211 @@
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Shouldly;
+using Stryker.Core.Mutants;
+using Stryker.Core.Mutators;
+using Xunit;
+
+namespace Stryker.Core.UnitTest.Mutators
+{
+    public class BlockMutatorTests
+    {
+        [Fact]
+        private void ShouldMutateNonEmptyConstructorOnClass()
+        {
+            var source = @"
+class Program
+{
+    int doesNotNeedToBeInitialized;
+
+    Program()
+    {
+        this.doesNotNeedToBeInitialized = 42;
+    }
+}";
+
+            var mutation = GetMutations(source).ShouldHaveSingleItem();
+            mutation.ReplacementNode.ChildNodes().ShouldBeEmpty();
+        }
+
+        [Fact]
+        private void ShouldNotMutateStructConstructorAssignments()
+        {
+            var source = @"
+struct Program
+{
+    int mustBeInitialized;
+    bool alsoThisMustBeInitialized;
+
+    Program(int value)
+    {
+        this.mustBeInitialized = value;
+
+        if (value == 0)
+        {
+            this.alsoThisMustBeInitialized = true;
+        }
+        else
+        {
+            this.alsoThisMustBeInitialized = false;
+        }
+    }
+}";
+
+            GetMutations(source).ShouldBeEmpty();
+        }
+
+        [Fact]
+        private void ShouldMutateStructConstructorNonAssignmentsAtRoot()
+        {
+            var source = @"
+struct Program
+{
+    Program(int value)
+    {
+        throw new Exception();
+    }
+}";
+
+            GetMutations(source).Count().ShouldBe(1);
+        }
+
+        [Fact]
+        private void ShouldMutateStructConstructorNonAssignmentChild()
+        {
+            var source = @"
+struct Program
+{
+    int mustBeInitialized;
+
+    Program(int value)
+    {
+        if (value == 0)
+        {
+            throw new Exception();
+        }
+
+        this.mustBeInitialized = value;
+    }
+}";
+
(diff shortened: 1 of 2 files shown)


src/Stryker.Core/Stryker.Core.UnitTest/Mutators/BlockMutatorTests.cs
```diff
@@ -0,0 +1,237 @@
...
+        return 1;
+    }
+}";
+
+            var mutations = GetMutations(source).ToList();
+            mutations.Count.ShouldBe(3);
+
+            AssertDefaultReturn(mutations[0]); // Whole method
+            mutations[1].ReplacementNode.ChildNodes().ShouldBeEmpty(); // First if
+            AssertDefaultReturn(mutations[2]); // Second if
```
richardwerkman, line 149:
```suggestion
            var methodBlock = mutations[0];
            var firstIfBlock = mutations[1];
            var secondIfBlock = mutations[2];
            AssertDefaultReturn(methodBlock);
            firstIfBlock.ReplacementNode.ChildNodes().ShouldBeEmpty();
            AssertDefaultReturn(secondIfBlock);
```

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/BlockMutatorTests.cs
```diff
@@ -0,0 +1,237 @@
...
+        void LocalFunction()
+        {
+            return;
+        }
+
+        return 42;
+    }
+}";
+
+            var mutation = GetMutations(source)
```
richardwerkman, line 187: Maybe also assert the number of mutations before filtering them in the where?

src/Stryker.Core/Stryker.Core/Mutators/BlockMutator.cs
```diff
@@ -13,7 +15,53 @@ class BlockMutator : MutatorBase<BlockSyntax>, IMutator
 
         public override IEnumerable<Mutation> ApplyMutations(BlockSyntax node)
         {
-            if (node.Parent is MethodDeclarationSyntax methodDeclaration && methodDeclaration.ReturnType.ToString() != "void")
+            // Note: using local functions below so that the check order can be easily reordered,
+            // if a need to optimize arises at some point.
+
+            bool IsInStructConstructor()
```
richardwerkman, line 21: The ApplyMutations() method is getting a bit big now. Maybe move this local function to a class method?

src/Stryker.Core/Stryker.Core/Mutators/BlockMutator.cs
```diff
@@ -13,7 +15,53 @@ class BlockMutator : MutatorBase<BlockSyntax>, IMutator
...
+            {
+                foreach (var ancestor in node.Ancestors())
+                {
+                    switch (ancestor)
+                    {
+                        // Don't count local functions as being part of the constructor.
+                        // They can't assign to members, so they can be mutated.
+                        case LocalFunctionStatementSyntax:
```
richardwerkman, line 29: This case is not needed right? Which kinda makes the whole switch unneeded. I think it could just be an if statement 👍

richardwerkman (commented):
Thanks for this great contribution! I have only some minor improvements but that means it's already good :)

Pentiva:
As it is in the same vain as this pull request, could it also ignore `while (true)` loops and not mutate them, as they will always result in a infinite loop/timeout? The check would probably be: 
`if (node.Parent is WhileStatementSyntax { Condition: LiteralExpressionSyntax cond } && cond.Kind() == SyntaxKind.TrueKeyword)`.
I'm happy to provide a separate pull request for this alone (along with tests) if it is needed.

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/BlockMutatorTests.cs
sbergen: I feel like it would unnecessarily cause the test to break when other aspects of the mutator change. I'm testing something very specific here, and don't think this test should care about the rest of the possible mutations.

src/Stryker.Core/Stryker.Core/Mutators/BlockMutator.cs
sbergen: `ShouldMutateLocalFunctionsInStructConstructors` breaks if this is removed. It needs to return early in in a local function within a struct constructor.

richardwerkman: Ah I get that now

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/BlockMutatorTests.cs
richardwerkman: Fair enough

richardwerkman (approved):

PR #1764, opened 2021-11-01
fix(Ignore mutations): Ignoring statement mutations is now possible

### LINKED ISSUE #1762: Statement mutation cannot be ignored entirely
**Describe the bug**
I want to migrate to Stryker 1.0.0 and I also need to keep the scores (as we lock them in), so I need to temporarily ignore the statement mutation for now. However this works weirdly.

Here's my code missing some tests:
image

Here's the report when nothing is ignored:
image

And here's what happens when I ignore the statement mutation:
image

So only one of the two statement mutations is ignored. This is unacceptable for me :(

**Expected behavior**
Ignore all the statement mutations, essentially give the same score as it was Stryker beta.

 - OS: Windows
 - Type of project: Core
 - Framework Version: 5.0
 - Stryker Version: 1.0.0

**Additional context**
Repro is here: [Repro.zip](https://github.com/stryker-mutator/stryker-net/files/7453333/Repro.zip)

richardwerkman:
Flip statement mutation to first in the enum so it will be picked first in case the statement mutator is chosen

closes #1762


rouke-broersma:
:|

PR #1771, opened 2021-11-02
Docs: aligning header sizes in the README

psfinaki:
Now they are too big for the amount of content there :D


richardwerkman:
Fair enough 😉

PR #1774, opened 2021-11-02
Updated mutator docs

psfinaki:
A few mutator related updates:
1. Fixed broken links in the docs
2. Renamed mutations to mutators in some places (I got lazy at some point)
3. Add the names of the mutations alongside the names of the mutators. I've been missing this myself and the names are not always obvious

### DIFF
--- a/docs/mutators.md
+++ b/docs/mutators.md
@@ -1,12 +1,14 @@
 ---
-title: Mutations
+title: Mutators
 sidebar_position: 40
 custom_edit_url: https://github.com/stryker-mutator/stryker-net/edit/master/docs/mutators.md
 ---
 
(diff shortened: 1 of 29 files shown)


rouke-broersma:
I think we actually wanted to standardize on Mutations instead of Mutators but I'm not 100 percent sure. @richardwerkman ?

docs/mutators.md
```diff
@@ -1,12 +1,14 @@
 ---
```
richardwerkman, line 1: Renaming this file will break the website. I'd suggest keeping this mutations for now as these docs contain both the mutators and the specific mutations

richardwerkman (changes requested):
It's good to have the actual names for the ignore functionality in the docs! Thanks

docs/mutators.md
psfinaki: Alright. I removed all the renamings, only left the _custom_edit_url_ change in that regard, I guess it should fix this issue here:
image

richardwerkman (approved):

PR #1795, opened 2021-11-12
feat(reporting): Open dashboard report with --open-report

richardwerkman:


src/Stryker.Core/Stryker.Core/Reporters/HtmlReporter/ProcessWrapper/ProcessWrapper.cs
```diff
@@ -1,18 +1,49 @@
...
+
+        public Process StartUrl(string url)
+        {
+            try
+            {
+                return Process.Start(url);
+            }
+            catch
+            {
+                // hack because of this: https://github.com/dotnet/corefx/issues/10361
```
rouke-broersma, line 28: The comments on that issue specify that this is no longer a problem, and opening a file works with the `UseShellExecute  true`. Did you actually come across a problem with this? Of so, why not use the same method for opening both?

richardwerkman: No, opening a file and a url work differently. But about this "fix" I just coppied it from somewhere so it might not be needed indeed.

rouke-broersma: Mhm as far as I understood it, the default program should automatically pick it up based on the registered URI handler for the protocol (file://, https://). Both solutions should basically do that afaik. If the first solution does not work for url's I would prefer to use the second solution for both.

richardwerkman: Yeah I tested and `UseShellExecute true` works flawless... So I don't know why I went through all these hoops only to remove this code haha

src/Stryker.Core/Stryker.Core/Reporters/HtmlReporter/ProcessWrapper/ProcessWrapper.cs
```diff
@@ -1,14 +1,15 @@
 using System.Diagnostics;
+using System.Runtime.InteropServices;
 
 namespace Stryker.Core.Reporters.HtmlReporter.ProcessWrapper
 {
     public class ProcessWrapper : IProcessWrapper
```
rouke-broersma, line 6: Perhaps we should rename this to something like `ReportOpener` since ProcessWrapper is pretty technical for what it actually does.

rouke-broersma: Nice!

PR #1796, opened 2021-11-12
chore: Add test for ignore method

### LINKED ISSUE #973: Restrict ignore methods to only method invocations
>> This will cause the `--ignore-methods` option to be quite expanded. Could you please update the documentation?

> I'm not sure if I still want this myself since I already dislike the ambiguity of given filter. The use of excluding method-calls seems appropriate since one can't really test the mutations on `ConfigureAwait` for example but excluding method-bodies (like it already does) is a whole different beast.

> Since you already accepted my other PR where I can opt-out of testing with the `ExcludeCoverage` attribute you maybe want to restrict this feature even more.

_Originally posted by @riezebosch in https://github.com/stryker-mutator/stryker-net/pull/961#issuecomment-597046635_

richardwerkman:
closes #973


rouke-broersma (changes requested):
Did you link this to the wrong issue? #973 is about only allowing `ignore-methods` to be used to ignore mutations on method invocations and no longer allow this to be used to filter method declarations. Or was that never implemented?

richardwerkman:
It seems this was never implemented or removed at some point. I tested it and it doesn't work in our current version. This test I added also proves this but I tested on a real project as well.

PR #1800, opened 2021-11-13
fix(baseline): Use regular fallback version as well as baseline fallback version

### LINKED ISSUE #1798: Fallback Disk baseline report is not looked for in the correct location
**Describe the bug**

When using
```
"reporters": [
    "progress",
    "html",
    "baseline"
],
"baseline": {
    "enabled": true,
    "provider": "Disk",
    "fallback-version": "main"
},
"since": {
    "enabled": true,
    "target": "main"
}
```

(see #1797 for details on why the baseline reporter has to be manually enabled)

1. Generate a baseline report on the `main` branch
2. Run Stryker on a branch that is not `main`
3. The baseline report can't be found, because it is looked for in the wrong directory:

`GitBaselineReporter.OnAllMutantsTested` and `DashboardMutantFilter.GetBaselineAsync` use `dashboard-compare/{projectVersion}`, but `DashboardMutantFilter.GetFallbackBaselineAsync` uses the version name directly. 

I'm not sure how this interacts with other baseline providers, so I wasn't comfortable suggesting a direct fix in a pull request.

**Logs**
I can provide logs if necessary, but I feel the issue is pretty clear.

**Expected behavior**
An existing fallback baseline report on disk should be found when it's present.

**Desktop:**
 - OS: MacOS
 - Type of project: class library
 - Framework Version: core 2.0
 - Stryker Version 1.0.1, master

rouke-broersma:
fix #1798

Technically this is a breaking change since I changed the folder paths but I don't think this will have a huge impact as it will simply mean regenerating the baseline once.

@richardwerkman do you agree with allowing this 'breaking' change in a fix?
@sbergen since you actually use this feature, do you agree with changing the folder path from `dashboard-compare` to `baseline`?

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -181,7 +181,7 @@ The levels are:
 | Arithmetic Operators | Basic|
 | Block (not yet implemented) | Basic|
 | Equality Operators | Standard |
-| flagean Literals | Standard|
+| Boolean Literals | Standard|
 | Assignment statements | Standard |
 | Collection initializer | Standard |
 | Unary Operators | Standard |
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -9,10 +9,10 @@ custom_edit_url: https://github.com/stryker-mutator/stryker-net/edit/master/docs
 You run stryker from the test project directory.
 
 On some dotnet core projects stryker can run without specifying any custom configuration. Simply run `dotnet stryker` to start testing.  
-On dotnet framework projects the solution path argument is always required. Run at least `dotnet stryker --solution <solution-path>` or specify the solution file path in the config file to start testing. See [solution](#solution-<path>).
+On dotnet framework projects the solution path argument is always required. Run at least `dotnet stryker --solution <solution-path>` or specify the solution file path in the config file to start testing. See [solution](#solution-path).
 
 ## Use a config file
-When using Stryker regularly we recommend using a config file. This way you won't have to document how to run Stryker, you can save the config file in version control. To use a config file create a file called `stryker-config.json` in the (unit test) project folder and add a configuration section called stryker-config. Alternatively use [init](#init-<bool>)
+When using Stryker regularly we recommend using a config file. This way you won't have to document how to run Stryker, you can save the config file in version control. To use a config file create a file called `stryker-config.json` in the (unit test) project folder and add a configuration section called stryker-config.
 
 Example `stryker-config.json` file:
 ``` javascript
@@ -24,14 +24,6 @@ Example `stryker-config.json` file:
     }
 }
 ```
-
-### `init` <`flag`>
-
-Default: `false`  
-Command line: `--init`  
-Config file: `N/A`  
-
-Creates a stryker config file with default options. Any options passed in the cli will override the defaults.
 
 ### `config-file` <`path`>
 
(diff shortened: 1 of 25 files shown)


sbergen:
> @sbergen since you actually use this feature, do you agree with changing the folder path from `dashboard-compare` to `baseline`?

I wouldn't say I use the feature yet, as I was just trying to set stuff up, but ran into issues. But I do find the Disk provider the easiest way to test the baseline functionality locally. And in this branch, it does work better :)

In any case, the path already contains `StrykerOutput/Baselines`, so I don't understand why it's necessary to have one more layer of directories here. But I did find `dashboard-compare` confusing, and think `baseline` is a better option, if an intermediate directory is really required.

rouke-broersma:
`StrykerOutput/Baselines` is iirc the base path for the azure file share provider, all other providers use the base directory + baseline. This is indeed a bit redundant.

sbergen:
> `StrykerOutput/Baselines` is iirc the base path for the azure file share provider, all other providers use the base directory + baseline. This is indeed a bit redundant.

https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/Baseline/Providers/DiskBaselineProvider.cs#L16

rouke-broersma:
> > `StrykerOutput/Baselines` is iirc the base path for the azure file share provider, all other providers use the base directory + baseline. This is indeed a bit redundant.
> 
> https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/Baseline/Providers/DiskBaselineProvider.cs#L16

🙈 I'll see if I can fix that up

rouke-broersma:
@sbergen Can you see if you would be happy with how the feature works now?

sbergen:
Now I'm getting either
`Project version cannot be empty when baseline is enabled`
or
`Fallback version cannot be set to the same value as the current project version, please provide a different fallback version`
when trying to generate an initial baseline.

I think we still need to support generating an initial baseline from the branch you are currently on, and I can't think of a trivial way to do that the way things are now.

Also, you got the issue reference wrong in your first comment, it should be #1798 :)

rouke-broersma:
> Now I'm getting either `Project version cannot be empty when baseline is enabled` or `Fallback version cannot be set to the same value as the current project version, please provide a different fallback version` when trying to generate an initial baseline.
> 
> I think we still need to support generating an initial baseline from the branch you are currently on, and I can't think of a trivial way to do that the way things are now.
> 
> Also, you got the issue reference wrong in your first comment, it should be #1798 :)

If you set the project version and not the fallback version I think this should work. This implicitly uses the since target version which might be the same as the current project version but I think this case is allowed now, but only when this is implicitly through target version, not when explicitly defined. This is a bit confusing..

Not allowed:
```
project version -> master
since target -> master
fallback version -> master
```

Allowed:
```
project version -> master
since target -> master
fallback version -> since target (master)
```

I think this requirement could be relaxed to make it easier to generate the baseline for master/main, but then we can't warn the user about incorrect configuration.

sbergen:
> If you set the project version and not the fallback version I think this should work. This implicitly uses the since target version which might be the same as the current project version but I think this case is allowed now, but only when this is implicitly through target version, not when explicitly defined. This is a bit confusing..
> 
> Not allowed:
> 
> ```
> project version -> master
> since target -> master
> fallback version -> master
> ```
> 
> Allowed:
> 
> ```
> project version -> master
> since target -> master
> fallback version -> since target (master)
> ```
> 
> I think this requirement could be relaxed to make it easier to generate the baseline for master/main, but then we can't warn the user about incorrect configuration.

Ok, I got it to work with these instructions now, and the original issue seems to be fixed: the baseline report is found in the correct location when running a second time. But I do agree the situation with these restrictions is confusing. If `since` is enforced, shouldn't we then also enforce using the since target as the fallback for baseline? Couldn't you otherwise get invalid results?

And also, regarding versioning, I think specifying that the project follows semver, **excluding** experimental features would make sense.

PR #1801, opened 2021-11-13
feat(options): Reintroduce threshold high and threshold low from the commandline

### LINKED ISSUE #1779: Allow the break threshold to be higher than others
Here's my "customer case" :)

Currently we execute Stryker in different ways:
1) We have nightly executions for all the projects (about 200 of them right now) to have the "big picture"
2) We execute it in PRs for changed projects to see if the changes don't lower the current mutation score (we have an original mechanism of keeping track of it :D) 

We also have some "quality standards", something like mutation score above 85 is good, 50 to 85 is disturbing (and someone should jump on addressing this), below 50 is very very bad - our systems won't allow the package to be normally released basically.

Now we've been getting by with one `stryker-config.json` so far, where we have reporters, ignored methods and so on but where we don't set any thresholds whatsoever. And then:
1) For nightly executions, we would set the scores in the CLI like `--threshold-high 85 --threshold-high 50 --threshold-break 0` - to have the nice picture
2) For PR executions, we would set them as `--threshold-high 100 --threshold-low 100 --threshold-break <current mutation score>` - just to prevent PRs from being merged if not sufficiently tested
---
Now, Stryker 1.0 removes the high and low thresholds from the CLI and makes us have them in the config. This leaves us with too options:
1. Set `high: 100, low: 100` in the current config - and lose in clarity of the big picture after nightly runs (the numbers will be basically only green or red there)
2. Have two configs, one for nightly with `high: 85, low: 50` and one for PRs without thresholds at all - and have all the other settings duplicated

The above options are not the end of the world but I thought - why does Stryker even break when the break threshold is higher than the high one? It makes sense that "low" must be lower than or equal to "high", but "break" has a bit different objective so probably doesn't need to an obstacle here - for those who know what they're doing.

We can maybe write a warning instead.

**Additional context**
I can do that :D

rouke-broersma:
fix #1779 

@psfinaki and @richardwerkman Is this solution acceptable to you?

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -286,7 +286,7 @@ Command line: `[-b|--break-at] 40`
 Config file: See [thresholds](#thresholds-object)
 
 Must be less than or equal to threshold low.  
-When threshold break is set to anything other than 0 Stryker will exit with a non-zero code. This can be used in a CI pipeline to fail the pipeline when you mutation score is not sufficient.  
+When threshold break is set to anything other than 0 Stryker will exit with a non-zero code. This can be used in a CI pipeline to fail the pipeline when your mutation score is not sufficient.  
 
 ### `threshold-high` <`number`>
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -267,7 +267,7 @@ Default: `{ high: 80, low: 60, break: 0 }`
 Command line: `N/A`  
 Config file: `"thresholds": { "high": 80, "low": 60, "break": 0 }`
 
-Configure the mutation score thresholds for your project. Thresholds should be a number between 0 and 100. Thresholds can all have the same value except for 0. Threshold high cannot be 0.
+Configure the mutation score thresholds for your project. Thresholds should be a number between 0 and 100. Thresholds can have the same value.
 
 Threshold calculations in order:
 - `mutation score >= threshold-high`: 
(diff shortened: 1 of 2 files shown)


docs/configuration.md
```diff
@@ -277,17 +277,32 @@ Threshold calculations in order:
...
 ### `break-at` <`number`>
 
 Default: `0`  
 Command line: `[-b|--break-at] 40`  
-Config file: `"thresholds": { "break": 0 }`
+Config file: See [thresholds](#thresholds-object)
 
-When threshold break is set to anything other than 0 Stryker will exit with a non-zero code. This can be used in a CI pipeline to fail the pipeline when you mutation score is not sufficient. Must be less than or equal to threshold low.
+Must be less than or equal to threshold low.  
+When threshold break is set to anything other than 0 Stryker will exit with a non-zero code. This can be used in a CI pipeline to fail the pipeline when you mutation score is not sufficient.
```
psfinaki, line 289:
```suggestion
When threshold break is set to anything other than 0 Stryker will exit with a non-zero code. This can be used in a CI pipeline to fail the pipeline when your mutation score is not sufficient.  
```

psfinaki:
```suggestion
When threshold break is set to anything other than 0 and the actual mutation score is lower than that, Stryker will exit with a non-zero code. This can be used in a CI pipeline to fail the pipeline when you mutation score is not sufficient.
```

psfinaki (approved):

docs/configuration.md
```diff
@@ -277,17 +277,32 @@ Threshold calculations in order:
...
 ### `break-at` <`number`>
 
 Default: `0`  
 Command line: `[-b|--break-at] 40`  
-Config file: `"thresholds": { "break": 0 }`
+Config file: See [thresholds](#thresholds-object)
 
-When threshold break is set to anything other than 0 Stryker will exit with a non-zero code. This can be used in a CI pipeline to fail the pipeline when you mutation score is not sufficient. Must be less than or equal to threshold low.
+Must be less than or equal to threshold low.  
+When threshold break is set to anything other than 0 Stryker will exit with a non-zero code. This can be used in a CI pipeline to fail the pipeline when your mutation score is not sufficient.
```
rouke-broersma, line 289:
```suggestion
When threshold break is set to anything other than 0 and the mutation score is lower than the threshold Stryker will exit with a non-zero code. This can be used in a CI pipeline to fail the pipeline when your mutation score is not sufficient.  
```

Issue #1565, opened 2021-05-14
Report on tests in stryker report (json/html)


rouke-broersma:
Update to the latest mutation testing elements
Gather test run information and provide this data to the report

https://github.com/stryker-mutator/mutation-testing-elements/pull/893

**Tasks**

- [ ] Disable bail from commandline #1784
- [x] Create json report components to hold test data #1802
- [x] Create project components to hold test data #1820
- [ ] Fill test project components with data #1850
- [ ] Save for each mutant which unit test killed it

JSmits-utwente:
For my graduation assignment I need a more complete JSON report from Stryker.NET according to the mutation-testing-elements [schema](https://github.com/stryker-mutator/mutation-testing-elements/blob/master/packages/report-schema/src/mutation-testing-report-schema.json). In particular I need to know for each mutant the properties `killedBy` and `coveredBy`. For this each test case must be given an ID in the report as well. 

If I'm not wrong closing this issue should give me the information I need. How much time / difficulty do you estimate it takes to implement this? If it's not too bad I may pick this up myself.

rouke-broersma:
@JSmits-utwente AFAIK we should already have all this information, we just need to make it available in the report json.

rouke-broersma:
We should make disable-bail available as a cli option so that users can on-demand request a html report with all test information.

PR #1802, opened 2021-11-13
chore(report on tests): introduce json report testfiles

rouke-broersma:
Related to #1565


src/Stryker.Core/Stryker.Core/Reporters/Json/SourceFiles/SourceFile.cs
```diff
@@ -0,0 +1,44 @@
...
+        public string Source { get; init; }
+        public ISet<JsonMutant> Mutants { get; init; }
+
+        public SourceFile()
+        {
+        }
+
+        public SourceFile(ReadOnlyFileLeaf file, ILogger logger = null)
+        {
+            logger ??= ApplicationLogging.LoggerFactory.CreateLogger<SourceFile>();
```
richardwerkman, line 20: Why is the logger created here and not in constructor?

rouke-broersma: This is the constructor? Also this was a rename, it has always been like this.

richardwerkman: Whoops, didn't pay attention sorry

richardwerkman (approved):

PR #1804, opened 2021-11-14
fix(since): Mutants ignored from source code comment are no longer re-enabled by the since filter

### LINKED ISSUE #1760: When using --since:<commitish> filtered mutants in changed files are run anyway
**Describe the bug**
In my project I disabled LINQ mutations for one location with `// Stryker disable once Linq`, because the mutant there is unkillable.
When now running `dotnet stryker` while in the corresponding test project, the mutant correctly gets ignored.
If I run `dotnet stryker --since:c76f6c8f53ba30efb3ff081c95867977548d2e2f` instead, the mutant is not ignored because the file has changed.

**Logs**
[WithoutSince.txt](https://github.com/stryker-mutator/stryker-net/files/7449019/WithoutSince.txt)
[WithSince.txt](https://github.com/stryker-mutator/stryker-net/files/7449021/WithSince.txt)

**Expected behavior**
I would expect that filtered mutants are ignored, even when they are in an changed file.

 - OS: Windows
 - Core
 - Framework Version net5.0
 - Stryker Version 1.0.0

**Additional context**
All code is available in the repository https://github.com/Deneas/MutationTestingExample, although be aware that he commands in the Readme are from before 1.0.0 :)

rouke-broersma:
Fix #1760

### DIFF
--- a/src/Stryker.Core/Stryker.Core/MutantFilters/BroadcastMutantFilter.cs
+++ b/src/Stryker.Core/Stryker.Core/MutantFilters/BroadcastMutantFilter.cs
@@ -17,7 +17,7 @@ namespace Stryker.Core.MutantFilters
 
         public IEnumerable<Mutant> FilterMutants(IEnumerable<Mutant> mutants, ReadOnlyFileLeaf file, StrykerOptions options)
         {
-            IEnumerable<Mutant> mutantsToTest = mutants.Where(m => m.ResultStatus is not MutantStatus.Ignored);
+            var mutantsToTest = mutants.Where(m => m.ResultStatus is not MutantStatus.Ignored);
 
             foreach (var mutantFilter in MutantFilters)
             {
--- a/src/Stryker.Core/Stryker.Core/MutantFilters/BroadcastMutantFilter.cs
+++ b/src/Stryker.Core/Stryker.Core/MutantFilters/BroadcastMutantFilter.cs
@@ -1,4 +1,4 @@
-﻿using Stryker.Core.Mutants;
+using Stryker.Core.Mutants;
 using Stryker.Core.Options;
 using Stryker.Core.ProjectComponents;
 using System.Collections.Generic;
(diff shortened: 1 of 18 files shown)


src/Stryker.Core/Stryker.Core/MutantFilters/BroadcastMutantFilter.cs
```diff
@@ -8,42 +8,33 @@ namespace Stryker.Core.MutantFilters
...
+                foreach (var skippedMutant in mutantsToTest.Except(remainingMutantsToTest))
                 {
                     skippedMutant.ResultStatus = MutantStatus.Ignored;
                     skippedMutant.ResultStatusReason = $"Removed by {mutantFilter.DisplayName}";
                 }
 
-                mutantsToTest = extraMutantsToTest;
-            }
-
-            // Diff filter goes last if it is enabled
```
rouke-broersma, line 40: This should never have been the job of the broadcast mutant filter, this is now solved by making the sorting of the mutant filters explicit when created in the mutant filter factory by using a sortedset.

src/Stryker.Core/Stryker.Core/MutantFilters/BroadcastMutantFilter.cs
```diff
@@ -8,42 +8,33 @@ namespace Stryker.Core.MutantFilters
 {
     public class BroadcastMutantFilter : IMutantFilter
     {
+        public MutantFilter Type => MutantFilter.Broadcast;
```
richardwerkman, line 11: Why do the filters need a Type now?

rouke-broersma: So we can use their index as the sort value (enums are ints so their position decides their order in a sorted list), this guarantees the order they are executed on.

richardwerkman: Maybe we should add a comment to the enum to make sure nobody changes the order :)

rouke-broersma: Yep good idea

src/Stryker.Core/Stryker.Core/MutantFilters/MutantFilter.cs
```diff
@@ -0,0 +1,21 @@
+namespace Stryker.Core.MutantFilters
+{
+    public enum MutantFilter
+    {
```
rouke-broersma, line 4:
```suggestion
    // Filters are executed in the order they appear in in this enum. If you change this order you change the order or filter execution
    public enum MutantFilter
    {
```

src/Stryker.Core/Stryker.Core/MutantFilters/BroadcastMutantFilter.cs
rouke-broersma: The comment is added to the MutantFilter enum

Issue #1809, opened 2021-11-18
Nightly pipeline fails with stryker on stryker


richardwerkman:
**Describe the bug**
Since stryker v1.0.0 we forgot to update the pipeline with our breaking changes. Now the pipeline fails every night.

PR #1815, opened 2021-11-26
chore: Fix stryker configs and pipeline

richardwerkman:
closes: #1809

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/stryker-config.json
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/stryker-config.json
@@ -6,7 +6,6 @@
       "!**/FSharp*.*"
     ],
     "project-info": {
-      "name": "github.com/stryker-mutator/stryker-net",
       "module": "core"
     }
   }
--- a/src/Stryker.Core/Stryker.Core.UnitTest/stryker-config.json
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/stryker-config.json
@@ -5,9 +5,9 @@
       "!Logging/**",
       "!**/FSharp*.*"
     ],
-    "dashboard-project": "github.com/stryker-mutator/stryker-net",
-    "dashboard-version": "master",
-    "dashboard-fallback-version": "master",
-    "dashboard-module": "core"
+    "project-info": {
+      "name": "github.com/stryker-mutator/stryker-net",
+      "module": "core"
+    }
   }
 }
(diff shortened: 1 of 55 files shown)


src/Stryker.sln
```diff
@@ -17,6 +17,11 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Stryker.RegexMutators", "St
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Stryker.RegexMutators.UnitTest", "Stryker.RegexMutators\Stryker.RegexMutators.UnitTest\Stryker.RegexMutators.UnitTest.csproj", "{7C8C7670-4636-4C56-9E3E-22B987337EF2}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{68DC3E8E-7CC2-4953-BAFC-08EFBF9F3307}"
+	ProjectSection(SolutionItems) = preProject
+		..\azure-pipelines.yml = ..\azure-pipelines.yml
```
rouke-broersma, line 22: Is this necessary? If so, why only the main pipeline file?

src/Stryker.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.28729.10
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31903.59
```
rouke-broersma, line 4: Is this backwards compatible with vs2019?

src/Stryker.Core/Stryker.Core.UnitTest/stryker-config.json
```diff
@@ -5,9 +5,9 @@
       "!Logging/**",
       "!**/FSharp*.*"
     ],
-    "dashboard-project": "github.com/stryker-mutator/stryker-net",
-    "dashboard-version": "master",
-    "dashboard-fallback-version": "master",
-    "dashboard-module": "core"
+    "project-info": {
+      "name": "github.com/stryker-mutator/stryker-net",
```
rouke-broersma, line 9: This is not needed, we automatically determine the project name from the project using SourceLink.

This is set using Reproducible Builds here: https://github.com/stryker-mutator/stryker-net/blob/master/src/Directory.Build.props#L15

src/Stryker.sln
richardwerkman: I've opened with vs2019 and it works

src/Stryker.sln
richardwerkman: Well no it's not necessary. I'm just used to the pipeline file being included in the solution. But we're missing the templates now. I could add them too?

rouke-broersma: Could you then also add the .editorconfig, Directory.Build.props and Directory.Build.targets? They are currently referenced in the Stryker.CLI project but it makes more sense as a solution level item imo. Perhaps even the docs folder as well?

richardwerkman: I can't add a whole folder... Adding individual docs files could work but we will surely forget to add new doc files in the future

rouke-broersma: That is a shame. In that case don't add them.

PR #1818, opened 2021-11-26
feat(Html report): New diff view in HTML report

richardwerkman:


src/Stryker.Core/Stryker.Core/Stryker.Core.csproj
```diff
@@ -69,6 +69,8 @@
   </ItemGroup>
 
   <ItemGroup>
+    <None Remove="Reporters\HtmlReporter\Files\dist\mutation-test-elements.js" />
+    <None Remove="Reporters\HtmlReporter\Files\mutation-report.html" />
```
rouke-broersma, line 73:
```suggestion
```

This seems redundant to me since they are already referenced at https://github.com/stryker-mutator/stryker-net/pull/1818/files#diff-5f45f5f909d8e285788be5cd8863594e906d58139e275a9583ca0826f50d1516R31-R32

rouke-broersma:
This is an update from 1.7.5 to 1.7.6. Why is a patch version change in upstream a minor version change for us?

PR #1819, opened 2021-11-26
fix(Project filter): Project filter works with both forward and backward slashes

### LINKED ISSUE #1769: Project not found when using / instead of \.
**Describe the bug**
In my config, when I use `"project": "CK.MQTT.Client/CK.MQTT.Client.csproj"`, stryker doesn't find the project. Using a backslash solve the issue.

**Expected behavior**
I can use both slash or backslash in the config.

 - OS: Windows
 - Type of project: Library
 - Framework Version: .NET 5
 - Stryker Version: 1.0.0

**Additional context**
Looks like a regression because it was working before 1.0.0.

richardwerkman:
closes #1769


rouke-broersma:
I think this is a problem with the json config because json uses `\` as an escape character so using `\` in the project name filter can result in characters being escaped accidentally.

richardwerkman:
It shouldn't. Our utils only replaces `\\` and not `\`. So any escaped character should stay escaped. If someone ever uses an escaped character in a project filename. Never seen that

PR #1829, opened 2021-12-10
Fix(Mutation score): NaN score on no mutants tested is now 0

### LINKED ISSUE #1825: Stryker reports NaN instead of 0% for code without real tests
**Describe the bug**
Accidentally noticed this when playing with another bug :D 

The simplest repro is to just have an empty test:
image

Then Stryker logs this:
image

The report itself correctly shows 0 though:
image

Also, apparently if I "cover" this test (with shit :D), Stryker does report 0%:
image

image

**Expected behavior**
0% in Stryker logs.

**Desktop:**
 - OS: Windows
 - Type of project: Core
 - Framework Version: .NET 6
 - Stryker Version: 1.1.0

**Additional context**
Repro: [Repro.zip](https://github.com/stryker-mutator/stryker-net/files/7620429/Repro.zip)

richardwerkman:
closes #1825


src/Stryker.Core/Stryker.Core/ProjectComponents/ProjectComponent.cs
```diff
@@ -60,6 +60,12 @@ public double GetMutationScore()
             double totalCount = TotalMutants.Count();
             double killedCount = DetectedMutants.Count();
 
+            if (totalCount == 0)
+            {
+                // prevent devide by zero
```
rouke-broersma, line 65: We should not prevent a divide by zero as NaN **is** a valid value for the mutation score if there are no mutants. The problem is that we are counting the detected wrong.

src/Stryker.Core/Stryker.Core/ProjectComponents/ProjectComponent.cs
```diff
@@ -57,10 +60,10 @@ public abstract class ProjectComponent<T> : ProjectComponent
         /// <returns>decimal between 0 and 1 or null when no score could be calculated</returns>
```
psfinaki, line 60: Looks like cannot be null anymore?

PR #1836, opened 2021-12-14
fix(Load preprocessor symbols): Correctly load preprocessor symbols so they can be used

### LINKED ISSUE #1828: Declare standard predefined symbols when building mutated assemblies
See issue #1824 

Stryker needs to declare predefined symbols when building the mutated assembly.

Those symbols can be explicitly declared in the (VS/VC) project settings, but this defeats the purpose of predefined symbols.

**Additional context**
It appears Roslyn does not add them automatically. So the code may have to declare them according the target framework version. Any official method would obviously be preferable.

richardwerkman:
Also add nice hint for opening the report
closes #1828

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Reporters/DashboardReporter.cs
+++ b/src/Stryker.Core/Stryker.Core/Reporters/DashboardReporter.cs
@@ -39,13 +39,17 @@ namespace Stryker.Core.Reporters
 
             if (reportUri != null)
             {
-                _logger.LogDebug("Your stryker report has been uploaded to: \n {0} \nYou can open it in your browser of choice.", reportUri);
-                _consoleWriter.Write(Output.Green($"Your stryker report has been uploaded to: \n {reportUri} \nYou can open it in your browser of choice."));
-
                 if (_options.ReportTypeToOpen == Options.Inputs.ReportType.Dashboard)
                 {
                     _processWrapper.Open(reportUri);
                 }
+                else
+                {
+                    _consoleWriter.Write(Output.Cyan("Hint: by passing \"--open dashboard or -o dashboard\" the report will open automatically once stryker is done."));
+                }
+
+                _logger.LogDebug("Your stryker report has been uploaded to: \n {0} \nYou can open it in your browser of choice.", reportUri);
+                _consoleWriter.Write(Output.Green($"Your stryker report has been uploaded to: \n {reportUri} \nYou can open it in your browser of choice."));
             }
             else
             {
--- a/src/Stryker.Core/Stryker.Core/Reporters/DashboardReporter.cs
+++ b/src/Stryker.Core/Stryker.Core/Reporters/DashboardReporter.cs
@@ -45,7 +45,7 @@ namespace Stryker.Core.Reporters
                 }
                 else
                 {
-                    _consoleWriter.Write(Output.Cyan("Hint: by passing \"--open dashboard or -o dashboard\" the report will open automatically once stryker is done."));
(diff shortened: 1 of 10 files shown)


src/Stryker.Core/Stryker.Core/Reporters/DashboardReporter.cs
```diff
@@ -39,13 +39,17 @@ public void OnAllMutantsTested(IReadOnlyProjectComponent reportComponent)
...
-                _logger.LogDebug("Your stryker report has been uploaded to: \n {0} \nYou can open it in your browser of choice.", reportUri);
-                _consoleWriter.Write(Output.Green($"Your stryker report has been uploaded to: \n {reportUri} \nYou can open it in your browser of choice."));
-
                 if (_options.ReportTypeToOpen == Options.Inputs.ReportType.Dashboard)
                 {
                     _processWrapper.Open(reportUri);
                 }
+                else
+                {
+                    _consoleWriter.Write(Output.Cyan("Hint: by passing \"--open dashboard or -o dashboard\" the report will open automatically once stryker is done."));
```
psfinaki, line 48: Is this correct? Per [this](https://github.com/stryker-mutator/stryker-net/pull/1795/files#diff-17ed18489a956f326ec0fe4040850c5bc9261d4631fb42da4c52891d74a59180) it should be something like `--open-report:dashboard`.

src/Stryker.Core/Stryker.Core/Reporters/HtmlReporter/HtmlReporter.cs
```diff
@@ -43,14 +43,17 @@ public void OnAllMutantsTested(IReadOnlyProjectComponent reportComponent)
...
-
             if (_options.ReportTypeToOpen == Options.Inputs.ReportType.Html)
             {
                 _processWrapper.Open(reportUri);
+            } else
+            {
+                _consoleWriter.Write(Output.Cyan("Hint: by passing \"--open or -o\" the report will open automatically once stryker is done."));
             }
+
+            _consoleWriter.WriteLine(Output.Green($"\nYour html report has been generated at: \n " +
```
psfinaki, line 54: Do you need spaces before and after the break?

src/Stryker.Core/Stryker.Core/Reporters/DashboardReporter.cs
psfinaki:
```suggestion
                    _consoleWriter.Write(Output.Cyan("Hint: by passing \"--open dashboard or -o dashboard\" the report will open automatically once Stryker is done."));
```

psfinaki (approved):
Put a few small notes BUT the important part is that symbols are now working correctly, manually double-checked this against a repro from the branch :)

If compiled against .NET 6:
image

If compiled against .NET 5:
image

So just what we need and no more weird behavior. Great job @richardwerkman! We'll benefit from that in 6 projects so would be awesome to see this released ASAP :)

src/Stryker.Core/Stryker.Core/Reporters/DashboardReporter.cs
richardwerkman: My bad 🙈 apparently I don't know Stryker as well as you do haha

richardwerkman:
@psfinaki v1.2.1 has been released!

psfinaki:
Thanks for the release, will try this out today :)

As for the new diff, sure it's very slick! I mean it's not that I was extremely annoyed by the previous version so I just take this as a facelift. I used to use that "show all" thing occasionally, but it wasn't that important. So yeah good stuff!

PR #1857, opened 2022-01-10
Add mutant diagnostic option

dupdob:
Add a diagnose mode to help understand result for a single mutation

WIP


dupdob:
hi everyone,
quick update on this record long PR.
As of today, I have a fully working diagnostic feature which implements a two phases diagnoses
# 1st Phase
Main phase
1. it works against one mutant of the user choosing (identified via its numerical id)
2. the mutation is then tested three different ways:
   1. 'normally': as it would by running a normal Stryker session, with other mutations in the same session (if possible)
   2. 'isolated': tested against the covering test(s) without any other mutations
   3. 'full': tested against all tests
3. if all results are the same, diagnostic ends and Stryker states there is no detectable issue; it means there is probably no actual issue. Any issue will probably be linked to Stryker not being able to run some tests through VsTest, or some misconfiguration from the user part
4. If the mutant is killed in 'full' but not in the other modes, it means the problem comes from coverage analysis. Stryker will then suggest the user to add a dedicated Stryker comment to force 'full mode' for this mutation. Either the mutations is triggered in a static context and Stryker was not able to identify it as such, or there is a bug in coverage analysis 
5. If the mutant is 'killed' in '2' but not in '1': there may be several causes so Stryker enter then a secondary diagnosis phase

# 2nd Phase
The 2nd phase is only triggered when the result changes between 'normal' and 'isolated'. Stryker then:
1. Try to find a conflicting mutations (with the diagnosed one) via binary search. A conflicting mutant is the one that when tested at the same than the diagnosed mutant change the test result (for the diagnose mutant)
2. If more than one mutations appear to be in conflict, Stryker assumes the diagnosed one is the responsible for the conflict
3. Then Stryker tries to identify which test trigger the detection failure, this information can then be used (by the user) to identify a non isolated test. See below for when this is useful
4. Stryker then suggests to add a stryker comment to be place before the conflicting mutant so that it is tested in isolation from other mutations (i.e. in its own test session) as well as the name of the test that triggered the conflict.

This situation can happen when a test alters some global constant but does not restores it when the test fails. Eg:
```csharp
  MyClass.SpecialMode = true;
  //... some setup
  Assert.AreEqual(expectedResult, sut.GetResult());
  MyClass.SpecialMode = false; // if the assertion fails, this line is skipped
```
Then the user should add a `try...finally` construct to protect this.

# Why it is still in draft mode
While testing the feature on some suspicious survivors, I got a diagnostic result that I could not justify looking at NFluent code. Stryker identify a conflict that does not make sens when looking at the code; on the other hand, it happens with mutations that can have large impacts on the test.
I realized then that I need a third diagnosis session. This session will capture the mutation trace, i.e., which mutations are executed, and in which order, when running the suspicious test in 'normal' and in 'isolated' modes. Any difference will be the signal of the behavioral change and a hint of where the user should look to understand why this is happening.

# Work status
I have implemented the trace capture, but I am kinda stuck at updating the various internal interfaces to support this new mode. 
On one hand, the diagnosis feature works properly and does fix anormal results via the proper comments. As such, the feature could be delivered as is; but on the other hand, I feel I ran into this inexplicable survivor a bit soon for my test, and I am worried it could happen to other as well, or that it may be related to some Stryker bug I simply failed to identify. Then suggesting I should keep working on this as is.

# Conclusion
While writing this, I realize that moving forward with the feature as-is is the best approach: it already delivers value, it is a useful diagnosis tool and it would help users. Furthermore, releasing it increase the chance of getting feedback from users that could help me understand how the feature should evolve, if necessary.
Finally, this remains an advanced feature that, at first, users will only discover when we point them to it, which reduces the risk of getting a lot of new issues due to this.

so, what do you think?

rouke-broersma:
I don't think there's anything wrong with introducing the feature as it currently works, and accepting that there are edge cases that cannot be solved by it.

dupdob:
I close it for now. I see value in this, but the fact is that nobody raised any diagnosis related question and this is hard work to maintain it.

Issue #1855, opened 2022-01-10
Stryker fails for self-contained apps


psfinaki:
**Describe the bug**
We have some apps that are published in a self-contained [manner](https://docs.microsoft.com/en-us/dotnet/core/deploying/#publish-self-contained) (and that's the future!). That's basically something like this in the app props:
```xml
<SelfContained>true</SelfContained>
<RuntimeIdentifier>win-x64</RuntimeIdentifier>
```

**Starting with 1.2.1**, Stryker breaks when running against this, with a preliminary warning to report things to GitHub:
<img width="878" alt="image" src="https://user-images.githubusercontent.com/5451366/148820848-6e6f7d6a-4ef7-476a-95a0-2500908801e1.png">

**Logs**
Logs: [log-20220110.txt](https://github.com/stryker-mutator/stryker-net/files/7841203/log-20220110.txt)
Repro: [Repro.zip](https://github.com/stryker-mutator/stryker-net/files/7841212/Repro.zip)

**Expected behavior**
Smooth execution.

**Desktop**
 - OS: Windows
 - Type of project: .NET Core
 - Framework Version: .NET 6
 - Stryker Version: 1.0.1

**Additional context**
So apparently this is caused by the latest fix in Buildalyzer, 1.2.0 work all right. _Most probably_ on the Buildalyzer side though - I reverted locally the changes in the Initialisation folder from that commit and was still getting the error.

On a bright side... What this means in practice is that of 134 projects we run Stryker against, 1.2.0 failed for 5 projects that had problems with preprocessor symbols whereas 1.2.1 fails for 2 self-contained projects. So it's kind of progress :D

rouke-broersma:
Awesome..! :p I think we've seen this error before but have so far have not been able to figure out why. Curious what changed in buildalyzer to break this.

psfinaki:
Yeah big codebase exposes all kinds of edge cases :D

richardwerkman:
Glad to hear I didn't break anything (as far as we know now 😄)

PR #1865, opened 2022-01-17
Potential fix for handling self-contained apps

psfinaki:
#1855 

See discussion [here](https://github.com/daveaglick/Buildalyzer/pull/194).

So technically this is not a pinpoint solution for the original problem, in that it fixes the problem in quite a dramatic way. 
On the other hand, from what I understand, similar problems were encountered before, probably for the same reason. So maybe it's worth applying as a prevention.

Giving that you guys here know quite a lot of stuff about MSBuild and I don't, this PR is partially meant as a discussion :) thoughts?

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/TargetFrameworkResolutionTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/TargetFrameworkResolutionTests.cs
@@ -1,6 +1,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using Buildalyzer;
+using Buildalyzer.Environment;
 using Moq;
 using Shouldly;
 using Stryker.Core.Initialisation;
@@ -24,7 +25,7 @@ namespace Stryker.Core.UnitTest.Initialisation
                 .Returns(projectAnalyzerMock.Object);
 
             projectAnalyzerMock
-                .Setup(m => m.Build())
+                .Setup(m => m.Build(It.IsAny<EnvironmentOptions>()))
                 .Returns(analyzerResultsMock.Object);
 
             analyzerResultsMock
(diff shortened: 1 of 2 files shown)


rouke-broersma:
Setting design time to false in all cases will impact the performance of Stryker in all cases so I don't think that is a viable solution. We should probably detect whether or not we are in a self-contained scenario and apply the workaround if necessary. But this feels like something that should be in Buildalyzer and not in Stryker. It seems like @daveaglick agrees and is open to discussion about the proper fix for this problem. It seems to me like the solution would be exactly the same whether it is applied in Buildalyzer or in Stryker so I would say fixing it in Buildalyzer has the most impact. It would fix the problem for all consumers.

psfinaki:
Yes, that's true, my point was to ask if this change will benefit Stryker in any other way. I mean the drawback is clear (some performance) but the extra information got from the build might be useful somehow.

If not, that yes ignore this PR, I'm working on fixing the actual problem in Buildalyzer anyway.

rouke-broersma:
Don't know if there is any benefit to doing a full build so I can't answer that. We already do a full build after buildalyzer using either msbuild or dotnet build but that's just to test if the project is actually buildable. We might be able to replace that first build by using buildalyzer for everything? @richardwerkman what do you think

daveaglick:
The discussion about this has been ongoing over at Buildalyzer, but just some quick comments to close the loop here...

> Setting design time to false in all cases will impact the performance of Stryker in all cases so I don't think that is a viable solution.

I'd also suggest we try to figure out how to continue using design-time builds. They can be an order-of-magnitude faster in some cases, and even though this has been a tricky one to solve, I have no doubt we'll find a way to perform design-time builds on self-contained apps given that Visual Studio can do it.

> But this feels like something that should be in Buildalyzer and not in Stryker.

I totally agree - being able to correctly analyze self-contained vs. non-self-contained applications isn't unique to Stryker and so I think it absolutely should be handled as well as we can upstream in Buildalyzer.

> Also: I know more about MSBuild than I care to, which is just enough to know that I don't want to know more, but also that I probably should know more for the benefit of Stryker

It me :)

rouke-broersma:
Closed in favor of #1868

PR #1866, opened 2022-01-18
Miscellaneous fixes and improvments

dupdob:
1. Block and statement mutations may now be ignored by IgnoredMethod option. Technically, statement mutations are ignored if the original code is an invocation of an 'ignored' method; block mutations are ignored if they only contain invocations of ignored method (or blocks of).
2. Fix exceptions when rolling back some expression body to statement body conversion. Technically, mutations are (now) rolled back with a depth first approach
3. Fix breaking `return default;` added to anonymous lambda that do not return any value. Technically, there is now a check to verify if the lambda contains a `return xxx` statement.-; if not, the lambda is assumed as `void`. This removes all related compilation errors but it should not have any impact on overall result.
4. Fix compilation errors with switch case due to block mutations. Technically, `BlockMutator` no longer mutates `case` related blocks. As this problem triggers safe mode, it should result in having more mutations tested for impacted code base.
5. (Definitive?) Fix for NREs linked to progress bar. Technically, the number of ticks is always at least '1', even when there is no mutant.

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/IgnoredMethodMutantFilterTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/IgnoredMethodMutantFilterTests.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Linq;
 using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Text;
 using Shouldly;
 using Stryker.Core.MutantFilters;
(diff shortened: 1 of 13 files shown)


rouke-broersma:
@dupdob awesome! I enabled rebase merging on the repo temporarily to preserve the history of what is fixed with this PR and so it shows up in the auto generated changelog. Your commit messages were excellent so that went great :) It would be easier if in the future you could make them separate pull requests. I know it's a chore to create all those as separate PR's but it makes it a lot easier to review and merge them in a way that they show up in the changelog properly.

dupdob:
I generally do PR for one feature and one feature only, but this was more practical for me to address several smallish issues into one. I took the time to have (relatively) clearly delineated commits. I will try several PRs next time, but it my lead to merge conflicts afterward.

rouke-broersma:
I suppose it is also acceptable to temporarily allow rebase merge for these cases if the commits are good. Fine, I'm convinced :)

PR #1868, opened 2022-01-20
fix: mutation testing for self-contained apps

### LINKED ISSUE #1855: Stryker fails for self-contained apps

psfinaki:
closes #1855

### DIFF
--- a/src/Stryker.CLI/Stryker.CLI.UnitTest/packages.lock.json
+++ b/src/Stryker.CLI/Stryker.CLI.UnitTest/packages.lock.json
@@ -80,10 +80,10 @@
       },
       "Buildalyzer": {
         "type": "Transitive",
-        "resolved": "3.2.6",
-        "contentHash": "dZJvQaVR9+v6O2S7zL8uzHGmDVXDBlk5EYOwedAWZiezg8nVJuRDDkJdU9DHdxtczWpxl6B/G68e7rf/3TY4qQ==",
+        "resolved": "3.2.8",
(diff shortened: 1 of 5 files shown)


psfinaki:
Now looking forward to the release here :D

PR #1875, opened 2022-01-21
fix(Rollback block mutations first): Improve stability by rollbacking block mutations before all mutations are removed

richardwerkman:


src/Stryker.Core/Stryker.Core.UnitTest/Mutants/CsharpMutantOrchestratorTests.cs
```diff
@@ -1,16 +1,17 @@
...
+using Stryker.Core.Options;
 using Xunit;
 
 namespace Stryker.Core.UnitTest.Mutants
 {
     public class CsharpMutantOrchestratorTests : MutantOrchestratorTestsBase
     {
         [Fact]
-        public void ShouldMutateInterfaces()
+        public void ShouldNotMutateInterfaces()
```
rouke-broersma, line 14: 🤔

src/Stryker.Core/Stryker.Core.UnitTest/Stryker.Core.UnitTest.csproj
```diff
@@ -40,6 +40,7 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(MicrosoftTestPlatform)" />  <!-- From Directory.Build.props -->
     <PackageReference Include="Moq" Version="4.16.1" />
     <PackageReference Include="Shouldly" Version="4.0.3" />
+    <PackageReference Include="System.Collections" Version="4.3.0" />
```
rouke-broersma, line 43: 🤔

rouke-broersma: Last updated 11/15/2016

src/Stryker.Core/Stryker.Core/Compiling/RollbackProcess.cs
```diff
@@ -123,25 +125,30 @@ public RollbackProcessResult Start(CSharpCompilation compiler, ImmutableArray<Di
             return (null, -1);
         }
 
-        private int? ExtractMutationIfAndId(SyntaxNode node)
+        private (int? Id, string Engine, string Type) ExtractMutationInfo(SyntaxNode node)
```
rouke-broersma, line 128: Since we use this construct in multiple places, and it is not only used to simply return multiple values, I think we should create a private class for it.

rouke-broersma: This helps avoid having to do this with 'Item3' https://github.com/stryker-mutator/stryker-net/pull/1875/files#diff-6a4265e13c259acff0fccc6bb044650b26193ed6c95b6cd6b88c99ff2ba401bdR238

src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs
```diff
@@ -121,24 +123,29 @@ public static SyntaxNode RemoveMutant(SyntaxNode nodeToRemove)
             throw new InvalidOperationException($"Unable to find an engine to remove injection from this node: '{nodeToRemove}'");
         }
 
-        public static (string engine, int id) FindEngine(SyntaxNode node)
+        public static (int id, string engine, string type) FindAnnotations(SyntaxNode node)
```
rouke-broersma, line 126: This seems somewhat similar to the rollback logic. Perhaps instead of a private class we should have a public class to hold this tuple data.

Issue #1864, opened 2022-01-16
ProgressBar NRE when no mutants tested


commonquail:
This is similar to #1806.

When Stryker.NET 1.2.2 generates some non-zero amount of mutations but skips all of them the ProgressBar reporter throws a NullReferenceException:

```
Version: 1.2.2

[14:57:24 INF] Identifying project to mutate.
[14:57:25 INF] The project /tmp/tmp.g40Wn4yMdA/l/l.csproj will be mutated.
[14:57:26 INF] Analysis complete.
[14:57:26 INF] Building test project /tmp/tmp.g40Wn4yMdA/t/t.csproj (1/1)
[14:57:29 INF] Total number of tests found: 1.
[14:57:29 INF] Initial testrun started.
[14:57:31 INF] 1 mutants created
[14:57:31 INF] Capture mutant coverage using 'CoverageBasedTest' mode.
[14:57:32 INF] 1     mutants got status NoCoverage.   Reason: Mutant has no test coverage
[14:57:32 INF] 1     total mutants are skipped for the above mentioned reasons
[14:57:32 INF] 0     total mutants will be tested
[14:57:32 WRN] It looks like all non-ignored mutants are not covered by a test. Go add some tests!
[14:57:33 INF] Time Elapsed 00:00:09.5080642
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Stryker.Core.Reporters.Progress.ProgressBar.Tick(String message)
   at Stryker.Core.Reporters.Progress.ProgressBarReporter.ReportFinalState()
   at Stryker.Core.Reporters.Progress.ProgressReporter.OnAllMutantsTested(IReadOnlyProjectComponent reportComponent)
   at Stryker.Core.Reporters.BroadcastReporter.OnAllMutantsTested(IReadOnlyProjectComponent reportComponent)
   at Stryker.Core.StrykerRunner.RunMutationTest(IStrykerInputs inputs, ILoggerFactory loggerFactory, IProjectOrchestrator projectOrchestrator)
   at Stryker.CLI.StrykerCli.RunStryker(IStrykerInputs inputs) in /_/src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs:line 90
   at Stryker.CLI.StrykerCli.<>c__DisplayClass9_0.<Run>b__0() in /_/src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs:line 65
[...]

PR #1882, opened 2022-01-27
fix: fix NRE

dupdob:
Fix Null Reference Exception with progressbarreporter when mutantfreeworld (#1864)
(my smallest PR so far 😉)


src/Stryker.Core/Stryker.Core/Reporters/Progress/ProgressBar.cs
```diff
@@ -28,7 +28,7 @@ public class ProgressBar : IProgressBar, IDisposable
 
         public void Start(int maxTicks, string message) => _progressBar = new ShellProgressBar.ProgressBar(Math.Max(maxTicks, 1), message, _options);
 
-        public void Tick(string message) => _progressBar.Tick(message);
+        public void Tick(string message) => _progressBar?.Tick(message);
```
psfinaki, line 32: So you've basically fixed [this](https://www.youtube.com/watch?v=1M02bAWDFkI) :D

PR #1901, opened 2022-02-10
Fix CS0579

bcollamore:


rouke-broersma:
Awesome! One of those things that you kinda just deal with without thinking about fixing it 😅

That file is very much needed and this issue was pretty annoying so thank you :)

Issue #466, opened 2019-04-21
Passing build variables


oicguy:
I was referred to stryker by a co-worker who is really excited about mutation testing in general. After reading the documentation I decided to get it a try. My project is using the .Net Framework, not .Net Core so I expected some hiccups.

After installing the global tool and specifing the sln file, msbuild finally ran but failed. The reason is because a targets file isn't in the expected location. How can this be since I can run it in  Visual Studio without a problem? Turns out, Visual Studio modifies some of the build variables so keep certain consistencies. 

In my case, I've installed the DacFx nuget package and it adds a *Microsoft.Data.Tools.Schema.SqlTasks.targets* files that allows me to build a SQL Server Data Tools project. This file is placed by the nuget install in the *C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\MSBuild\Microsoft\VisualStudio\v15.0\SSDT* because the `MSBuildExtensionsPath` was modified to be *C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\MSBuild*. Normally the path is *C:\Program Files (x86)\MSBuild*.

If there were a way to set the property values from either the command line or the configuration file the issue could be resolved with very little change to the stryker code itself. While I'm approaching this from a .Net Framework perspective, I've used the dontnet cli as well and am aware that it can use the MSBuild properties as well. It seems this would be a powerful addition that would allow the developers to harness more of the build process without having to build in specific support for something.

Mobrockers:
Could you give an example of the arguments you need passed to msbuild to be able to use stryker?

oicguy:
If I were using MSBuild directly, I'd need to use `msbuild.exe /p:MSBuildExtensionPath=C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\MSBuild`. Obviously that specific for my computer since not everyone would use the C: drive or even the Program Files directory. Allowing us to pass in variable like this would allow use to maybe use different configurations `/p Configuration=Debug` or even different Preprocessor directive `/p:DefinedConstants=Trace;Debug;Analysis`.

If I may make a suggestion, I'd like to see in the configuration file an msbuild section where you can define variables and other msbuild sections like the following:
msbuild:{
    properties: {
        MSBuildExtensionsPath: "..."
    },
    switches:["m"]
}
With this layout, you'll only have to format the output and this setup would become /p:MSBuildExtensionsPath=... /m

Mobrockers:
[...]

PR #1903, opened 2022-02-12
dev(buildalyzer): Begin encapsulating Buildalyzer

bcollamore:
Start encapsulating Buildalyzer, so as to begin exploring possible alternatives in the future.
* New interfaces/classes hierarchy to encapsulate Buildalyzer classes
* ProjectOrchestrator is updated - it's an entry point to the new classes.
* ProjectFileReader - it's not updated; kept old API for now (updating this API to fully encapsulate Buildalyzer would be next step.)
* Added an AnalyzerOption: currently Buildalyzer is the only supported option.  (This arguably should have been excluded for now for smaller batch size.)

Other minor changes:
* IAnalyzerResultExtension moved from \Buildalyzer folder. Minimal changes, just added "ToAnalyzerResult", which is needed for now to keep ProjectFileReader on the previous API.
* Buildalyzer.cs moved, repurposed, and renamed to SolutionAnalyzerManagerProvider.cs
* TestHelper.cs: removed PackageReferences as it's not used elsewhere


bcollamore:
* Updated ProjectFileReader to use abstraction layer. (Now 1 place where Buildalyzer is directly used.)
* Removed AnalyzerOption: (Explicitly only support Buildalyzer.)
* Move Logging of AnalyzerResult details into AnalyzerResult
* BuildalyzerProjectAnalyzer: Move ProjectFileReader targetFramework logic here
* BuildalyzerProjectAnalyzerManager: Move some ProjectFileReader logic here

bcollamore:
Understood @rouke-broersma. In the meantime I completed encapsulation of Buildalyzer (i.e., porting ProjectFileReader to use the abstraction layer).  There would seem to be benefits of encapsulating Buildalyzer independent of replacing it, and the Buildalyzer API already has some cognitive load associated with it.  For example, 
* Uses IAnalysisResults despite Stryker only needing one result.
* Redundant/mixed logic between ProjectFileReader and ProjectAnalyzerManager (ie solution analyses) (eg., TargetFramework support in one path only, one path has try/catch, one path has logging, one path uses NugetRestore on failure) 
* Buildalyzer seems to be analyzing the same project multiple times (when solution analyses performed)

It's reasonable to believe that a cleaner abstraction layer could simplify the interface with Buildalyzer.

I had an unpolished experiment from ~18 months ago that uses MsBuild in lieu of Buildalyzer. It was not ready for prime time. It's also entirely possible the additional work could create excessive support situations, as you point out. I will dust that off, hopefully soon, to see the proof of concept. It combined the work of encapsulating Buildalyzer and replacing it. This PR focused on the former, to open future investigations, in an attempt to help issues such as https://github.com/stryker-mutator/stryker-net/issues/466 and worked on by @pawelhevy.

Having said all that, I understand if this PR is deemed unworthy or too risky. Stay tuned for a POC on replacing Buildalyzer.

bcollamore:
See https://github.com/stryker-mutator/stryker-net/pull/1919 for an as-of-yet non-operational proof-of-concept for using MsBuild for analysis. I can try to make it functional after I return from a vacation in a week.

bcollamore:
See the aforementioned PR for a now-operational proof-of-concept of how replacing BuildAlyzer can work (or more specifically, how a compile-time or run-time choice could be made to use BuildAlyzer or MsBuild).

PR #1905, opened 2022-02-14
fix(mutation placing): Support statement mutations in expression bodies lambdas

dupdob:
properly inject statement mutations in expression body lambda
fixes #1902

PR #1906, opened 2022-02-16
Use the NuGet.Frameworks package to parse target frameworks

0xced:
This is safer to use than a hand-rolled version. The public methods and the public `Framework` enum have been marked obsolete.

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultExtensions.cs
+++ b/src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultExtensions.cs
@@ -2,12 +2,12 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
-using System.Text.RegularExpressions;
 using Buildalyzer;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.Extensions.Logging;
+using NuGet.Frameworks;
 using Stryker.Core.Exceptions;
 
 namespace Stryker.Core.Initialisation.Buildalyzer
@@ -97,37 +97,21 @@ namespace Stryker.Core.Initialisation.Buildalyzer
             public Assembly LoadFromPath(string fullPath) => Assembly.LoadFrom(fullPath);
         }
 
-        public static Framework GetTargetFramework(this IAnalyzerResult analyzerResult)
+        internal static NuGetFramework GetNuGetFramework(this IAnalyzerResult analyzerResult)
         {
-            try
+            var framework = NuGetFramework.Parse(analyzerResult.TargetFramework ?? "");
+            if (framework == NuGetFramework.UnsupportedFramework)
             {
-                return ParseTargetFramework(analyzerResult.TargetFramework);
+                var atPath = string.IsNullOrEmpty(analyzerResult.ProjectFilePath) ? "" : $" at '{analyzerResult.ProjectFilePath}'";
+                var message = $"The target framework '{analyzerResult.TargetFramework}' is not supported. Please fix the target framework in the csproj{atPath}.";
+                throw new InputException(message);
             }
-            catch (ArgumentException)
-            {
-                throw new InputException($"Unable to parse framework version string {analyzerResult.TargetFramework}. Please fix the framework version in the csproj.");
-            }
+            return framework;
         }
 
-        /// <summary>
-        /// Extracts a target <c>Framework</c> and <c>Version</c> from the MSBuild property TargetFramework
-        /// </summary>
-        /// <returns>
-        /// A tuple of <c>Framework</c> and <c>Version</c> which together form the target framework and framework version of the project.
-        /// </returns>
-        /// <example>
-        /// <c>(Framework.NetCore, 3.0)</c>
-        /// </example>
-        public static (Framework Framework, Version Version) GetTargetFrameworkAndVersion(this IAnalyzerResult analyzerResult)
+        internal static bool TargetsFullFramework(this IAnalyzerResult analyzerResult)
         {
-            try
-            {
-                return (ParseTargetFramework(analyzerResult.TargetFramework), ParseTargetFrameworkVersion(analyzerResult.TargetFramework));
-            }
-            catch (ArgumentException)
-            {
-                throw new InputException($"Unable to parse framework version string {analyzerResult.TargetFramework}. Please fix the framework version in the csproj.");
-            }
+            return GetNuGetFramework(analyzerResult).IsDesktop();
         }
 
         public static Language GetLanguage(this IAnalyzerResult analyzerResult)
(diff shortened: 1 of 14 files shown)


src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultExtensions.cs
```diff
@@ -97,16 +96,20 @@ private sealed class AnalyzerAssemblyLoader : IAnalyzerAssemblyLoader
...
+        {
+            return NuGetFramework.Parse(analyzerResult.TargetFramework ?? "");
+        }
+
+        internal static bool TargetsFullFramework(this IAnalyzerResult analyzerResult)
+        {
+            return GetNuGetFramework(analyzerResult).IsDesktop();
+        }
+
+        [Obsolete("Please consider using the NuGetFramework type instead (from the 'NuGet.Frameworks' package)", error: true)]
```
rouke-broersma, line 109: Marking things as Obsolete is not something we generally do. It might be a good idea if we currently cannot easily replace all usages, but then I would expect to keep the unit test for now.

dupdob:
Thanks. Externalizing this parsing to a third party library is a good idea.
I failed to find any documentation on this Nuget.Frameworks. Do you have any reference ?
Also, you will want to keep unit tests: the target framework string has already changed several times across VS/VC/MSBuild/Nuget version. So we need to have unit tests to ensure we are still compatible with older csproj file, those are also handy when a new framework version string breaks everything.
You also want unit tests for when we fail to recognize the framework version: as of now, Stryker halts with an exception, on purpose, if we fail to parse the version.
As I did not find any documentation for Nuget.Frameworks, I am not sure what will happen if it receives an invalid version string.

src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultExtensions.cs
0xced: I'm not sure to understand what you mean. Should I completely delete the now unused methods or should I just keep them around for compatibility even if they are not referenced anymore?

I initially chose to keep them and mark them obsolete since the `IAnalyzerResultExtensions` class and the `Framework` enum are public types. If someone is using those types when consuming the `stryker` NuGet package , deleting them would leave them in the dark. With the `Obsolete` attribute, a clear error message hints at a solution for replacing the removed types and methods.

Anyway, I'll update the pull request to either keep the methods or delete them as you see fit.

0xced:
> I failed to find any documentation on this Nuget.Frameworks. Do you have any reference ?

Unfortunately, documentation about the `Nuget.Frameworks` package is almost non-existent. 😔 The official [NuGet Client SDK][1] does not mention it at all. All you'll find is a link to [Exploring the NuGet v3 Libraries, Part 2][2] blog post by Dave Glick that briefly mention it. The next best option is reading the [NuGet.Frameworks][3] source code on GitHub. 🤷‍♂️

> Also, you will want to keep unit tests: the target framework string has already changed several times across VS/VC/MSBuild/Nuget version. So we need to have unit tests to ensure we are still compatible with older csproj file, those are also handy when a new framework version string breaks everything.

I have restored the `ProjectAnalyzerShouldDecodeFramework` tests and adapted them to the new `NuGetFramework` type in 5253ddb6a3f5b65c941de9fd955d3ac2479cbced.

> You also want unit tests for when we fail to recognize the framework version: as of now, Stryker halts with an exception, on purpose, if we fail to parse the version.

There does not seem to be a test covering that case currently. Additionally I don't think what you're saying is true. The previous code that parses the tfm does not throw but returns `Framework.Unknown` for invalid tfms. Note that the current implementation is roughly equivalent (see my answer just below about how parsing `NuGetFramework` works).

```csharp
private static Framework ParseTargetFramework(string targetFrameworkVersionString)
{
    return targetFrameworkVersionString switch
    {
        string framework when framework.StartsWith("netcoreapp") => Framework.DotNet,
        string framework when framework.StartsWith("netstandard") => Framework.DotNetStandard,
        string framework when framework.StartsWith("net") && char.GetNumericValue(framework[3]) >= 5 => Framework.DotNet,
        string framework when framework.StartsWith("net") && char.GetNumericValue(framework[3]) <= 4 => Framework.DotNetClassic,
        _ => Framework.Unknown
    };
}
```
 
> As I did not find any documentation for Nuget.Frameworks, I am not sure what will happen if it receives an invalid version string.

A quick test shows that the only way to make `NuGetFramework.Parse` throw an exception is to pass `null`:

```csharp
using NuGet.Frameworks;

foreach (var tfm in new[] { "net5.0-windows", "", "xyz", "🤡", null })
{
    try
    {
        Console.WriteLine($"<{tfm}> = {NuGetFramework.Parse(tfm)}");
    }
    catch (Exception exception)
    {
        Console.WriteLine($"[ERROR] {exception.Message}");
    }
}
```

Resluting in

```
<net5.0-windows> = net5.0-windows
<> = Unsupported,Version=v0.0
<xyz> = Unsupported,Version=v0.0
<🤡> = Unsupported,Version=v0.0
[ERROR] Value cannot be null. (Parameter 'folderName')
```

So basically, when the target framework moniker is invalid, `NuGetFramework.UnsupportedFramework` is returned.

[1]: https://docs.microsoft.com/en-us/nuget/reference/nuget-client-sdk#third-party-documentation
[2]: https://www.daveaglick.com/posts/exploring-the-nuget-v3-libraries-part-2#framework
[3]: https://github.com/NuGet/NuGet.Client/tree/dev/src/NuGet.Core/NuGet.Frameworks

src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultExtensions.cs
rouke-broersma: I did not check if all usages were already replaced. 

If they are, you can remove the methods and classes instead of marking as obsolete. As far as I know no one else currently uses the Stryker core package and I am fine with breaking in this way until we have actual consumers.

If they are not yet all replaced (within Stryker), marking as obsolete seems fine to me, but then we should keep the unit tests.

dupdob:
Yes, the  test was missing. I added it in a PR I opened 2 days ago (you can see the test [here](https://github.com/stryker-mutator/stryker-net/pull/1905/commits/6d9c0a95adfa50df4f78a800439321df7833a95b#diff-7ac954b3982a221de06b52bcdde548ea509a15bfc79fef82cf7b87071a223197L102)). But the current version does throw when the string does not respect any pattern. See here: https://github.com/stryker-mutator/stryker-net/blob/d883f58ff66555d05cc3207be01a2bab0e71902b/src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultExtensions.cs#L100.

In any case, we need to have a clear error message and to stop Stryker when we do not recognize the framework [version].
Based on your testing, I suggest raising an exception when the framework is 'Unknown' with a message providing the version string returned by Buildalyzer.

This way, user can report the problem, helping us decide if the problem is that the framework is not supported or that Buildadalyzer failed to report the good version.

0xced:
I tried to take in account all your comments and force-pushed everything in 12862208d9b6fbe3854f316ecc0fd6f7861cc935.

@dupdob It seems something weird happened when merging #1905 since the `ProjectAnalyzerShouldRaiseExceptionsForIllFormedFramework` test does not exist on the `master` branch (currently fb7ed23a5fc9e323bf498abefc9dfabc945cf9a3)! So I added it back and adapted it for the new parsing using the `NuGetFramework` type.

rouke-broersma:
I'll test and review this tomorrow, thanks! By the way, force pushing is not required, we exclusively use squash merging. If you prefer local squashing/amending and force pushing you are of course still free to do so. Just know that it does not matter to us either way.

PR #1930, opened 2022-02-24
Guard against `ProcessorCount` reporting values lower then `1`

### LINKED ISSUE #1927: ConcurrencyInput should not default to 0 if Environment.ProcessorCount reports 0
When a system does return a processor count of 0 (for example gitlabs shared runners) the default value will be set to 0. This will result in a crash unless the user overrides the concurrency setting. [This issue has existed before](https://github.com/stryker-mutator/stryker-net/issues/122) and was fixed with ```Math.Max(Environment.ProcessorCount / 2, 1)```, but now exists again.

If the Default is 0 a warning should be logged that the ProcessorCount is not reported by the system and 1 should be used to still work without overriding the value.

Something like this

```c#
if (SuppliedInput == null)
{
    if (Default.Value == 0)
    {
        logger.LogWarning("Logical processor count is not reported by the system, using concurrency of 1. "
            + "The recommended concurrency is processorCount / 2.");
        return 1;
    }
        
    return Default.Value;
}
```

Currently the code looks like this [ConcurrencyInput.cs](https://github.com/stryker-mutator/stryker-net/blob/285afbed4116cdda8c35dd68fc4d7dc4fa8b393b/src/Stryker.Core/Stryker.Core/Options/Inputs/ConcurrencyInput.cs).

- Settings the default to ```Math.Max(Environment.ProcessorCount / 2, 1)``` and not logging a warning.
- Logging a warning that ProcessorCount is 0 regardless of what happens. In case the value is set by the user the message that the the concurrency you set is bigger than 0 also does not make a whole lot of sence / is not a very good suggestion, so it could be skipped. But also might be overthinking this at this point. I mostly care about it not crashing by default.

eNeRGy164:
Should fix #1927

I have seen reports online about the method also reporting `-1`, so wanted to be safe and did check on any value lower then `1`.

Added a warning and a message that the user can get rid of the warning by supplying a value themselves, they probably know the amount of processors that would make sense.

Can't think of an easy way of adding unit test coverage for this.

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Options/Inputs/ConcurrencyInput.cs
+++ b/src/Stryker.Core/Stryker.Core/Options/Inputs/ConcurrencyInput.cs
@@ -15,14 +15,19 @@ Reasons you might want to lower this setting:
     - You're running on a shared server
     - You are running stryker in the background while doing other work";
 
-        public override int? Default => Environment.ProcessorCount / 2;
+        public override int? Default => Math.Max(Environment.ProcessorCount / 2, 1);
 
         public int Validate(ILogger<ConcurrencyInput> logger = null)
         {
             logger ??= ApplicationLogging.LoggerFactory.CreateLogger<ConcurrencyInput>();
 
-            if (SuppliedInput == null)
+            if (SuppliedInput is null)
             {
+                if (Environment.ProcessorCount < 1)
+                {
+                    logger.LogWarning("Processor count is not reported by the system, using concurrency of 1. Set a concurrency to remove this warning.");
+                }
+
                 return Default.Value;
             }
 


src/Stryker.Core/Stryker.Core/Options/Inputs/ConcurrencyInput.cs
```diff
@@ -15,14 +15,19 @@ public class ConcurrencyInput : Input<int?>
     - You're running on a shared server
     - You are running stryker in the background while doing other work";
 
-        public override int? Default => Environment.ProcessorCount / 2;
+        public override int? Default => Math.Max(Environment.ProcessorCount / 2, 1);
```
psfinaki, line 18:
```suggestion
        public override int? Default => Environment.ProcessorCount != 0 ? Environment.ProcessorCount / 2 : 1;
```

psfinaki: Just a suggestion for better explicitness

psfinaki (approved):

richardwerkman:
I agree this would be hard to unit test. Adding a wrapper around Environment.ProcessorCount also isn't desirable. So this seems fine to me. Thanks for the quick fix!

Issue #1394, opened 2021-02-03
Clickable named report url in terminal


rouke-broersma:
**Is your feature request related to a problem? Please describe.**
Our report urls are fully qualified. Some terminals support named urls. It might also be good to add a file protocol to the existing url's so it's easier to detect the local filepath as an url by terminals that already support this.

This is different from #1395 in that the URI's would no longer need to be prefixed by `file://` as we would prefix them with invisible ANSI characters that provide the same hint to supported terminals.

**Describe the solution you'd like**
Apparently there is ansi support for clickable named links. Some terminals implement this and in those terminals we could use this. https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

**Additional context**
We do not want to print the named url stuff on terminals that do not support this as the characters might simply be printed to the console which makes it hard for users to open the file links.

I could find javascript libraries for this but no csharp library. We could create our own library (and publish it?) based on for example https://github.com/sindresorhus/terminal-link/blob/main/index.js 

Support for clickable links is added to the new windows terminal which means our largest user base will at some point have this feature available to them: https://github.com/microsoft/terminal/issues/204

eNeRGy164:
I did something like this for FluentAssertions's build process: https://github.com/fluentassertions/fluentassertions/blob/develop/Build/Build.cs#L171

Is that what you are looking for?

eNeRGy164:
I did something like this for FluentAssertions's build process: https://github.com/fluentassertions/fluentassertions/blob/develop/Build/Build.cs#L171

Is that what you are looking for?

eNeRGy164:
I did something like this for FluentAssertions's build process: https://github.com/fluentassertions/fluentassertions/blob/develop/Build/Build.cs#L171

Is that what you are looking for?

rouke-broersma:
I think that is the same yes, however I'm worried about

> We do not want to print the named url stuff on terminals that do not support this as the characters might simply be printed to the console which makes it hard for users to open the file links.

I don't know if that's an actual problem. If terminals that don't support this simply drop the unrecognizable ansi then that's fine to me. If they don't then we will need some way of keeping track of which terminal supports this and only print the characters in those consoles. Preferably we would not maintain a list of supported terminals in stryker but so far I have not found a package for this.

eNeRGy164:
I think I remember CMD just ignored the ANSI hyperlink part.

[...]

PR #1936, opened 2022-03-02
feat(reporting): Make links clickable in terminals that support Ansi magic

eNeRGy164:
After discussion in #1394

* Replaced `Crayon` with `Spectre.Console`
* Made links clickable in the terminal
* Added a smiley, as requested by the Info Support open source council ;)

I think Spectre.Console removes a lot of clutter when generating the table/tree output.
Allows for more fancy stuff in the future, and is well maintained.

I didn't replace the progress bar implementation at this moment in time.

Output using the changed code:
spectre table

spectre tree

spectre dot

spectre html

spectre json

spectre pb


eNeRGy164:
Hmm, the failing test is flaky.. it succeeds when it's run individual, but fails when running all tests together...

richardwerkman:
@eNeRGy164 are you looking into the flaky test? I'd very much like to merge this 🙌

eNeRGy164:
Yes, already did have a quick look, but did not find a solution yet

eNeRGy164:
> @eNeRGy164 are you looking into the flaky test? I'd very much like to merge this 🙌

Go ahead :)

eNeRGy164:
@rouke-broersma now you introduced conflicts again |:( :)

rouke-broersma:
> @rouke-broersma now you introduced conflicts again |:( :)

😅😅

richardwerkman:
You can rebuild the packages.lock files by building the project with the latests changes included :+1:

src/Stryker.Core/Stryker.Core/Reporters/ClearTextReporter.cs
```diff
@@ -41,71 +40,79 @@ public void OnMutantTested(IReadOnlyMutant result)
...
-                _consoleWriter.WriteLine($"┌─{dashes}┬──────────┬──────────┬───────────┬────────────┬──────────┬─────────┐");
-                _consoleWriter.WriteLine($"│ File{new string(' ', filePathLength - 4)}│  % score │ # killed │ # timeout │ # survived │ # no cov │ # error │");
-                _consoleWriter.WriteLine($"├─{dashes}┼──────────┼──────────┼───────────┼────────────┼──────────┼─────────┤");
-
-                DisplayComponent(reportComponent, filePathLength);
+                _console.WriteLine();
+                _console.WriteLine();
+                _console.WriteLine("All mutants have been tested, and your mutation score has been calculated");
+
+                var table = new Table()
```
rouke-broersma, line 50: This is super sexy!

rouke-broersma:
For some reason the report path is printed over two distinct lines, does spectre.console enforce some kinda line length?

https://dev.azure.com/stryker-mutator/Stryker/_build/results?buildId=9425&view=logs&j=99031f74-bf3d-58da-997c-4e095f08dc64&t=4fae8bbe-7dd3-5ecd-793d-f1d2f82f0225&l=3834

``` 
Your json report has been generated at:
D:\a\1\s\integrationtest\TargetProjects\NetCoreTestProject.XUnit\StrykerOutput\2
022-03-04.22-22-53\reports\mutation-report.json
Hint: by passing "--open-report or -o" the report will open automatically once 
Stryker is done.

Your html report has been generated at:
D:\a\1\s\integrationtest\TargetProjects\NetCoreTestProject.XUnit\StrykerOutput\2
022-03-04.22-22-53\reports\mutation-report.html
You can open it in your browser of choice.

```

rouke-broersma (commented):
For some reason the report path is printed over two distinct lines, does spectre.console enforce some kinda line length?

https://dev.azure.com/stryker-mutator/Stryker/_build/results?buildId=9425&view=logs&j=99031f74-bf3d-58da-997c-4e095f08dc64&t=4fae8bbe-7dd3-5ecd-793d-f1d2f82f0225&l=3834

``` 
Your json report has been generated at:
D:\a\1\s\integrationtest\TargetProjects\NetCoreTestProject.XUnit\StrykerOutput\2
022-03-04.22-22-53\reports\mutation-report.json
Hint: by passing "--open-report or -o" the report will open automatically once 
Stryker is done.

Your html report has been generated at:
D:\a\1\s\integrationtest\TargetProjects\NetCoreTestProject.XUnit\StrykerOutput\2
022-03-04.22-22-53\reports\mutation-report.html
You can open it in your browser of choice.
``` 

From a quick search it seems like Spectre.Console out-detects max console width, that makes me think if it cannot be auto detected a max-width of 80 is chosen (legacy terminal character limit). On the one hand this seems fine, on the other hand in a CI pipeline it's not really necessary to cut off a filepath.

It seems to also affect the table and file tree reports in azure devops output making them less usefull because the content is simply cut off in some cases or placed on new lines in others.

```
╭────────────────────────────┬───────┬───────┬────────┬─────────┬───────┬──────╮
│ File                       │    %  │    #  │     #  │      #  │ # no  │   #  │
│                            │ score │ kill… │ timeo… │ surviv… │   cov │ err… │
├────────────────────────────┼───────┼───────┼────────┼─────────┼───────┼──────┤
│ All files                  │ 16.92 │     9 │      2 │      92 │    50 │    2 │
│ Csharp7.cs                 │  0.00 │     0 │      0 │       1 │     1 │    0 │
│ Csharp8.cs                 │  0.00 │     0 │      0 │      14 │     6 │    0 │
│ Endlessloop.cs             │ 40.00 │     0 │      2 │       3 │     1 │    0 │
│ LinqMagic.cs               │  0.00 │     0 │      0 │      17 │    15 │    0 │
│ Math\Calc\DummyCalc.cs     │  0.00 │     0 │      0 │      13 │    13 │    0 │
│ Math\DummyMath.cs          │ 66.67 │     2 │      0 │       1 │     0 │    0 │
│ Person.cs                  │   N/A │     0 │      0 │       9 │     0 │    0 │
│ String\StringMagic.cs      │  0.00 │     0 │      0 │      17 │    13 │    2 │
│ String\StringReplaceChain. │  0.00 │     0 │      0 │       8 │     1 │    0 │
│ cs                         │       │       │        │         │       │      │
│ Student.cs                 │ 75.00 │     3 │      0 │       9 │     0 │    0 │
│ UnsafeCode.cs              │ 100.0 │     4 │      0 │       0 │     0 │    0 │
│                            │     0 │       │        │         │       │      │
╰────────────────────────────┴───────┴───────┴────────┴─────────┴───────┴──────╯
```

I think we can accept fixing this in a new issue though it only affects CI runs and is probably what you want in a non-CI scenario.

src/Stryker.Core/Stryker.Core/Reporters/DashboardReporter.cs
```diff
@@ -45,19 +43,19 @@ public void OnAllMutantsTested(IReadOnlyProjectComponent reportComponent)
                 }
                 else
                 {
-                    _consoleWriter.Write(Output.Cyan("Hint: by passing \"--open-report:dashboard or -o:dashboard\" the report will open automatically once Stryker is done."));
+                    _console.Write("[Cyan]Hint: by passing \"--open-report:dashboard or -o:dashboard\" the report will open automatically once Stryker is done.[/]");
                 }
 
                 _logger.LogDebug("Your stryker report has been uploaded to: \n {0} \nYou can open it in your browser of choice.", reportUri);
-                _consoleWriter.Write(Output.Green($"Your stryker report has been uploaded to: \n {reportUri} \nYou can open it in your browser of choice."));
+                _console.Write($"[Green]Your stryker report has been uploaded to: \n {reportUri} \nYou can open it in your browser of choice.[/]");
```
rouke-broersma, line 50: This could probably also be upgraded to an Ansi link.

src/Stryker.Core/Stryker.Core/Reporters/HtmlReporter/HtmlReporter.cs
```diff
@@ -38,22 +37,30 @@ public void OnAllMutantsTested(IReadOnlyProjectComponent reportComponent)
...
+            }
+            else
+            {
+                _console.MarkupLine("[Cyan]Hint: by passing \"--open-report or -o\" the report will open automatically once Stryker is done.[/]");
+            }
+
+            _console.WriteLine();
+            _console.MarkupLine("[Green]Your html report has been generated at:[/]");
+
+            if (_console.Profile.Capabilities.Links)
```
rouke-broersma, line 54: It appears the 'Ansi link shortcuts' capabilities check of Spectre.Console is not that advanced, it will probably only detect cmd.exe as not supporting Ansi links. I tested this on visual studio terminal, git bash, powershell, Gnome terminal and Guake. All showed as supporting Ansi links which is not technically true. However all do support Ansi so they simply ignored the special ansi link characters and only printed the link text. This is fine, however it does mean that we must print the full report path to disk. I had preferred to print something like [your report is here](#here) instead but at this time that does not seem possible.

See: https://github.com/spectreconsole/spectre.console/blob/52c1d9122bce4c73ae9205e2f2aff01cba03a4c4/src/Spectre.Console/AnsiConsoleFactory.cs#L49

src/Stryker.Core/Stryker.Core/Reporters/HtmlReporter/HtmlReporter.cs
```diff
@@ -38,22 +37,30 @@ public void OnAllMutantsTested(IReadOnlyProjectComponent reportComponent)
...
+            {
+                _console.MarkupLine($"[Green][link={reportUri}]{reportPath}[/][/]");
+            }
+            else
             {
-                _consoleWriter.Write(Output.Cyan("Hint: by passing \"--open-report or -o\" the report will open automatically once Stryker is done."));
+                _console.MarkupLine($"[Green]{reportPath}[/]");
```
rouke-broersma, line 60: I think we should print the report uri and not the report path if ansi links are not supported, because some terminals have built-in uri detection but not ansi link support. For a user with a terminal that supports neither nothing changes as browsers support full file uri's in general. It would provide the largest cover.

src/Stryker.Core/Stryker.Core/Reporters/HtmlReporter/HtmlReporter.cs
rouke-broersma: Created an issue at spectre.console: https://github.com/spectreconsole/spectre.console/issues/764

Issue #1937, opened 2022-03-03
[Regression] NRE in Stryker 1.4.0


jfheins:
**Describe the bug**
Stryket.net 1.4.0 crashes unexpectedly.

**Logs**
I am not sure how much project data is contained in the log, can I mail it?

**Expected behavior**
I expect stryker not to crash.
I tested Stryker.net 1.3.1 on Tuesday and it worked fine but (with the same code of course) but version 1.4.0 does not work any more.

**Desktop (please complete the following information):**
 - OS: ubuntu-20.04
 - Type of project asp.net core on .net 6
 - Framework Version .net 6
 - Stryker Version 1.4.0

**Additional context**

```
[20:53:48 INF] Initial testrun started.
[20:58:33 WRN] Initial test run encounter a unexpected test case (mytestcase(null,7ec6a09b-ef05-4ba0-8fc4-d4ab8241b506,False,"123")), mutation tests may be inaccurate. Disable coverage analysis if your have doubts.
[20:58:33 WRN] Initial test run encounter a unexpected test case (mytestcase(00000000-0000-0000-0000-000000000000,a6e2f34d-dcd2-47f3-96ea-0f6be47212c5,False,"234")), mutation tests may be inaccurate. Disable coverage analysis if your have doubts.
[20:58:33 WRN] Initial test run encounter a unexpected test case (mytestcase(15909311-1dc0-495a-b0f5-a938faae7e5c,c312cf38-aa26-4560-a44e-97810cd6d0e9,False,"345")), mutation tests may be inaccurate. Disable coverage analysis if your have doubts.
[20:58:33 WRN] Initial test run encounter a unexpected test case (mytestcase(98c42e5d-0a76-443a-8825-57b9130d6a3c,98c42e5d-0a76-443a-8825-57b9130d6a3c,True,"456")), mutation tests may be inaccurate. Disable coverage analysis if your have doubts.
[20:58:40 ERR] Mutation not found, this should not happen
[20:58:42 ERR] Mutation not found, this should not happen
[20:58:42 ERR] Mutation not found, this should not happen
[20:58:42 ERR] Mutation not found, this should not happen
[20:58:42 ERR] Mutation not found, this should not happen
[...]

PR #1940, opened 2022-03-03
Fix Null Reference Exception during mutation rollback

dupdob:
make sure mutations are only rolled back once. Added regression test
see issue #1937


rouke-broersma:
I think the code coverage is a bug in our pipeline, you can ignore it.

dupdob:
yes, I know those lines are covered

richardwerkman:
Thanks for fixing this so quick!

PR #1956, opened 2022-03-09
fix: Version update check at startup no longer recommends upgrading to preview and latest

0xced:
- Only check for a preview release if a stable release is _not_ available in order to avoid having two messages inviting to upgrade to two different versions of Stryker.NET
- Actually check for a preview release instead of just the last release (versions.Max() would get the last release)

Also, use the `NuGet.Protocol` package to retrieve the list of versions so that Stryker.NET can benefit from the global NuGet caching mechanism.


rouke-broersma:
Thank you :)

Issue #1665, opened 2021-08-18
Stryker freezes with too many concurrent test runners


psfinaki:
**Describe the bug**

We are now executing Stryker during the night against a big codebase. Recently we switched to very strong machines and Stryker suddenly started to "freeze" during executions. After a few days of frustration I figured out that concurrency is the issue. The machines we use now have 16 logical processors so Stryker spins up 8 concurrent test runners and that occasionally (actually more often than not) just leeds to freezes in random places.

I don't know where the problem exactly is. I can figure out the exact number of test runners when things start going bad. Now I only know the 5 are ok and 8 are not ok. For us the execution time is the same for 4 and 5 concurrent test runners so we stopped there.

**Expected behavior**

No freezes, well.

**Workarounds**

Thanks god Stryker has an option to specify this amount in the config. This saved my ass :D

rouke-broersma:
We have actually experienced this ourselves before but only when using the solution run or multiple test projects feature! We had never had a user report before as far as I know.

We have a bug report open for this with the vstest team 😉😉😉

psfinaki:
Well we don't use the solution run, rather I wrote a script back in a day to glue project runs together heh :D 

I will try to get VsTest logs and verify then. So we could push together :)

rouke-broersma:
Well good because solution runs are **not** recommended (because of this problem but also because of inaccuracies).

richardwerkman:
This does sound different from issue #1136. We have crashing testrunners there. So lowering the number of testrunners should *increase* the chance of stryker hanging. But you could check. If you find this in strykers logs (not vstest logging) it's the same issue: `[Error] Cancelling the operation as requested.`

psfinaki:
@rouke-broersma @richardwerkman okay sorry it took so much time, I was implementing mutation score lock in solution wide and it took a while :D 

Now I got back to this, seems like VsTest is indeed the issue. I cannot collect VsTest logs from the build machine (I guess for that I would need to execute custom Stryker version there?) but I collected Stryker logs for hung executions and I see this `[Error] Cancelling the operation as requested.` there.

Should we try pinging the guys internally?

rouke-broersma:
If you could do that, that would be amazing :)

psfinaki:
Oh alright. Thanks! I'll see what I can do. This is maybe causing us problems even with running unit tests, we're investigating this now. I'll keep you posted on that.

richardwerkman:
[...]

Issue #1136, opened 2020-07-09
Stryker hangs when using the multi test project feature


richardwerkman:
**Describe the bug**
When using stryker with 2 or more testprojects the run seems to stop (without exiting) during the testing mutant phase. 

**Logs**
Will add soon

**Expected behavior**
The run should continue without any interuption.

**Desktop (please complete the following information):**
 - OS: Windows
 - Type of project: Core
 - Framework Version Core 2.1
 - Stryker Version 0.18

**Additional context**
The problem does not always occur, so this could prove hard to debug

richardwerkman:
After further investigation it looks like the vstest runners crash one by one. After the last one has crashed stryker keeps waiting forever.

It's not clear why the vstest runners crash. The last output in the logs contain:
```
2020-07-09T21:06:07.0150805+02:00  [DBG] Runner 1: Each mutant's fate has been established, we can stop. (b49f87dc)
2020-07-09T21:06:07.0373632+02:00  [VRB] Runner 1: [Error] Cancelling the operation as requested. (3478123c)
2020-07-09T21:06:07.0374326+02:00  [VRB] Runner 1: [Error] Cancelling the operation as requested. (3478123c)
```
This is the same for each runner.

ohoward2001:
Any update on when this issue might be sorted? Thanks

richardwerkman:
I'm sorry I don't know. This is not in our hands. We addressed this issue @ vstest almost two years ago. Currently a fix is ready to be merged but I'm not in the position to know when this release will take place. I hope soon. Once vstest releases the fix we will update stryker immediately

psfinaki:
Small insider update: this is hoped to be deliver along with VS2022 17.2. When is that - I don't know :D But now it's 17.0.5 so maybe not so long from now.

rouke-broersma:
That's awesome :) Hopefully by then we can update since we're currently blocked on #1870 due to https://github.com/microsoft/vstest/issues/3144 :D

PR #1967, opened 2022-03-17
Add timeouts to workaround frozen VsTest

dupdob:
Add Stryker controlled timeout for VsTest to detect frozen VsTest runs and retry them. Fix test runner frozen after test discovery.
May fix: #1665 , #1136 ,


dupdob:
Real world tests with NFluent reveal many issues. I prefer to cancel this PR and start over.

PR #1987, opened 2022-04-04
fix(mutating): Better support for mutating local function

### LINKED ISSUE #1986: Exception thrown by LocalFunctionStatementOrchestrator in CLI 1.5.0
**Describe the bug**
When I have a local function defined in a block that also has a block level mutation for another part, the local function tries to implement the mutation and falls over (an InvalidOperationException in RoslynHelper). I suspect that LocalFunctionStatementOrchestrator should have the same override of PrepareContext as AnonymousFunctionExpressionOrchestrator (this fixes the problem in my example), but I don't know the code well enough to know if this is truly the right fix.

**Additional context**
Please see the attached zip file (containing a small VS 2022 solution) for a replication of the bug.

[BreakStryker.zip](https://github.com/stryker-mutator/stryker-net/files/8404699/BreakStryker.zip)

dupdob:
Prevent migration of mutations from a method to a locally defined function as those lead to errors. 
Significant rewrite for 'LocalFunctionStatementOrchestrator' and 'AnonymousFunctionExpressionOrchestrator'.
Isolated the migration of the expression body (if present) to ensure a proper handling of the mutation context which prevents mutation migrating to other places.
Should also resolve some compile error mutant (see `CsharpMutantOrchestratorTests`: previous version of the test shows that the generated code will not compile)

Fix #1986

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/CsharpMutantOrchestratorTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/CsharpMutantOrchestratorTests.cs
@@ -145,8 +145,7 @@ namespace StrykerNet.UnitTest.Mutants.TestResources
 void SomeMethod()  => (true && SomeOtherMethod(out var x)) ? x : 5;
 }";
             string expected = @"void TestMethod(){if(StrykerNamespace.MutantControl.IsActive(0)){}else{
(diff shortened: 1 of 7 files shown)


rouke-broersma:
The windows integration test seems to be failing due to missing dotnet framework 4.6.1 targets, weird.

dupdob:
I am not sure why you think it is weird....
 My guess is that the distribution for Net 4.6.1 has been quietly faded by MS, so now the integration run will fail until the dependency is updated (via merging of my other PR, hopefully).
If you are confused because of my remark that integration tests were no longer executed by the CI, sorry, this was an improper description. The integration test step is executed, but 0 unit test is executed as the filtering condition is invalid (and results in no match).

dupdob:
also, I am close to getting an improved version for this PR. I will flag this one as draft

rouke-broersma:
Ah I forgot that I had not yet reviewed and merged the other pr.. Sorry. I was confused because I had thought this was already fixed.

dupdob:
Not sure what just happened: I clicked on 'ready for review' for the PR to leave the draft status and it closed it instead.
So I reopened it.
I will update the PR description to reflect the new version (which is FAR better).

dupdob:
Did a successful run against NFluent.

PR #1994, opened 2022-04-07
fix(since): Add possibility to specify a git tag for the since.target option

0xced:
In addition to a branch or a commit.


dupdob:
thanks for PR. Sadly, three unit tests are failing at this stage.
Could be please fix the code and/or tests for them to be green ? Updating to the master/head would be great also if you can do it alongside the fixes.

0xced:
I just rebased on master and fixed the tests, everything is ✅ now. 😀

PR #2007, opened 2022-04-21
docs(getting-started): manifest should be created in an unit test project

simondel:


rouke-broersma:
We support multiple test projects these days, in which case you will want to run stryker from the target project instead. In that case you will need to install stryker to the target project and not the unit test project. We also have plans to fix support for solution runs in which case you may want to install stryker to the solution folder (though I am not sure if this is supported by dotnet cli).

rouke-broersma:
We should write better documentation on the different ways to run stryker.

Issue #2016, opened 2022-04-26
Stryker throws an exception after initial test run


sobcvogl:
**Describe the bug**
If you run stryker it throws an exception after initial test run.
For Duplicated Tests using XUnit an Exception is Thrown.

```
Unhandled exception. System.InvalidOperationException: Sequence contains no elements
   at System.Linq.ThrowHelper.ThrowNoElementsException()
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source)
   at Stryker.Core.TestRunners.VsTest.VsTestDescription.get_InitialRunTime() in /_/src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestDescription.cs:line 42
   at Stryker.Core.TestRunners.VsTest.VsTestRunner.<>c.<BuildTestRunResult>b__23_3(VsTestDescription t) in /_/src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs:line 212
   at System.Linq.Enumerable.Sum[TSource](IEnumerable`1 source, Func`2 selector)
   at Stryker.Core.TestRunners.VsTest.VsTestRunner.BuildTestRunResult(IRunResults testResults, Int32 expectedTests, Boolean compressAll) in /_/src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs:line 212
   at Stryker.Core.TestRunners.VsTest.VsTestRunner.InitialTest() in /_/src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs:line 99
   at Stryker.Core.TestRunners.VsTest.VsTestRunnerPool.InitialTest() in /_/src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunnerPool.cs:line 57
   at Stryker.Core.Initialisation.InitialTestProcess.InitialTest(StrykerOptions options, ITestRunner testRunner) in /_/src/Stryker.Core/Stryker.Core/Initialisation/InitialTestProcess.cs:line 44
   at Stryker.Core.Initialisation.InitialisationProcess.InitialTest(StrykerOptions options) in /_/src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs:line 101
[...]

PR #2020, opened 2022-04-27
Add default duration for not executed duplicated xunit tests

sobcvogl:
Fix exception for duplicated XUnit tests, which are not executed.
#2016


sobcvogl:
#2016

Issue #2028, opened 2022-05-03
.NET Framework 4.7.2 Exception - Not enough information to troubleshoot


Christian-Oleson:
**Describe the bug**
Styker is failing in a .NET Framework project. I do not have enough information to understand what the issue is nor to troubleshoot. I followed the docs found here: 

**Logs**
dotnet tool run dotnet-stryker

   _____ _              _               _   _ ______ _______  
  / ____| |            | |             | \ | |  ____|__   __| 
 | (___ | |_ _ __ _   _| | _____ _ __  |  \| | |__     | |    
  \___ \| __| '__| | | | |/ / _ \ '__| | . ` |  __|    | |    
  ____) | |_| |  | |_| |   <  __/ |    | |\  | |____   | |    
 |_____/ \__|_|   \__, |_|\_\___|_| (_)|_| \_|______|  |_|    
                   __/ |                                      
                  |___/                                       

Version: 1.5.3

[15:14:07 INF] Identifying project to mutate.
[15:14:08 INF] Time Elapsed 00:00:01.6935843
Unhandled exception. System.InvalidOperationException: Sequence contains no matching element
   at System.Linq.ThrowHelper.ThrowNoMatchException()
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
   at Stryker.Core.Initialisation.ProjectFileReader.SelectAnalyzerResult(IAnalyzerResults analyzerResults, String targetFramework) in /_/src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs:line 107
   at Stryker.Core.Initialisation.ProjectFileReader.AnalyzeProject(String projectFilePath, String solutionFilePath, String targetFramework) in /_/src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs:line 55
   at Stryker.Core.Initialisation.InputFileResolver.ResolveInput(StrykerOptions options) in /_/src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs:line 62
   at Stryker.Core.Initialisation.InitialisationProcess.Initialize(StrykerOptions options) in /_/src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs:line 64
[...]

PR #2030, opened 2022-05-04
fix(initialisation): Improve error messages when we fail to correctly analyze the project

Christian-Oleson:


dupdob:
I am afraid you somehow broke the logic: it seems integration tests are now failing raising errors on project analysis, see [logs here](https://dev.azure.com/stryker-mutator/Stryker/_build/results?buildId=9811&view=logs&j=99031f74-bf3d-58da-997c-4e095f08dc64&t=b79c55ad-f33a-5882-eb36-5a64c1e39811&l=1335)

dupdob:
Did you try your change on your failing project ? I am asking because judging from issue #2028, the exception is triggered [here](https://github.com/stryker-mutator/stryker-net/blob/f295c0f78674bb5526579c13ab7c16ee1c1c021f/src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs#L107), hence before the code you changed. 
It seems to me the logic of the `SelectAnalyzerResult` method is to:
1. If no specific framework is expected, return the first found analyzer results (with a non null framework description)
2. otherwise, find the first analyzer result with a matching Net framework
3. if there is no match, return the first result.

Based on #2028, you want to improve error reporting for when step one fails, which is not handled as of today: there is an assumption that BuildAlyzer will return at least one valid result at this stage.

your modification assumes an exception must be raised if step 2 fails, whereas it is expected and handled in the current version: step 2 failing relates to the user expecting an non available (or detected) framework version.

I hope I am clear.

Anyway, it seems to me that trying to confirm why Buildalayzer returns no proper result on your project is what we should focus on at this stage. Is your project publicly available ? or can you try and debug what happens here?

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
```diff
@@ -107,21 +107,29 @@ private IAnalyzerResult SelectAnalyzerResult(IAnalyzerResults analyzerResults, s
...
             }
 
-            var analyzerResultForFramework = analyzerResults.SingleOrDefault(result => result.TargetFramework == targetFramework);
+            var analyzerResultList = analyzerResults?.Where(result => result.TargetFramework == targetFramework).ToList();
+
+            switch (analyzerResultList?.Count)
+            {
+                case 0:
+                    throw new InputException($"The target framework of {targetFramework} has no analyzer results.");
+                case > 1:
```
psfinaki, line 116: So what do you need to have in your `<TargetFrameworks>` to hit this case?

Christian-Oleson: > I am afraid you somehow broke the logic: it seems integration tests are now failing raising errors on project analysis, see [logs here](https://dev.azure.com/stryker-mutator/Stryker/_build/results?buildId=9811&view=logs&j=99031f74-bf3d-58da-997c-4e095f08dc64&t=b79c55ad-f33a-5882-eb36-5a64c1e39811&l=1335). Can you fix your PR and make sure integration tests are passing?

I made a modification to this after @dupdob indicated there were failing integration tests. Hopefully, this passes. There are several ways to hit this, but as of right now I'll just start with the fact that analyzerResults can have no results. This causes an issue when we do First() in the IF statement. Likewise, you can pass in a target framework that doesn't match and I wasn't getting a validation error (should there be validation indicating if I flubbed a framework name, like nett6 as a typo example?

Christian-Oleson: I did try passing in net472 as an option. I've tried others. I don't see the framework options listed on the site, so I've assumed that the options are identical to the MS listed options? Likewise, I can see that it is picking up my project, but it gets No analyzer results. I'm obviously newer to the framework but was hoping to get this up and running to show the value add. Am I wrong in thinking that this project supports .NET  Framework? 

Based on [this doc](https://stryker-mutator.io/docs/stryker-net/getting-started#dotnet-framework-specific) I assume it is supported. Likewise, I know it isn't my nuget registry as I see the nuget requests in the logs when I have a trace level set for the logging.

Christian-Oleson: I'll start looking there then. Certainly, I have an edge case of MSBuild. I'm still desiring to have this project provide more information about the condition that was hit, if possible. It makes me wonder less about why I'm seeing something that I believe shouldn't be exceptional.

rouke-broersma: I believe we can enable debug logging for buildalyzer. We have not enabled this yet because it logs the complete build output of MSBuild but perhaps we should include this as part of trace logging so we can better understand these cases. You could enable this on your local project: https://github.com/daveaglick/Buildalyzer#logging

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
```diff
@@ -104,24 +104,29 @@ private IAnalyzerResult SelectAnalyzerResult(IAnalyzerResults analyzerResults, s
         {
             if (targetFramework == null)
             {
-                return analyzerResults.First(e => e.TargetFramework != null);
+                var analyzerResult = analyzerResults.FirstOrDefault(e => e.TargetFramework != null);
+
+                if (analyzerResult == null)
```
rouke-broersma, line 109:
```suggestion
                if (analyzerResult is null)
```

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
```diff
@@ -104,24 +104,29 @@ private IAnalyzerResult SelectAnalyzerResult(IAnalyzerResults analyzerResults, s
...
+                {
+                    throw new InputException("No analyzer result found and the targetFramework was null.");
+                }
+
+                return analyzerResult;
             }
 
             var analyzerResultForFramework = analyzerResults.SingleOrDefault(result => result.TargetFramework == targetFramework);
             if (analyzerResultForFramework != null)
```
rouke-broersma, line 118:
```suggestion
            if (analyzerResultForFramework is not null)
```

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
```diff
@@ -104,24 +104,29 @@ private IAnalyzerResult SelectAnalyzerResult(IAnalyzerResults analyzerResults, s
         {
             if (targetFramework == null)
             {
-                return analyzerResults.First(e => e.TargetFramework != null);
+                var analyzerResult = analyzerResults.FirstOrDefault(e => e.TargetFramework is not null);
+
+                if (analyzerResult is null)
+                {
+                    throw new InputException("No analyzer result found and the targetFramework was null.");
```
rouke-broersma, line 111:
```suggestion
                    throw new InputException("No analyzer result with a valid target framework could be found.");
```

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
```diff
@@ -104,24 +104,29 @@ private IAnalyzerResult SelectAnalyzerResult(IAnalyzerResults analyzerResults, s
...
-            {
-                var firstAnalyzerResult = analyzerResults.First();
-                _logger.LogWarning(
-                    "The configured target framework '{0}' isn't available for this project. " +
-                    "It will be built against the first framework available " +
-                    "which is {1}.", targetFramework, firstAnalyzerResult.TargetFramework);
 
-                return firstAnalyzerResult;
-            }
+            var firstAnalyzerResult = analyzerResults.First();
```
rouke-broersma, line 123:
```suggestion
            var firstAnalyzerResult = analyzerResults.First(e => e.TargetFramework is not null);
```

We cannot test against a analyzer result that does not have a valid target framework.

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
```diff
@@ -104,24 +104,29 @@ private IAnalyzerResult SelectAnalyzerResult(IAnalyzerResults analyzerResults, s
...
-            {
-                var firstAnalyzerResult = analyzerResults.First();
-                _logger.LogWarning(
-                    "The configured target framework '{0}' isn't available for this project. " +
-                    "It will be built against the first framework available " +
-                    "which is {1}.", targetFramework, firstAnalyzerResult.TargetFramework);
 
-                return firstAnalyzerResult;
-            }
+            var firstAnalyzerResult = analyzerResults.First(e => e.TargetFramework is not null);
```
rouke-broersma, line 123: We can extract https://github.com/stryker-mutator/stryker-net/pull/2030/files#diff-b73ef193446dc576c0b7b41967963436819ee09852c44e4c10d4e1df4e4d726fR107-R114 into a function and reuse it here, since we should once again test if we can actually find a valid analyzer result and if not we should throw.

rouke-broersma: If we invert the if (targetFramework == null) perhaps we can avoid this duplicate check completely.

dupdob: yes, logic could be:
1. if a specific `targetframework` is expected
    1. try to find a matching result
    2. if none find, log a warning: `The configured target...
1. find the first result with a valid framework
    1. if not, raise an exception

Christian-Oleson: I started implementing that logic and instead found a way to simplify. Please let me know.

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
```diff
@@ -102,26 +102,24 @@ private void LogAnalyzerResult(IAnalyzerResult analyzerResult)
 
         private IAnalyzerResult SelectAnalyzerResult(IAnalyzerResults analyzerResults, string targetFramework)
         {
+            var firstAnalyzerResult = analyzerResults?.First(e => e.TargetFramework is not null);
             if (targetFramework == null)
```
rouke-broersma, line 106:
```suggestion
            if (targetFramework is null)
```

Christian-Oleson: My bad. You've made this suggestion before and I didn't take care of it. 

On another note, good heavens you are fast.

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
rouke-broersma: Looks good to me

src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
rouke-broersma: I can also be very slow :)

rob-mur:
I've ran into a case where no error message is reported still, namely when analyzerResults is empty. The first line of SelectAnalyzerResults then calls First from Linq which leads to a crash.

I couldn't easily work out why analyzerResults was empty for my project as I'm not familiar with the source code of Stryker, but happy to provide extra information if guidance is provided.

rouke-broersma:
> I've ran into a case where no error message is reported still, namely when analyzerResults is empty. The first line of SelectAnalyzerResults then calls First from Linq which leads to a crash.
> 
> I couldn't easily work out why analyzerResults was empty for my project as I'm not familiar with the source code of Stryker, but happy to provide extra information if guidance is provided.

I missed that in review. Good point. Do you think you could provide the fix? Could you also open a new issue with the details about your project and that Buildalyzer fails to properly analyze it?

rob-mur:
Sure, I've provided a basic fix [here](https://github.com/stryker-mutator/stryker-net/pull/2037).

I'll open an issue with my particular issue!

PR #2032, opened 2022-05-05
2031 Target .NET 6 within the Stryker Projects

Christian-Oleson:
This is a breaking change! Please review thoroughly. 
[x] Unit Tests Pass
[?] Integration Tests?

### DIFF
--- a/src/Stryker.CLI/Stryker.CLI.UnitTest/Stryker.CLI.UnitTest.csproj
+++ b/src/Stryker.CLI/Stryker.CLI.UnitTest/Stryker.CLI.UnitTest.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
--- a/src/Stryker.CLI/Stryker.CLI/packages.lock.json
+++ b/src/Stryker.CLI/Stryker.CLI/packages.lock.json
@@ -1,7 +1,7 @@
 {
   "version": 1,
   "dependencies": {
-    ".NETCoreApp,Version=v5.0": {
+    "net6.0": {
       "DotNet.ReproducibleBuilds": {
         "type": "Direct",
         "requested": "[1.1.1, )",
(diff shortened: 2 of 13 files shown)


src/Stryker.CLI/Stryker.CLI/packages.lock.json
```diff
@@ -1181,12 +1160,12 @@
...
         "dependencies": {
-          "Buildalyzer": "4.1.3",
+          "Buildalyzer": "4.1.4",
           "DotNet.Glob": "3.1.3",
           "FSharp.Compiler.Service": "38.0.0",
           "LibGit2Sharp": "0.27.0-preview-0182",
           "Microsoft.CodeAnalysis.CSharp": "4.1.0",
-          "Microsoft.Extensions.Configuration": "6.0.0",
+          "Microsoft.Extensions.Configuration": "6.0.1",
           "Microsoft.Extensions.Configuration.Json": "6.0.0",
```
rouke-broersma, line 1169: We will get these version updates automatically, you do not need to include them in this PR unless they are required for .net 6.

Christian-Oleson: I'll revert those then. 😄

Christian-Oleson: done

src/Stryker.CLI/Stryker.CLI.UnitTest/Stryker.CLI.UnitTest.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
rouke-broersma, line 4: Perhaps while we're at it we should move the version to Directory.build.props so we have one consistent version number for all projects.

rouke-broersma:
I created a new PR #2036 for this change because some of our checks such as sonarcloud cannot run on contributor PR's. Normally that isn't such a big deal but for this change I would prefer that we run the full pipeline. Thank you for reminding us that this is imminent.

Christian-Oleson:
> I created a new PR #2036 for this change because some of our checks such as sonarcloud cannot run on contributor PR's. Normally that isn't such a big deal but for this change I would prefer that we run the full pipeline. Thank you for reminding us that this is imminent.

Understandable. Thanks!

PR #2037, opened 2022-05-10
fix(initialisation): Improve error message when no analyzer results are found

rob-mur:
Adds a simple catch for if no analyzerResults are found.


rouke-broersma:
@rob-mur I thought the solution was getting a bit complex so I simplified it. Do you agree with this solution?

PR #2055, opened 2022-05-13
fix(Reporting): Fix issue with relative paths that miss the root folder

### LINKED ISSUE #1380: Html reporter doesn't work for multi project runs
**Describe the bug**
The html reporter seems to just put all files from all projects under the same directory. Possibly even overwriting duplicate file names. 

Also the cache used by the incremental compare feature is overwritten.

**Expected behavior**
I want every project to have it's own directory in the html report so I can inspect each project.

richardwerkman:
closes #1380


src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
```diff
@@ -15,6 +15,9 @@ public class StrykerOptions
         public bool DevMode { get; init; }
 
         public string BasePath { get; init; }
+
+        private readonly string RootPathField;
+        public string RootPath { get => RootPathField ?? BasePath; init => RootPathField = value; }
```
rouke-broersma, line 20:
```suggestion
        private readonly string _rootPath;
        public string RootPath { get => _rootPath?? BasePath; init => _rootPath= value; }
```

Issue #2008, opened 2022-04-22
Stryker silently runs with default config if the custom is not found


psfinaki:
**Describe the bug**
So the use case is that we've just spent some time on why the custom configuration is not used until we noticed a typo in the path to it :)

**Expected behavior**
 I'd expect an error or at least some kind of warning for this.

**Desktop**
 Stryker Version: 1.5.2

**Additional context**
Repro: [Repro.zip](https://github.com/stryker-mutator/stryker-net/files/8541221/Repro.zip)
You can run it like from the folder like `dotnet stryker -f ..\blah.json` :)

PR #2064, opened 2022-05-19
fix(options): Stryker silently runs with default config if custom config file is set but not found

ggingras:
To fix this issue, I thought that an InputException should only be thrown if the invalid config was provided by the user.

If no config is provided and the default config is not present, do as before and silently continue.

Note: This is my first PR for the Stryker project, do not hesitate to suggest changes.

### DIFF
--- a/src/Stryker.CLI/Stryker.CLI/ConfigReader.cs
+++ b/src/Stryker.CLI/Stryker.CLI/ConfigReader.cs
@@ -1,5 +1,6 @@
 using System.IO;
 using McMaster.Extensions.CommandLineUtils;
+using Stryker.Core.Exceptions;
 using Stryker.Core.Options;
 
 namespace Stryker.CLI
@@ -24,12 +25,20 @@ namespace Stryker.CLI
             var basePath = Directory.GetCurrentDirectory();
             inputs.BasePathInput.SuppliedInput = basePath;
 
+            var configFileOption = cmdConfigHandler.GetConfigFileOption(args, app);
+
             // read config from json and commandline
-            var configFilePath = Path.Combine(basePath, cmdConfigHandler.GetConfigFilePath(args, app));
+            var configFilePath = Path.Combine(basePath, configFileOption?.Value() ?? "stryker-config.json");
             if (File.Exists(configFilePath))
             {
                 JsonConfigHandler.DeserializeConfig(configFilePath, inputs);
             }
+            else if (configFileOption.HasValue())
+            {
+                // only throw exception if the config file path provided by the user
+                throw new InputException($"Config file not found {configFilePath}");
+            }
+
             cmdConfigHandler.ReadCommandLineConfig(args, app, inputs);
         }
     }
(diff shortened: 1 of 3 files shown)


rouke-broersma:
Hi @ggingras 

Could you explain the problem? If no custom config is provided stryker will already use defaults. Am I misunderstanding the scenario?

ggingras:
> Hi @ggingras
> 
> Could you explain the problem? If no custom config is provided stryker will already use defaults. Am I misunderstanding the scenario?

It is to address this issue: https://github.com/stryker-mutator/stryker-net/issues/2008 log by another user, but I also faced the same issue while working with Stryker. In my use case, it was a relative path issue.

src/Stryker.CLI/Stryker.CLI/ConfigReader.cs
```diff
@@ -24,12 +25,20 @@ public void Build(IStrykerInputs inputs, string[] args, CommandLineApplication a
...
-            var configFilePath = Path.Combine(basePath, cmdConfigHandler.GetConfigFilePath(args, app));
+            var configFilePath = Path.Combine(basePath, configFileOption?.Value() ?? "stryker-config.json");
             if (File.Exists(configFilePath))
             {
                 JsonConfigHandler.DeserializeConfig(configFilePath, inputs);
             }
+            else if (configFileOption.HasValue())
+            {
+                // only throw exception if the config file path provided by the user
+                throw new InputException($"Config file not found {configFilePath}");
```
psfinaki, line 39:
```suggestion
                throw new InputException($"Config file not found at {configFilePath}");
```

src/Stryker.CLI/Stryker.CLI/ConfigReader.cs
```diff
@@ -24,12 +25,20 @@ public void Build(IStrykerInputs inputs, string[] args, CommandLineApplication a
...
             // read config from json and commandline
-            var configFilePath = Path.Combine(basePath, cmdConfigHandler.GetConfigFilePath(args, app));
+            var configFilePath = Path.Combine(basePath, configFileOption?.Value() ?? "stryker-config.json");
             if (File.Exists(configFilePath))
             {
                 JsonConfigHandler.DeserializeConfig(configFilePath, inputs);
             }
+            else if (configFileOption.HasValue())
+            {
+                // only throw exception if the config file path provided by the user
```
psfinaki, line 38:
```suggestion
                // only throw exception if the config file path is provided by the user
```

psfinaki (dismissed):
Thanks a lot!

src/Stryker.CLI/Stryker.CLI/ConfigReader.cs
```diff
@@ -24,12 +25,20 @@ public void Build(IStrykerInputs inputs, string[] args, CommandLineApplication a
...
             // read config from json and commandline
-            var configFilePath = Path.Combine(basePath, cmdConfigHandler.GetConfigFilePath(args, app));
+            var configFilePath = Path.Combine(basePath, configFileOption?.Value() ?? "stryker-config.json");
             if (File.Exists(configFilePath))
             {
                 JsonConfigHandler.DeserializeConfig(configFilePath, inputs);
             }
+            else if (configFileOption.HasValue())
+            {
+                // only throw exception if the config file path is provided by the user
```
rouke-broersma, line 38:
```suggestion
                // throw exception if the config file path is provided by the user yet it does not exist
```

PR #2081, opened 2022-06-12
Document 6.0 SDK requirement for app under test

eatdrinksleepcode:
Existing applications which are built using an SDK prior to 6.0 may not build under 6.0, which will prevent Stryker from analyzing them. For example, an SDK project which still explicitly includes assembly info attributes in its source and does not explicitly define the "GenerateAssemblyInfo" MSBuild property may build successfully using the 3.1 SDK, but will fail with the 6.0 SDK with "Duplicate attribute" errors.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -9,9 +9,9 @@ custom_edit_url: https://github.com/stryker-mutator/stryker-net/edit/master/docs
 
 Stryker is installed using the dotnet cli as a [dotnet tool](https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools)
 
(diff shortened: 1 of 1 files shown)


rouke-broersma:
Hi David,

Could you also open an issue with the details of why this doesn't work currently? This is not intentional and not related to the runtime requirement.

eatdrinksleepcode:
Sure! I'll see if I can create a minimal repo.

eatdrinksleepcode:
The problem is not what I originally thought it was, my project builds with 6.0 but does _not_ build with Stryker. I have an idea of why and will open an issue when I have minimized the repro.

Issue #1582, opened 2021-05-29
Add more mutants to the UpdateOpeartors


HakamFostok:
The [Update Operators](https://stryker-mutator.io/docs/stryker-net/Mutators#update-operators) just reverse the operator in its place, so ++ will be -- in the same way -- will be ++

this is good but should the following mutants be added?

Original              Mutant
variable++  => ++variable
++variable  => variable++
variable--    => --variable
--variable  => variable--

I know it could produce more equivalent mutants, 
I do not know I just thought about that, feel free to close this if this will not consider.

Thank you this is great library

dupdob:
The main problem is that incrementor/decrementor are used in various contexts, and are often used on their own (in a `for` statement for example).
`for(...;...;i++)` is the same than `for(...;...;++i)`
It could make sense for when the value is also used, as a parameter:
`f(i++);` is not the same than `f(++i);`
or as the right hand expression of an assignment
`var x = i++;` is not the same then `var x = ++i;`

so it should be a specific mutator, more than an extension of the existing one

HakamFostok:
@dupdob Yes I totally agree with you, wish this could be available,
by the way, thank you very much for this great library

Liam-Rougoor:
Hi! I was just browsing through issues, and would like to put my hands on this.

dupdob:
For info, we just added a markdown document explaining how to add a mutator:
https://github.com/stryker-mutator/stryker-net/blob/master/adding_a_mutator.md

PR #2092, opened 2022-06-17
feat(unary mutators): add unary to inverse mutator (#1582)

Liam-Rougoor:


Liam-Rougoor:
I originally extended the existing unary mutators, but noticed in the issue that this way is incorrect. Closing this PR.

Issue #2100, opened 2022-06-28
YAML configuration for stryker


chripf:
When creating a stryker-config, it would be helpful to add comments e.g. to explain an "ignore-methods" entry. `json` however does not allow comments.

Having the option to use a `stryker-config.yaml` configuration will allow a leaner structure of the configuration file (a matter of taste, admittedly) and the option to  add comments while staying valid.
Example:

```
---
stryker-config:
  ignore-methods:
  - "*Log" # ignore log mutations
  - Console.Write* # do not care about command line output
  - "*Exception.ctor"
  - ApiError.ctor
  project: My.Test.Project
```

rouke-broersma:
This shouldn't be too hard to implement :)

PR #2101, opened 2022-06-28
allow loading of yaml-config

chripf:
Provides a solution for #2100

### DIFF
--- a/src/Stryker.CLI/Stryker.CLI.UnitTest/Stryker.CLI.UnitTest.csproj
+++ b/src/Stryker.CLI/Stryker.CLI.UnitTest/Stryker.CLI.UnitTest.csproj
@@ -28,6 +28,7 @@
 
   <ItemGroup>
     <None Include="filled-stryker-config.json" CopyToOutputDirectory="Always" />
+    <None Include="filled-stryker-config.yaml" CopyToOutputDirectory="Always" />
     <None Include="stryker-config.json" CopyToOutputDirectory="Always" />
   </ItemGroup>
 </Project>
--- a/src/Stryker.CLI/Stryker.CLI.UnitTest/YamlConfigFileTests.cs
+++ b/src/Stryker.CLI/Stryker.CLI.UnitTest/YamlConfigFileTests.cs
@@ -1,0 +1,63 @@
+using System.IO;
+using Microsoft.Extensions.Logging;
+using Moq;
+using Shouldly;
+using Stryker.Core;
+using Stryker.Core.Initialisation;
+using Stryker.Core.Options;
+using Stryker.Core.Reporters;
+using Xunit;
+
+namespace Stryker.CLI.UnitTest
+{
+    public class YamlConfigFileTests
+    {
+
+
+        [Fact]
+        public void YamlConfigFile_IsLoadedCorrectly()
+        {
+            IStrykerInputs actualInputs = null;
+            var options = new StrykerOptions()
+            {
+                Thresholds = new Thresholds()
+                {
+                    High = 80,
+                    Low = 60,
+                    Break = 0
+                }
+            };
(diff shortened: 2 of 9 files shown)


rouke-broersma:
Can you rename the config loader etc to something like file config loader since it's no longer json specific?

Also I remember reading that yaml is a superset of json, does that perhaps mean that the yaml serializer can deserialize json as well as yaml? We could then have one single implementation.

chripf:
> Can you rename the config loader etc to something like file config loader since it's no longer json specific?
Good point. Did that. 
> Also I remember reading that yaml is a superset of json, does that perhaps mean that the yaml serializer can deserialize json as well as yaml? We could then have one single implementation.
This seems possible.  Funny enough, comments in json seem to be a problem here. Will find a way around that, but I have to agree it seems better to have a single parser here.

chripf:
Did a partial solution in https://github.com/stryker-mutator/stryker-net/pull/2101/commits/d3102ff357aa08422aa9e4e26dc1e1551a0cd7fc, but this cannot handle trailing comments, which was possible before (not part of the test file thou). I have added such a comment to the json-configfile in the test (https://github.com/stryker-mutator/stryker-net/pull/2101/commits/5d51769d8e1f90f89a1d7dd853cbddb0827b8093) and suggest find a way to handle this too.

src/Stryker.CLI/Stryker.CLI/ConfigHandler.cs
```diff
@@ -0,0 +1,126 @@
...
+            inputs.IgnoreMutationsInput.SuppliedInput = config.IgnoreMutations;
+            inputs.IgnoredMethodsInput.SuppliedInput = config.IgnoreMethods;
+
+            inputs.ReportFileNameInput.SuppliedInput = config.ReportFileName;
+        }
+
+        private static FileBasedInput LoadConfig(string configFilePath)
+        {
+            var lines = File.ReadAllLines(configFilePath);
+            // filter json comments
```
psfinaki, line 69: So I'm confused a bit: here you write that we can't handle trailing comments but then you add a trailing comment in the test json config file - and things work?

rouke-broersma:
@chripf unfortunately it looks like json deserializing using the yaml deserializer actually does not work, that's why the unit tests are failing. I have reverted to your initial solution with separate deserializers for json and yaml.

src/Stryker.CLI/Stryker.CLI/ConfigHandler.cs
rouke-broersma: I have solved this by reverting to separate deserializers.

src/Stryker.CLI/Stryker.CLI/FileConfigReader.cs
```diff
@@ -0,0 +1,145 @@
...
+        private static FileBasedInput LoadConfig(string configFilePath)
+        {
+            using var streamReader = new StreamReader(configFilePath);
+            var fileContents = streamReader.ReadToEnd();
+
+            FileBasedInput input;
+            try
+            {
+                FileBasedInputOuter root;
+                if (configFilePath.EndsWith(".yaml") || configFilePath.EndsWith(".yml"))
```
psfinaki, line 75: Probably this could be a case for a OOP exercise - like create some interface and some polymorphism :)

But either way thanks for this functionality.

rouke-broersma:
Thanks for contributoring @chripf !!

chripf:
Thanks for your adaptions @rouke-broersma , did not have time to finish my ideas, but it is a good thing to have others looking into it anyway.

I did make a solution to handle trailing comments while using the yaml reader (by removing them before passing it to the yaml reader). Maybe will look into it another time. For now having this separated in two methods seems nice and makes it easy for some OOP refactoring.

Issue #2126, opened 2022-08-08
Detect test projects not supported by VsTest


dupdob:
**Is your feature request related to a problem? Please describe.**
At least one user had troubles with Stryker because its test project was missing key references for proper VsTest support.

**Describe the solution you'd like**
Stryker should properly detect such scenario and guide users toward resolution. Arguably, Stryker could automatically inject needed dependencies.

**Describe alternatives you've considered**
This problem may be rare and may not warrant development effort.

**Additional context**
This issue is open to discuss implementation strategy

dupdob:
My original intent was to perform a dependency analysis when no test was detected at initialization; Stryker could then output a guidance message, such as `Add NUnit3TestAdapter package to project 'TestProject.csproj'`.
Alternatively, Stryker could fail when needed dependencies were not detected like it does for MsTest V1 projects.

rouke-broersma:
Imo we don't need to get too fancy here. 

If we don't find any of mstest testadapter, xunit adapter or nunit adapter (whatever the name is for each) we can fail with a recommendation to add the correct adapter for their used testframework(s). 

As far as I'm concerned we don't even have to check which test framework they use nor if test discovery is failing. We simply fail if there is not at least one test adapter. We can fine tune later and add checks for other test adapters that might be used if that ever happens. 

We could also add something to the getting started about requiring the test project be executable with dotnet test (though I'm not sure if that should only be for dotnet (core), I think dotnet framework tests can't be executed with dotnet test).

dupdob:
> if we don't find any of mstest testadapter, xunit adapter or nunit adapter (whatever the name is for each) we can fail with a recommendation to add the correct adapter for their used testframework(s).
I am not confortable with this, I think this is too restrictive. There are many test frameworks and several test adapters for the most popular ones ([see here](https://www.nuget.org/packages?q=testadapter)). So this logic would result in Stryker supporting only xUnit,NUnit or MSTest for no valid reason.
And if a popular framework updates the name of its adapter, we need to upgrade Stryker and impacted people must upgrade.

Hence my intent to do it only when there is an issue.

Alternatively, we could raise a warning message, such as: _"no known test adapter detected, please add the proper one if no test is detected"_.

My intent with being specific for NUnit, XUnit and MSTest was to provide automatic assistance to those concerned as an attempt to prevent them to open issues about this.

rouke-broersma:
Looking at that list I'm not as worried as you, those testadapters are near unused.

[...]

PR #2139, opened 2022-08-17
feat(testrunner): Diagnose failed test discovery and provide fix hints

dupdob:
Implements #2126:
when no tests are detected Stryker verifies if there is a test runner present for each tes framework and stops on exception otherwise.

Limit: this only triggers if no test is detected in any test projects (if more than one is present).

_Note: the implementation is in a dedicated method and be run at another point in time if need arises._

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs
+++ b/src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
@@ -41,7 +42,9 @@ namespace Stryker.Core.Initialisation
         private readonly IInitialTestProcess _initialTestProcess;
         private readonly IAssemblyReferenceResolver _assemblyReferenceResolver;
         private ITestRunner _testRunner;
+        private ProjectInfo _projectInfo;
         private readonly ILogger _logger;
+       
 
         public InitialisationProcess(
             IInputFileResolver inputFileResolver = null,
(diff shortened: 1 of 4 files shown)


src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs
```diff
@@ -79,26 +82,59 @@ public MutationTestInput Initialize(StrykerOptions options)
...
         }
 
-        public InitialTestRun InitialTest(StrykerOptions options) =>
+        public InitialTestRun InitialTest(StrykerOptions options)
+        {
             // initial test
-            _initialTestProcess.InitialTest(options, _testRunner);
+            var result = _initialTestProcess.InitialTest(options, _testRunner);
+
+            if (_testRunner.DiscoverTests().Count != 0)
```
rouke-broersma, line 107: We should rename this at some point because testrunner.discoverTests never actually runs test discovery anymore.

src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs
```diff
@@ -79,26 +82,59 @@ public MutationTestInput Initialize(StrykerOptions options)
...
+
+        public static void DiagnoseLackOfDetectedTest(ProjectInfo projectInfo)
+        {
+            foreach (var testProject in projectInfo.TestProjectAnalyzerResults)
+            {
+                foreach (var (framework, adapter) in TestFrameworks)
+                {
+                    if (testProject.References.Any(r => r.Contains(framework)) && !testProject.References.Any(r => r.Contains(adapter)))
+                    {
+                        throw new InputException($"Project '{testProject.ProjectFilePath}' is not supported by VsTest because it is missing an appropriate VstTest adapter for '{framework}'. " +
```
rouke-broersma, line 131: While I agree with this I am surprised you went with this option since you were so vehemently against it. I expected you to go with a warning instead of an error as discussed. Nevertheless, I don't expect this to be a big problem.

src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs
rouke-broersma, line 107: We should rename this at some point because testrunner.discoverTests never actually runs test discovery anymore.

rouke-broersma (approved):

src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs
dupdob: This is exactly what I intended all along: as of now, Stryker stops if no test is detected and provides hints for known/supported test libraries.
I was against doing this preventively, i.e. before trying to detect some tests.
I sensed we had trouble understanding each other, hence I provided this PR as a better basis for conversation 😄 .

src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs
dupdob: yes, felt silly typing this.

src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs
rouke-broersma: Ah I see, fair point :)

PR #2143, opened 2022-08-22
fix: Stryker not properly flagging mutant as TimedOut

dupdob:
I found circumstances where Stryker does not properly marks TimedOut mutants (while working on diagnose mode).
This PR fixes this issue.

Note: current design for TimeOut handling is insufficiently cohesive. This comes from the fact that Stryker must handle situations where it cannot identify which test timed out, which happens when this is the very first test to be executed by VsTest (apparently).
An improved refactoring will be provided with diagnose mode. I made this call to avoid a complex merge in the coming weeks 😄 .

### DIFF
--- a/src/Stryker.Core/Stryker.Core/TestRunners/CoverageRunResult.cs
+++ b/src/Stryker.Core/Stryker.Core/TestRunners/CoverageRunResult.cs
@@ -1,7 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.Collections.Immutable;
-using System.Linq;
 
 namespace Stryker.Core.TestRunners
 {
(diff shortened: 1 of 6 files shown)


src/Stryker.Core/Stryker.Core/TestRunners/CoverageRunResult.cs
```diff
@@ -43,9 +41,6 @@ public class CoverageRunResult
 
         public CoverageConfidence Confidence { get; private set; }
 
-        public IReadOnlyCollection<int> LeakedMutations  => _mutationFlags
```
dupdob, line 46: Note: opportunistic improvement. This property is not used and should have been removed earlier.

PR #2149, opened 2022-08-26
fix(azure-baseline): Use full SAS token format instead of making assumptions and transformations on the token

### LINKED ISSUE #2148: AzureFileShareBaseline SAS Token Generation in Azure Devops
**Describe the bug**
AzureFileShareBaseline does not seem to be compatible with the SAS token generated by Azure CLI in Azure Devops

Example Azure CLI Task
```
    task: AzureCLI@2
    displayName: Create SAS Token for Stryker
    inputs:
      azureSubscription: ""
      scriptType: pscore
      scriptLocation: inlineScript
      inlineScript: |
        $end = date -u -d "30 minutes" '+%Y-%m-%dT%H:%MZ'

        $storageAccountKey = az storage account keys list -g ${{parameters.resourceGroup}} -n ${{parameters.storageName}} --query [0].value -o tsv
        $sasToken = az storage fs directory generate-sas --name stryker/ --file-system ${{parameters.storageName}} --account-name ${{parameters.storageName}} --account-key $storageAccountKey --expiry $end --https-only --permissions rwld -o tsv

        Write-Host "##vso[task.setvariable variable=ContainerSasToken]$sasToken"
        Write-Host "##vso[task.setvariable variable=ContainerSasToken;isOutput=true]$sasToken"
```

The above task generates the sas token sucesfully.
`se=2022-08-25T14%3A27Z&sp=rwdl&spr=https&sv=2021-06-08&sr=d&sdd=1&sig=XXXXXXXXXXXXX`

However when passing the token to Stryker, the token is not valid.
```
[13:57:58 ERR] Creating directory failed with status Forbidden and message 
<?xml version="1.0" encoding="utf-8"?>
   <Error>
      <Code>AuthenticationFailed</Code>
     <Message>Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
Time:2022-08-25T13:57:58.5748519Z
      </Message>
     <AuthenticationErrorDetail>Signature fields not well formed.</AuthenticationErrorDetail>
   </Error>
```

Believe the issue to either be with the token generation. If it is then how to generate the token correctly is not easy to find. Or an issue within AzureFileShareBaseline, specifically where the sas token is appended to the fileshare url. Appending the sas token to  `sv=`  is leading to an incorrect token.
```
var fileUrl = $"{_options.AzureFileStorageUrl}/{_outputPath}/{version}/stryker-report.json";
            var url = new Uri($"{fileUrl}?sv={_options.AzureFileStorageSas}");

            using var requestMessage = new HttpRequestMessage(HttpMethod.Get, url);
```

**Logs**
For us to best assist you in diagnosing the problem we would ask you to run stryker with the option `--log-to-file` or `-L` and to attach the resulting logs to the issue.

**Expected behavior**
Expect an Azure CLI SAS token to be valid when using AzureFileShareBaseline in Azure Devops

 - OS: Linux, Azure Devops
 - Type of project: Core
 - Framework Version: .Net 6
 - Stryker Version: 2.2.0

**Additional context**

tidusjar:
This PR stops Stryker from making the assumption that the SAS Token is in a specific format.

It will now take the SAS token that has been provided, do some basic validation to ensure that it has the correct parameters (SV & Sig) and will then pass it along to the file share operations.

One thing to note, the order of the query parameters has changed, but the outcome is still the same.

Fixes #2148

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Baseline/Providers/AzureFileShareBaselineProvider.cs
+++ b/src/Stryker.Core/Stryker.Core/Baseline/Providers/AzureFileShareBaselineProvider.cs
@@ -3,11 +3,13 @@ using Stryker.Core.Logging;
 using Stryker.Core.Options;
 using Stryker.Core.Reporters.Json;
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Text;
 using System.Threading.Tasks;
+using System.Web;
 
 namespace Stryker.Core.Baseline.Providers
 {
@@ -32,7 +34,8 @@ namespace Stryker.Core.Baseline.Providers
         public async Task<JsonReport> Load(string version)
         {
             var fileUrl = $"{_options.AzureFileStorageUrl}/{_outputPath}/{version}/stryker-report.json";
-            var url = new Uri($"{fileUrl}?sv={_options.AzureFileStorageSas}");
+
+            var url = GenerateUri(fileUrl, _options.AzureFileStorageSas).ToString();
 
             using var requestMessage = new HttpRequestMessage(HttpMethod.Get, url);
 
@@ -53,14 +56,15 @@ namespace Stryker.Core.Baseline.Providers
 
         public async Task Save(JsonReport report, string version)
         {
-            var fileUrl = $"{_options.AzureFileStorageUrl}/{_outputPath}/{version}/stryker-report.json";
-            var url = new Uri($"{fileUrl}?comp=range&sv={_options.AzureFileStorageSas}");
-
             var existingReport = await Load(version);
 
             var reportJson = report.ToJson();
 
             int byteSize = Encoding.UTF8.GetByteCount(report.ToJson());
+
+            var fileUrl = $"{_options.AzureFileStorageUrl}/{_outputPath}/{version}/stryker-report.json";
+
+            var url = GenerateUri(fileUrl, _options.AzureFileStorageSas, new Dictionary<string, string> { { "comp", "range" } }).Uri;
 
             if (existingReport == null)
             {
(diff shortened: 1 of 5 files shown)


rouke-broersma:
Could you also update the docs here to include a valid looking sas: https://github.com/stryker-mutator/stryker-net/blob/master/docs/configuration.md?plain=1#L555

src/Stryker.Core/Stryker.Core/Options/Inputs/AzureFileStorageSasInput.cs
```diff
@@ -18,8 +20,16 @@ public string Validate(BaselineProvider baselineProvider)
                     throw new InputException("The azure file storage shared access signature is required when azure file storage is used for dashboard compare.");
                 }
 
-                // Normalize the SAS
-                return SuppliedInput.Replace("?sv=", "");
```
rouke-broersma, line 22: This technically makes the change backwards incompatible, we will have to discuss if we want to make a breaking change at this point. I will get back to you on that.

src/Stryker.Core/Stryker.Core/Baseline/Providers/AzureFileShareBaselineProvider.cs
```diff
@@ -53,15 +56,16 @@ public async Task<JsonReport> Load(string version)
 
         public async Task Save(JsonReport report, string version)
         {
-            var fileUrl = $"{_options.AzureFileStorageUrl}/{_outputPath}/{version}/stryker-report.json";
-            var url = new Uri($"{fileUrl}?comp=range&sv={_options.AzureFileStorageSas}");
```
rouke-broersma, line 57: Since we would no longer inject sv= in front of the key this makes the change technically backwards incompatible, we will have to discuss if we want to make a breaking change at this point. I will get back to you on that. We don't have any analytics on if users would typically include the ?sv= in the key or not. If they don't (as our example does not), we would break everyone's setup.

tidusjar: Yeah that’s a good point. 

From my point of view, you cannot use the later versions anyway due to the SAS has changed. 
you cannot use the old token generation either as in a previous PR we have removed the required headers. 

I’ll update the docs as part of this PR

PR #2151, opened 2022-08-29
feat(initial test): Configure stryker to fail when initial testrun fails

BlackGad:
Adds ability to configure behavior on initial test run failure


BlackGad:
I definitely missed something. Also naming is under the question. :) Please point me what else I need to do here.

src/Stryker.Core/Stryker.Core/Options/TestFailureBreakBehavior.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Stryker.Core.Options
+{
+    public enum TestFailureBreakBehavior
```
rouke-broersma, line 3: I don't think we need this many options. Either the user does not want any failing unit tests, or stryker decides what is an acceptable level of test failure where we can still have an accurate result for the non-failing tests.

BlackGad: So do I need to change this enum to simple bool?

rouke-broersma: I would prefer that yes

src/Stryker.CLI/Stryker.CLI/CommandLineConfig/CommandLineConfigReader.cs
```diff
@@ -136,8 +136,8 @@ private void PrepareCliOptions(IStrykerInputs inputs)
             AddCliInput(inputs.DashboardApiKeyInput, "dashboard-api-key", null, category: InputCategory.Reporting);
             AddCliInput(inputs.AzureFileStorageSasInput, "azure-fileshare-sas", null, category: InputCategory.Reporting);
 
+            AddCliInput(inputs.TestsFailureBreakInput, "tests-failure-break", null, optionType: CommandOptionType.SingleOrNoValue, category: InputCategory.Misc);
```
rouke-broersma, line 139: What do you think of this for the name

```suggestion
            AddCliInput(inputs.BreakOnInitialTestFailure, "break-on-intitial-test-failure", null, optionType: CommandOptionType.SingleOrNoValue, category: InputCategory.Misc);
```

src/Stryker.CLI/Stryker.CLI/FileBasedInput.cs
```diff
@@ -90,6 +90,9 @@ public class FileBasedInput : IExtraData
 
         [JsonExtensionData]
         public Dictionary<string, JsonElement> ExtraData { get; init; }
+
+        [JsonPropertyName("tests-failure-break")]
+        public bool? TestsFailureBreak { get; init; }
```
rouke-broersma, line 95: Is there a good usecase for having the config option in the config file?

rouke-broersma:
Once we have chosen the final name for the new option, can you also update the configuration documentation?

src/Stryker.CLI/Stryker.CLI/FileBasedInput.cs
BlackGad: We relay on config files in our pipelines, so it is crucial for me :)

src/Stryker.CLI/Stryker.CLI/CommandLineConfig/CommandLineConfigReader.cs
BlackGad: Fixed on new commits

src/Stryker.CLI/Stryker.CLI/FileBasedInput.cs
rouke-broersma: Our philosophy is this:

Config file options are those options that rarely change for a project. They are the defaults necessary for correctly running stryker on the project within the context of the team using it. Commandline options are those options that are used most often in one-off scenarios. Configuration you need to overwrite from it's default depending on the current use case (CI vs developer machine for example). To limit configuration overload (especially on the commandline) we prefer to only enable configuration options where they fit with the above definitions.

With this explanation do you believe that the option is needed both in the configfile and on the commandline? I can see a use for having it both in the configfile and on the commandline but if you don't then I would prefer to only add it where you would need it. We can always add it later.

BlackGad: But it is up to you to decide ofc

BlackGad: Ok then lets leave only CLI option

BlackGad:
Renamed and option removed from config file. Only CLI left

BlackGad:
> Once we have chosen the final name for the new option, can you also update the configuration documentation?

I dont know where documentation is located :)

rouke-broersma:
https://github.com/stryker-mutator/stryker-net/blob/master/docs/configuration.md

BlackGad:
Will this PR be merged?

rouke-broersma:
When I have time.

richardwerkman:
Could you add tests to StrykerCLITests for this new option? To prove the option is actually processed when passed through the CLI

BlackGad:
Thank you!

psfinaki:
@rouke-broersma I think we have a usecase to have this in a config-file. We currently have about 200 projects and run mutation testing for them with the same stryker-config.json - and we want to have the same policy on this point (breaking on the initial test run).

Can I add this is a config-file option then in a separate PR?

rouke-broersma:
@psfinaki is it necessary to have this in the config file? I would assume you also run the unit tests separately, and that your pipelines are already highly templated and can contain the flag in the command line argument.

psfinaki:
Well rather it is as necessary there as other config options like reporters or concurrency. We could have them all in the command line but it feels cleaner to have them in the config file to shift there all this "responsibility" about invariant Stryker options. So that is just following the philosophy you mentioned in one of the discussions in the PR.

rouke-broersma:
@psfinaki the reason concurrency is in the config file is so you can disable concurrency if your project has shared state during unit tests so that would be static config. Reporters perhaps shouldn't actually be in the config file, good point.

But I suppose having one config file for ci and one for manual running can also make sense. Sure create the pr.

psfinaki:
@rouke-broersma alrighty, there you go: https://github.com/stryker-mutator/stryker-net/pull/2390

PR #2155, opened 2022-08-31
chore: Refactor dependency injection

dupdob:
Refactor process classes to simplify design. Get rid of some intermediate factory classes.
Reduce coupling with C# and F# coupling, giving the opportunity to move the incomplete F# support to a dedicated project, or at least branch

### DIFF
--- a/src/Stryker.Core/Stryker.Core/MutantFilters/FilePatternMutantFilter.cs
+++ b/src/Stryker.Core/Stryker.Core/MutantFilters/FilePatternMutantFilter.cs
@@ -35,15 +35,15 @@ namespace Stryker.Core.MutantFilters
                 }
 
                 // Check if the mutant is excluded.
-                if (_excludePattern.Any(MatchesPattern))
-                {
-                    return false;
-                }
-
-                return true;
+                return !_excludePattern.Any(MatchesPattern);
 
                 bool MatchesPattern(FilePattern pattern)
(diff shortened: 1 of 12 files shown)


src/Stryker.Core/Stryker.Core/MutantFilters/FilePatternMutantFilter.cs
```diff
@@ -35,15 +35,15 @@ bool IsMutantIncluded(Mutant mutant)
...
-                    return false;
-                }
-
-                return true;
+                return !_excludePattern.Any(MatchesPattern);
 
                 bool MatchesPattern(FilePattern pattern)
                 {
+                    // if we do not have the original node, we cannot exclude the mutation according to its location
+                    if (mutant.Mutation.OriginalNode == null)
```
rouke-broersma, line 43: In which case would we not have the original node at this point?

dupdob: during unit tests. Curent version explicitly disable filters to avoid this issue. Requiring to have a specific dependency injection just to disable filters for test, which does not seem clean to me.

Better alternative would be to ensure Mutant for test do have an original node.
Alternatively, we could remove the default value for FilePatternMatcher (**.* I think), since it is a systematic passthrough

Issue #1194, opened 2020-09-08
Mutators for Math


Myoxocephalus:
I am thinking about adding mutations for System Math.
For exemple i would like to add mutations like:
| Before | After |
|--------|-------|
| Min    | Max   |
| Max    | Min   |

If you think it´s ok i am happy to send a PR.

Mobrockers:
Hi! 

I would appreciate an as complete proposal as possible. Could you list all the possible and useful mutations you would like to add in the table so we can judge it as a whole?

I would rather not piecemeal together a mutator as it will be hard to keep track of progress on the completeness of the mutator if we don't at least try to identify the possibilities from the start.

That said, I think this could be useful for sure :)

Myoxocephalus:
Yes good comment, i could have provided some more information 😃 
Some more mutations could be:
| Before | After |
|--------|-------|
| Min    | Max   |
| Max    | Min   |
| Ceiling    | Floor   |
| Floor    | Ceiling   |
| MinMagnitude    | MaxMagnitude   |
| MaxMagnitude    | MinMagnitude   |
| Cos, Sin and Tan    | Sin, Cos or Tan   |
| BitDecrement    | BitIncrement   |
| BitIncrement    | BitDecrement   |
| Log    | Exp   |
| Exp    | Log   |

richardwerkman:
We could also mutate other kinds like:
```cs
Math.Abs(10);
// to
-Math.Abs(10);
```
or
```cs
Math.Sqrt(10);
// to
10 * 10;
```

But this would make the mutator a lot more difficult. So I'd say we just start with the suggested swapping of method names :+1:

richardwerkman:
About the Cos, Sin and Tan mutations. I think we have a risk they are equivalent mutations. For example a rectangle with 2 exactly the same angles. If we swap the Cos and Tan with equal angles, the outcome will be the same too right? (I'm not a math expert so this might be a wrong assumption.) If the outcome is the same we have created an unkillable mutant.

richardwerkman:
I've been thinking some more about the Cos, Sin and Tan mutation. It could indeed be possible that the outcome of the methods it exactly the same. But only in the case of a perfect equal triangle. So if that happens to someone they should change their tests to use some other values. 

So I agree with the proposed mutations. @Myoxocephalus will you start working on the implementation?

Myoxocephalus:
Yes, i expect to show some WIP PR in 2-3 weeks.

[...]

PR #2244, opened 2022-10-14
feat(mutators): Add Math methods mutator

yakivyusin:
Mutator from #1194 .
In addition to the proposed swaps from the issue, the implementation covers hyperbolic, inverse hyperbolic, inverse trigonometric functions, and some functions from newer .NET releases.

### DIFF
--- a/docs/mutations.md
+++ b/docs/mutations.md
@@ -159,3 +159,56 @@ Do you have a suggestion for a (new) mutator? Feel free to create an [issue](htt
 
 ## Regular Expressions (_regex_)
 For the full list of all available regex mutations, see the [regex mutator docs](./regex-mutations.md).
+
+## Math Methods (_math_)
+|      Original           |         Mutated         |
+| ----------------------- | ----------------------- |
+| Acos()                  | Acosh()                 |
+| Acos()                  | Asin()                  |
+| Acos()                  | Atan()                  |
+| Acosh()                 | Acos()                  |
+| Acosh()                 | Asinh()                 |
+| Acosh()                 | Atanh()                 |
+| Asin()                  | Asinh()                 |
+| Asin()                  | Acos()                  |
+| Asin()                  | Atan()                  |
+| Asinh()                 | Asin()                  |
+| Asinh()                 | Acosh()                 |
+| Asinh()                 | Atanh()                 |
+| Atan()                  | Atanh()                 |
+| Atan()                  | Acos()                  |
+| Atan()                  | Asin()                  |
+| Atanh()                 | Atan()                  |
+| Atanh()                 | Acosh()                 |
+| Atanh()                 | Asinh()                 |
+| BitDecrement()          | BitIncrement()          |
+| BitIncrement()          | BitDecrement()          |
+| Ceiling()               | Floor()                 |
+| Cos()                   | Cosh()                  |
+| Cos()                   | Sin()                   |
+| Cos()                   | Tan()                   |
+| Cosh()                  | Cos()                   |
+| Cosh()                  | Sinh()                  |
+| Cosh()                  | Tanh()                  |
+| Exp()                   | Log()                   |
+| Floor()                 | Ceiling()               |
+| Log()                   | Exp()                   |
+| Log()                   | Pow()                   |
+| MaxMagnitude()          | MinMagnitude()          |
+| MinMagnitude()          | MaxMagnitude()          |
+| Pow()                   | Log()                   |
+| ReciprocalEstimate()    | ReciprocalSqrtEstimate() |
+| ReciprocalSqrtEstimate() | ReciprocalEstimate()   |
+| ReciprocalSqrtEstimate() | Sqrt()                 |
+| Sin()                   | Sinh()                  |
+| Sin()                   | Cos()                   |
+| Sin()                   | Tan()                   |
+| Sinh()                  | Sin()                   |
+| Sinh()                  | Cosh()                  |
+| Sinh()                  | Tanh()                  |
+| Tan()                   | Tanh()                  |
+| Tan()                   | Cos()                   |
+| Tan()                   | Sin()                   |
+| Tanh()                  | Tan()                   |
+| Tanh()                  | Cosh()                  |
+| Tanh()                  | Sinh()                  |
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutators/MathMutatorTest.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutators/MathMutatorTest.cs
@@ -1,0 +1,146 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Shouldly;
+using Stryker.Core.Mutators;
+using Xunit;
+
+namespace Stryker.Core.UnitTest.Mutators
+{
+    public class MathMutatorTest : TestBase
+    {
+        /// <summary>
+        ///     Generator for different Math expressions
+        /// </summary>
+        /// <param name="expression"></param>
+        /// <returns></returns>
+        private InvocationExpressionSyntax GenerateExpression(string expression)
+        {
+            SyntaxTree tree = CSharpSyntaxTree.ParseText($@"
+using System;
+
+namespace TestApplication
+{{
+    class Program
+    {{
+        static void Main(string[] args)
+        {{
+            Math.{expression}(5.0);
+        }}
+    }}
+}}");
+            var memberAccessExpression = tree.GetRoot()
+                .DescendantNodes()
+                .OfType<InvocationExpressionSyntax>()
+                .Single();
+
+            return memberAccessExpression;
+        }
+
+        [Fact]
+        public void ShouldBeMutationLevelAdvanced()
+        {
+            var target = new MathMutator();
+            target.MutationLevel.ShouldBe(MutationLevel.Advanced);
+        }
+
+        /// <summary>
+        ///     Test method to check for correct mutation of different Math Expression Mutations
+        /// </summary>
+        [Theory]
+        [MemberData(nameof(GetTrigonometricHyperbolicTestData))]
+        [InlineData(MathExpression.BitDecrement, new[] { MathExpression.BitIncrement })]
+        [InlineData(MathExpression.BitIncrement, new[] { MathExpression.BitDecrement })]
+        [InlineData(MathExpression.Ceiling, new[] { MathExpression.Floor })]
+        [InlineData(MathExpression.Exp, new[] { MathExpression.Log })]
+        [InlineData(MathExpression.Floor, new[] { MathExpression.Ceiling })]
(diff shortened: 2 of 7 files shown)


rouke-broersma:
Cool! Thank you :)

rouke-broersma:
@richardwerkman since you seem to have more affinity with math, could you review this new mutator?

docs/mutations.md
```diff
@@ -159,3 +159,56 @@ Do you have a suggestion for a (new) mutator? Feel free to create an [issue](htt
 
 ## Regular Expressions (_regex_)
 For the full list of all available regex mutations, see the [regex mutator docs](./regex-mutations.md).
+
+## Math Methods (_math_)
+|      Original           |         Mutated         |
+| ----------------------- | ----------------------- |
+| Acos()                  | Acosh()                 |
+| Acos()                  | Asin()                  |
+| Acos()                  | Atan()                  |
```
psfinaki, line 168: Is there a specific reason to have multiple mutated variants for the same thing? 

I vaguely remember from my trigonometry lessons at school 15 years ago that there are some cases when some of the pairs will have the same values (like errrrhhhhh sin and cos for 45 degrees?) but I think philosophically Stryker in the vast majority of cases puts up with that (@rouke-broersma correct me if I am wrong).

Like, Stryker changes + to - and other way round, but it doesn't also change + to * or /, even though `x + 0` and `x - 0` will give the same result. AFAIU the framework kind of tries to find a balance between the amount of value and amount of generated mutants (because the amount eventually affects the performance).

TL;DR - I'd stick with one mutant for each expression.

src/Stryker.Core/Stryker.Core/Mutators/MathMutator.cs
```diff
@@ -0,0 +1,146 @@
...
+            while (node is ConditionalAccessExpressionSyntax conditional)
+            {
+                foreach (var subMutants in FindMutableMethodCalls(conditional.Expression, original))
+                {
+                    yield return subMutants;
+                }
+                node = conditional.WhenNotNull;
+            }
+
+            for (; ; )
```
psfinaki, line 71: First, that's an obscure way to create an infinite loop from the times of C, I'd prefer any clearer form, e.g. `while (true)`.

Second, I haven't created mutators myself but I don't think I've seen infinite loops in the mutators' code (@dupdob correct me if I am wrong) and so I'm just generally paranoid about them. Can you somehow verbally describe this need here? Or maybe extract a method and give it a very expressive name :)

psfinaki (commented):
Got a bit of time, looked through the code :) thanks - it was a nice journey into System.Math, which I have never had an opportunity for in my career heh.

docs/mutations.md
yakivyusin: Hello.

The case with `sin(x) = cos(x)` was discussed in the initial issue - they are indeed equal for 45 and 225 degrees. From my experience with mutation testing, if someone gets such survived mutant then this is a good point to improve tests - just add a new test case with an another angle.

The same for other swaps, they will test well other "magic" angles that can be used in incomplete tests - e.g. `sin(x) = sinh(x)` for 0 and `sin(x) = tan(x)` for `pi*n`.

And swap trigonometric <-> hyperbolic function looks like a real possible error with the function selection in IntelliSense :D

psfinaki: Alright. I mean, I haven't worked in codebases where this stuff was heavily used, just wanted to share the concern that it can generate many mutants. But yeah, that's an advanced mutator anyway and will probably not be applicable for most repos.

src/Stryker.Core/Stryker.Core/Mutators/MathMutator.cs
psfinaki: Hmm, alright, good to know! But my feedback (and I am just a random dev) is the same - the code could be clearer here, we don't need to be uniform in crypticity ;)

yakivyusin: Yes, sure, I will think about how to refactor this place. Then, maybe, the good idea will be to refactor the same place in `LinqMutator` in the same way.

dupdob: I understand why you picked `LinqMutator` as a template, but it is needlessly complex for Math methods: as those are static methods, you do not need to worry about conditional accessor or chained invocation. You also don't need to worry about MemberBinding expression.
So you need to look for invocation looking like `Math.[Sin|Cos|Tan|...]`, which are `InvocationExpression` using  a `SimpleMemberAcessExpression` as its Expression. This one has an `IdentifierName` (token 'Math') as an expression and another `IndentifierName`(identifier 'Sin', 'Cos', ...).

Therefore you can drop the initial `while` loop and transform the switch case to a simpler `if`.

You can use sharplab.io to have an idea of how different syntax constructs are represented with Roslyn.
Here is a [simple example](https://sharplab.io/#v2:C4LghgzsA0AmIGoA+ABATARgLACgUGYACdQgYUIG9dCbiiUAWQgWQAoBKS62n5sYABYA6AMoBLAHasMABnYBubjQC+uZUA==).
I think you will find this useful.

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/MathMutatorTest.cs
```diff
@@ -0,0 +1,146 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Shouldly;
+using Stryker.Core.Mutators;
+using Xunit;
+
+namespace Stryker.Core.UnitTest.Mutators
+{
+    public class MathMutatorTest : TestBase
```
richardwerkman, line 13: I would like to see a test to see if these methods are *not* mutated when the class has a different name. I can imagine we will see a class with a method Ceiling and we could mutate that as well (which will fail)

src/Stryker.Core/Stryker.Core/Mutators/MathMutator.cs
```diff
@@ -0,0 +1,146 @@
...
+            [MathExpression.ReciprocalEstimate] = new[] { MathExpression.ReciprocalSqrtEstimate },
+            [MathExpression.ReciprocalSqrtEstimate] = new[] { MathExpression.ReciprocalEstimate, MathExpression.Sqrt },
+            [MathExpression.Sin] = new[] { MathExpression.Sinh, MathExpression.Cos, MathExpression.Tan },
+            [MathExpression.Sinh] = new[] { MathExpression.Sin, MathExpression.Cosh, MathExpression.Tanh },
+            [MathExpression.Tan] = new[] { MathExpression.Tanh, MathExpression.Cos, MathExpression.Sin },
+            [MathExpression.Tanh] = new[] { MathExpression.Tan, MathExpression.Cosh, MathExpression.Sinh }
+        };
+
+        /// <summary> Apply mutations to an <see cref="InvocationExpressionSyntax"/> </summary>
+        public override IEnumerable<Mutation> ApplyMutations(ExpressionSyntax node)
```
richardwerkman, line 46: Can't we target a more specific Syntax type? This mutator will now be called for almost every type of node...

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/MathMutatorTest.cs
yakivyusin: What about `using static System.Math` or `using AnyName = System.Math`?

src/Stryker.Core/Stryker.Core/Mutators/MathMutator.cs
yakivyusin: Thank you. I will change the implementation to use this way.

src/Stryker.Core/Stryker.Core/Mutators/MathMutator.cs
yakivyusin: Thank you, I will limit method to `InvocationExpressionSyntax` during refactoring.

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/MathMutatorTest.cs
dupdob: I guess you should handle `using static System.Math` which implies supporting the short call form `Cos(x)` : easy to do and with limited compilation errors risk (those should only happens WITHIN a class providing methods with mathematical names).

I think the alias form is so rare there is no need to support it. Remember: mutation testing is about stress testing your tests, this is not some form of extensive testing.

It should be easy to properly support aliasing when the decision is made to generate semantic information alongside the syntax tree (if ever).

psfinaki: > I think the alias form is so rare there is no need to support it.

Murphy's law says that whatever can happen will happen :D

src/Stryker.Core/Stryker.Core/Mutators/MathMutator.cs
```diff
@@ -0,0 +1,125 @@
...
+        {
+            MemberAccessExpressionSyntax memberAccess => ApplyMutationsToMemberCall(node, memberAccess),
+            IdentifierNameSyntax methodName => ApplyMutationsToDirectCall(node, methodName),
+            _ => Enumerable.Empty<Mutation>()
+        };
+
+        private static IEnumerable<Mutation> ApplyMutationsToMemberCall(InvocationExpressionSyntax node, MemberAccessExpressionSyntax memberAccessExpressionSyntax)
+        {
+            if (memberAccessExpressionSyntax.Expression is not IdentifierNameSyntax memberName ||
+                !memberName.Identifier.ValueText.StartsWith("Math"))
```
psfinaki, line 57: Maybe at least it's possible to verify the "System" namespace here for it?

psfinaki (approved):

src/Stryker.Core/Stryker.Core/Mutators/MathMutator.cs
psfinaki: Anyway, code is much clearer now, good job!

src/Stryker.Core/Stryker.Core/Mutators/MathMutator.cs
yakivyusin: From my point of view, it looks like we can't check the fully qualified name without the semantic model. But maybe I'm wrong.

dupdob: yes and no; you can check the given name, so you can verify that the name is either `Math.Con` or `Con` (no namespace provided) not 'Something.Con` nor `Something.Math.Con`. Less precise than using the semantic model, but good enough for most cases

yakivyusin: @dupdob But the existing code already works like that, doesn't it? The code exactly checks what you described.

psfinaki: Alrighty then, you can resolve this.

yakivyusin:
@richardwerkman @dupdob Hello, could you please review the new implementation to resolve the old comments & accept the PR?

dupdob (approved):
Comments have been made and remarks integrated.
Thanks for this good PR PR.

PR #2259, opened 2022-10-28
feat(InitializerMutator): Add object initializer mutation

### LINKED ISSUE #1025: Feature request: Mutate object initializer
This issue comes from #1023. At this time we don't mutate object initializers like:

``` cs
new Dictionary<int, int>
{
    [0] = 1
};
```
or 

``` cs
var person = new Person()
{
    Age = 1
};
```

We could mutate them into an empty form:

``` cs
new Dictionary<int, int>{};
```
or 

``` cs
var person = new Person(){};
```

For the Dictionary this makes a lot of sense as we already mutate all collection types. For regular objects I'm not sure yet.

Liam-Rougoor:
This PR adds mutation of object initializers.

Closes #1025


src/Stryker.Core/Stryker.Core/Mutators/Mutator.cs
```diff
@@ -33,7 +33,7 @@ public enum Mutator
         String,
         [Description("Bitwise operators")]
         Bitwise,
-        [Description("Array initializer")]
+        [Description("Initializer")]
```
rouke-broersma, line 36: This creates a breaking change. To avoid that we might support multiple descriptions for a mutator. We can then deprecate the old description and remove it in the next breaking change.

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/InitializerMutatorTests.cs
```diff
@@ -34,6 +34,28 @@ public void ShouldRemoveValuesFromArrayInitializer()
             replacement.Expressions.ShouldBeEmpty();
         }
 
+        [Fact]
+        public void ShouldRemoveValuesFromObjectInitializer()
```
richardwerkman, line 38: could you add a test for dictionary initializer?

src/Stryker.Core/Stryker.Core/Mutators/Mutator.cs
richardwerkman, line 36: This is a breaking change... We need to find a way to keep this backwards compatible

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/InitializerMutatorTests.cs
richardwerkman: Also: I prefer to use `SyntaxFactory.ParseExpression("code here")` in unit tests as it is more readable than using the SyntaxFactory to create your nodes.

richardwerkman:
Could you do a manual test to see if this doesn't result in compile errors in cases where a property is not initialized before it's used (thanks to this mutation)

richardwerkman:
Don't forget to update the documentation with this new mutation

src/Stryker.Core/Stryker.Core/Mutators/Mutator.cs
richardwerkman: Revert this back to Array initializer

Liam-Rougoor: Reverted 👍

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/InitializerMutatorTests.cs
Liam-Rougoor: The dictionary test apparently already existed in the existing mutator.
Switched to using ParseExpression 👍

src/Stryker.Core/Stryker.Core/Mutators/ObjectCreationMutator.cs
```diff
@@ -22,6 +22,16 @@ public override IEnumerable<Mutation> ApplyMutations(ObjectCreationExpressionSyn
                     Type = Mutator.Initializer
                 };
             }
+            if (node.Initializer?.Kind() == SyntaxKind.ObjectInitializerExpression && node.Initializer.Expressions.Count > 0)
```
rouke-broersma, line 25: Can we not merge all this with the InitializerMutator?

src/Stryker.Core/Stryker.Core/Mutators/InitializerMutator.cs
```diff
@@ -11,11 +11,12 @@ public class InitializerMutator : MutatorBase<InitializerExpressionSyntax>, IMut
 
         public override IEnumerable<Mutation> ApplyMutations(InitializerExpressionSyntax node)
         {
-            if (node.Parent is ArrayCreationExpressionSyntax || node.Parent is ImplicitArrayCreationExpressionSyntax || node.Parent is StackAllocArrayCreationExpressionSyntax)
+            if (node.Parent is ArrayCreationExpressionSyntax || node.Parent is ImplicitArrayCreationExpressionSyntax || node.Parent is StackAllocArrayCreationExpressionSyntax || node.Expressions.Count <= 0)
```
psfinaki, line 14: Sorry, I'm late to the party (as usual), was just looking through the PRs now... Just to double-check, is this correct? Feels like >= should be here :) 

@rouke-broersma

rouke-broersma: Should probably be `==`. The `>` case is handled in the ObjectCreationMutator.

@Liam-Rougoor typo?

Liam-Rougoor: @psfinaki This is a check whether the node doesn't have any expressions. `==` would indeed have sufficed.

psfinaki: Alright. I noticed it - I'll fix it :D

PR #2267, opened 2022-10-28
feat(reporting): Specify output path for reports and logs

### LINKED ISSUE #1078: Output folder CLI argument
Currently, the default folder for the mutation report is called _StrykerOutput_. I can't find any option or setting to change this folder name or path.

It would be nice to have an option to specify the folder name as one of the CLI arguments. Or, if this is already available, then could you please point out how to use it?

One of the blocked scenarios is the GitlabCI pipeline. We are running several steps of mutation testing via Stryker and collecting reports via artifacts, but because it writes reports into a hardcoded directory only one step ends up being with artifacts in that folder. In order to overcome this, we'd have to rename the output folder after the mutator run.

Thanks!

ysbakker:
Added a new CLI input option (-O or --output) to specify the output path for reports and logs.

Closes #1078


src/Stryker.CLI/Stryker.CLI/Logging/LoggingInitializer.cs
```diff
@@ -31,16 +31,25 @@ public void SetupLogOptions(IStrykerInputs inputs, IFileSystem fileSystem = null
...
+            string outputPath;
+
+            if (Path.IsPathRooted(basePath))
+            {
+                outputPath = Path.Combine(basePath, DateTime.Now.ToString("yyyy-MM-dd.HH-mm-ss"));
+            }
+            else
+            {
+                outputPath = Path.Combine(inputs.BasePathInput.SuppliedInput, basePath, DateTime.Now.ToString("yyyy-MM-dd.HH-mm-ss"));
+            }
```
rouke-broersma, line 46: If the path is supplied by the user, we do not generate the name of the folder using the timestamp. We use the exact path as supplied by the user.

src/Stryker.Core/Stryker.Core/Options/Inputs/OutputPathInput.cs
```diff
@@ -5,7 +5,7 @@ namespace Stryker.Core.Options.Inputs
 {
     public class OutputPathInput : Input<string>
     {
-        protected override string Description => string.Empty;
+        protected override string Description => @"Changes the output directory for Stryker reports and logs.";
```
rouke-broersma, line 8:
```suggestion
        protected override string Description => "Changes the output directory for Stryker reports and logs.";
```

Does not need to be a string literal

src/Stryker.CLI/Stryker.CLI/CommandLineConfig/CommandLineConfigReader.cs
```diff
@@ -135,9 +135,11 @@ private void PrepareCliOptions(IStrykerInputs inputs)
             AddCliInput(inputs.ProjectVersionInput, "version", "v", category: InputCategory.Reporting);
             AddCliInput(inputs.DashboardApiKeyInput, "dashboard-api-key", null, category: InputCategory.Reporting);
             AddCliInput(inputs.AzureFileStorageSasInput, "azure-fileshare-sas", null, category: InputCategory.Reporting);
+            AddCliInput(inputs.OutputPathInput, "output", "O", optionType: CommandOptionType.SingleValue, category: InputCategory.Reporting);
```
rouke-broersma, line 138: Should also be added to the configuration docs

src/Stryker.Core/Stryker.Core.UnitTest/Options/Inputs/OutputPathInputTests.cs
```diff
@@ -14,7 +14,7 @@ public class OutputPathInputTests : TestBase
         public void ShouldHaveHelpText()
         {
             var target = new OutputPathInput();
-            target.HelpText.ShouldBe(@"");
+            target.HelpText.ShouldBe(@"Changes the output path for Stryker logs and reports. This can be an absolute or relative path.");
```
rouke-broersma, line 17: Still not a string literal

PR #2274, opened 2022-10-28
feat(assignment mutator): Mutate coalesce assignment operations

PBonnema:
Added a mutator for the `??=` assignment


dupdob:
you need to update the expected results for integration tests (after manually confirming you are ok with them.

PR #2276, opened 2022-10-29
fix(assignment statement mutator): Add trivia between operands

PBonnema:
Fixed that the output of the AssignmentMutator is missing a space after the operator.

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Mutators/AssignmentStatementMutator.cs
+++ b/src/Stryker.Core/Stryker.Core/Mutators/AssignmentStatementMutator.cs
@@ -41,10 +41,12 @@ namespace Stryker.Core.Mutators
 
             foreach (var targetAssignmentKind in targetAssignmentKinds)
             {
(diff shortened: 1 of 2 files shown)


src/Stryker.Core/Stryker.Core/Mutators/AssignmentStatementMutator.cs
```diff
@@ -41,10 +41,12 @@ public override IEnumerable<Mutation> ApplyMutations(AssignmentExpressionSyntax
 
             foreach (var targetAssignmentKind in targetAssignmentKinds)
             {
+                var replacementNode = SyntaxFactory.AssignmentExpression(targetAssignmentKind, node.Left, node.Right);
+                replacementNode = replacementNode.WithOperatorToken(replacementNode.OperatorToken.WithTriviaFrom(node.OperatorToken));
```
richardwerkman, line 45: Can you add a test for this?

PR #2277, opened 2022-10-29
fix(mutator description): multiple mutator descriptions

### LINKED ISSUE #2261: Support multiple mutator types
The [mutator types](https://github.com/stryker-mutator/stryker-net/blob/4efadc2c5874efb2f5f712e777b825e47ef26da5/src/Stryker.Core/Stryker.Core/Mutators/Mutator.cs) support only one description at the moment. If we want to rename a type we have to deprecate the old name and introduce a new name. This requires support for multiple description attributes.

I'd like to be able to set two or more descriptions for each mutator type like:
```cs
        [Description("Arithmetic operators")]
        [Description("Math operators")]
        Arithmetic,
```

Both names should work to exclude them in the options.

We don't want to mark descriptions as deprecated but will announce this using the changelog and blogposts.

Liam-Rougoor:
This PR adds support for multiple [Description]s on a Mutator. IgnoreMutationsInputs ignores the mutator based on any of the descriptions.

Closes #2261


src/Stryker.Core/Stryker.Core/Attributes/DescriptionAttribute.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Stryker.Core.Attributes
+{
+    [AttributeUsage(AttributeTargets.All, AllowMultiple = true)]
+    public class DescriptionAttribute : Attribute
```
rouke-broersma, line 7: I would like to make this more specific. Something like MutatorDescription

PR #2285, opened 2022-10-29
fix(PrefixUnaryExpression mutator): Handle mutations of prefixed increment/decrement

### LINKED ISSUE #2272: Mutants on some pre-increment/decrement operators don't compile
**Describe the bug**
Mutants for pre-increment/decrement operators at the start of a statement don't compile:
image

**Logs**
[logs.txt](https://github.com/stryker-mutator/stryker-net/files/9890932/logs.txt)

**Expected behavior**

 - OS: Windows
 - Type of project core
 - Framework Version 6.0
 - Stryker Version 3.0.1

**Additional context**
@rouke-broersma  believes this might have to do with the fact these statements are always placed with the tertiary operator instead of using the ifs even though this doesn't work in the above case.

dupdob:
Fixes #2272


dupdob:
For info, I will not address Sonar warning. This is a false positive.

PR #2289, opened 2022-10-29
fix(reporting): Correctly show whitespace in report

### LINKED ISSUE #2270: Incorrect whitespace in output of the assignment mutator
**Describe the bug**
image

As shown, when having an assignment as the first/only thing in the statement (without anything in front of it except for indentation), you get too much whitespace in front of the resulting mutant. I also verified this with other statements that are also expressions like post-increments/decrements (e.g. `a++;`):
image

**Logs**
For the `a++;` case:
[logs.txt](https://github.com/stryker-mutator/stryker-net/files/9890810/logs.txt)

**Expected behavior**
I expected there not to be so much whitespace.

 - OS: Windows
 - Type of project: core
 - Framework Version 6.0
 - Stryker Version 3.0.1

**Additional context**

ysbakker:
Now using `SyntaxNode.ToString()` instead of `SyntaxNode.ToFullString()` in the `JsonMutant` class. [`SyntaxNode.ToFullString()`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.syntaxnode.tofullstring?view=roslyn-dotnet-4.3.0) also includes leading and trailing trivia. Leading trivia should not be necessary as the location is already set here:

https://github.com/stryker-mutator/stryker-net/blob/085a0b7837370b709b00285ba4efe5ad55718a0a/src/Stryker.Core/Stryker.Core/Reporters/Json/SourceFiles/JsonMutant.cs#L35

This change renders the report correctly: 
image

Q: Should trailing trivia be included?

Fixes #2270

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Reporters/Json/SourceFiles/JsonMutant.cs
+++ b/src/Stryker.Core/Stryker.Core/Reporters/Json/SourceFiles/JsonMutant.cs
@@ -31,7 +31,7 @@ namespace Stryker.Core.Reporters.Json.SourceFiles
             MutatorName = mutant.Mutation.DisplayName;
             Description = mutant.Mutation.Description;
 
-            Replacement = mutant.Mutation.ReplacementNode.ToFullString();
+            Replacement = mutant.Mutation.ReplacementNode.ToString();
             Location = new Location(mutant.Mutation.OriginalNode.GetLocation().GetMappedLineSpan());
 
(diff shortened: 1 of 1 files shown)


rouke-broersma:
Can you test the change on multiple mutants on the same line to check if trailing trivia is needed?

ysbakker:
Like this?

image

image

image

image

Even comments seem to work (?):
image

Any other cases I should test?

PR #2291, opened 2022-10-30
docs(operating modes): Add operating modes page

### LINKED ISSUE #2281: Document all possible ways to run stryker
We're missing documentation on how to run stryker.

We should add 3 chapters explaining how to:
- Run from testproject context
- Run from targetproject context (with --test-projects)
- Run from solution context

ysbakker:
Added a new page to the documentation to specify the different ways to run Stryker. I did not document the solution context way, as it's still in progress (right?).

Closes #2281

### DIFF
--- a/docs/operating-modes.md
+++ b/docs/operating-modes.md
@@ -20,17 +20,11 @@ Just run
 dotnet stryker
 ```
 
-in your test project folder. Stryker will automatically detect the target project based on the project reference. If your test project has multiple project references, you need to specify the target project with `--project <filename>` ([docs here](https://stryker-mutator.io/docs/stryker-net/configuration/#project-file-name)). You can only specify one target project at the time.
+in your test project folder. Stryker will automatically detect the source project based on the project reference. If your test project has multiple project references, you need to specify the source project with `--project <filename>` ([docs here](https://stryker-mutator.io/docs/stryker-net/configuration/#project-file-name)). You can only specify one source project at the time.
 
-## Target project context
+## Source project context
 
-Run
-
-```bash
-dotnet stryker -tp "../Tests/Tests.csproj"
-```
-
-in your target project folder. If multiple test projects (ex unit tests, specs) target the same source project you need to specify each test project. You can include multiple test projects like this ([see docs](https://stryker-mutator.io/docs/stryker-net/configuration/#test-projects-string)):
+Run `dotnet stryker -tp "../Tests/Tests.csproj"` in your source project folder. If multiple test projects (ex unit tests, specs) target the same source project you need to specify each test project. You can include multiple test projects like this ([see docs](https://stryker-mutator.io/docs/stryker-net/configuration/#test-projects-string)):
 
 ```bash
 dotnet stryker -tp "../Tests/Tests.csproj" -tp "../MoreTests/MoreTests.csproj"
--- a/docs/operating-modes.md
+++ b/docs/operating-modes.md
@@ -14,13 +14,8 @@ You will also be able to run Stryker from your solution context in the future.
 
 ## Test project context
 
-Just run
-
-```bash
-dotnet stryker
-```
-
-in your test project folder. Stryker will automatically detect the source project based on the project reference. If your test project has multiple project references, you need to specify the source project with `--project <filename>` ([docs here](https://stryker-mutator.io/docs/stryker-net/configuration/#project-file-name)). You can only specify one source project at the time.
+Just run `dotnet stryker` in your test project folder.  
+Stryker will automatically detect the source project based on the project reference. If your test project has multiple project references, you need to specify the source project with `--project <filename>` ([docs here](https://stryker-mutator.io/docs/stryker-net/configuration/#project-file-name)). You can only specify one source project at the time.
 
 ## Source project context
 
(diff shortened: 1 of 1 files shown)


docs/operating-modes.md
```diff
@@ -0,0 +1,41 @@
...
+title: Operating modes
+sidebar_position: 25
+custom_edit_url: https://github.com/stryker-mutator/stryker-net/edit/master/docs/operating-modes.md
+---
+# Operating modes
+
+There are currently two different ways to run Stryker:
+
+- From the test project context
+- From the target project context
```
rouke-broersma, line 11:
```suggestion
- From the source project context
```

@richardwerkman what do we want to call this? We've called it target project for a long time but that really only makes sense from the context of the test project. It doesn't really make sense anymore now that we can run stryker from different contexts.

docs/operating-modes.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Target project context
+
+Run
+
+```bash
+dotnet stryker -tp "../Tests/Tests.csproj"
+```
+
+in your target project folder. If your target project uses multiple test projects that you want to include in the Stryker run, you need to specify each test project. You can include multiple test projects like this ([see docs](https://stryker-mutator.io/docs/stryker-net/configuration/#test-projects-string)):
```
rouke-broersma, line 33:
```suggestion
in your target project folder. If multiple test projects (ex unit tests, specs) target the same source project you need to specify each test project. You can include multiple test projects like this ([see docs](https://stryker-mutator.io/docs/stryker-net/configuration/#test-projects-string)):
```

docs/operating-modes.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Test project context
+
+Just run
+
+```bash
+dotnet stryker
+```
+
+in your test project folder. Stryker will automatically detect the target project based on the project reference. If your test project has multiple project references, you need to specify the target project with `--project <filename>` ([docs here](https://stryker-mutator.io/docs/stryker-net/configuration/#project-file-name)). You can only specify one target project at the time.
```
rouke-broersma, line 23:
```suggestion
Just run `dotnet stryker` in your test project folder.  
Stryker will automatically detect the target project based on the project reference. If your test project has multiple project references, you need to specify the target project with `--project <filename>` ([docs here](https://stryker-mutator.io/docs/stryker-net/configuration/#project-file-name)). You can only specify one target project at the time.
```

ysbakker: Hier ook source project noemen dan?

```suggestion
Just run `dotnet stryker` in your test project folder.  
Stryker will automatically detect the source project based on the project reference. If your test project has multiple project references, you need to specify the source project with `--project <filename>` ([docs here](https://stryker-mutator.io/docs/stryker-net/configuration/#project-file-name)). You can only specify one source project at the time.
```

docs/operating-modes.md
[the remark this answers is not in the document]
ysbakker: Run `dotnet stryker -tp "../Tests/Tests.csproj"` in your target project folder.  
If your source project uses multiple test projects that you want to include in the Stryker run, you need to specify each test project. You can include multiple test projects like this ([see docs](https://stryker-mutator.io/docs/stryker-net/configuration/#test-projects-string)):

docs/operating-modes.md
rouke-broersma: If @richardwerkman agrees with that naming

docs/operating-modes.md
richardwerkman: Source makes more sense now for this. And should we call to solution one "source solution context"? Once it's done of course

docs/operating-modes.md
```diff
@@ -0,0 +1,30 @@
...
+sidebar_position: 25
+custom_edit_url: https://github.com/stryker-mutator/stryker-net/edit/master/docs/operating-modes.md
+---
+# Operating modes
+
+There are currently two different ways to run Stryker:
+
+- From the test project context
+- From the source project context
+
```
psfinaki, line 12: Wait... you can run Stryker from wherever, if you specify both source and test projects, at least it works with full paths.

docs/operating-modes.md
```diff
@@ -0,0 +1,30 @@
+---
+title: Operating modes
+sidebar_position: 25
+custom_edit_url: https://github.com/stryker-mutator/stryker-net/edit/master/docs/operating-modes.md
+---
+# Operating modes
+
+There are currently two different ways to run Stryker:
```
psfinaki, line 8:
```suggestion
There are currently two ways to run Stryker:
```

docs/operating-modes.md
richardwerkman: Wait does that work?... I never tried it but I would expect stryker to fail if you run it from a location outside of your project

psfinaki: It absolutely does :)

psfinaki: Well it's been there for a while, we have 1.5.3 in our pipeline currently and we are running things like this. 
So I don't know why and since when it works but it does... might be worth checking that the new custom output option is okay with that and also adding some integration tests then.

ysbakker: I see, I don't know if it's a good idea to add this to the documentation as it's not really officially supported right now. It can be added later when this feature is properly tested.

PR #2295, opened 2022-11-01
chore: initializermutator expressions cannot be less than 0

psfinaki:
Based on [this](https://github.com/stryker-mutator/stryker-net/pull/2259#discussion_r1009704958) comment.


rouke-broersma:
Since we had not caught this mistake before, could you add a unit test for this case 😅

psfinaki:
Well it's not trivial to add a test for a readonly value which can never happen without some refactoring here :D so I hacked things a bit and made use of the fact that the ">=" case is already covered. If you don't buy this or don't like `Any` for performance reasons, we can think of another condition maybe.

psfinaki:
Soooo? Can we check it in like this?

PR #2313, opened 2022-11-11
feat(mutators): Mutator for relational and logical patterns in is\switch expressions

### LINKED ISSUE #1338: Mutators for C# 9 Pattern Matching
I'm only just learning about this now so this might not make sense. I've got code like this;

```C#
if (myInput is not null)
{
    doSomething();
}
```
The `is not` pattern combination is new to C# 9 and Stryker is currently unable to mutate it. It looks like there are other pattern operators such as `and` and/or `or` 😄 . It would be nice to have mutators for those as well.

Someone will need to take a look at the pattern matching specification of C# 9. The following might help.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/patterns3

### LINKED ISSUE #2311: Mutator for is and switch expressions (pattern matching)
Stryker has mutations for logical and equality operators, but not for relational and logical patterns in `is/switch` expressions.
It would be useful to add support for them (with the same 'Basic' mutation level as for operators), since they do the similar thing. These mutators can later be extended by supporting other patterns, e.g. list pattern from C# 11 (with the higher mutation level)

Two new mutators (one for `is` expression and one for `switch` expression) that will support mutation of relational and logical patterns in the same way as for operators:
| `and` | `or` |
| `or` | `and` |
| `>` | `<` |
| `>` | `>=` |
| `>=` | `<` |
| `>=` | `>` |
| `<` | `>` |
| `<` | `<=` |
| `<=` | `>` |
| `<=` | `<` |

**Additional context**
Sometimes, these mutators could produce mutations with compile errors (the compiler is too smart and checks for unreacheable arms in `switch`), but the percentage of them should not be large.
I have already made a working PoC, so it will not take too much time :)

yakivyusin:
Closes #2311 , closes #1338


docs/mutations.md
```diff
@@ -159,6 +159,20 @@ Do you have a suggestion for a (new) mutator? Feel free to create an [issue](htt
...
+| Original | Mutated |
+| ------------- | ------------- |
+| `>` | `<` |
+| `>` | `>=` |
+| `>=` | `<` |
+| `>=` | `>` |
+| `<` | `>` |
+| `<` | `<=` |
+| `<=` | `>` |
+| `<=` | `<` |
```
richardwerkman, line 172: I wouldn't say these are new mutations. These are just Equality Operator mutations but in a different context. I'd say we add the `and` and `or` mutations to the Equality Operators chapter. @rouke-broersma agree?

rouke-broersma: We should check with the other stryker teams if they already have similar mutations and in which category they place them

yakivyusin: @richardwerkman @rouke-broersma Imo, it's better to separate pattern mutations from simple operators, even if they look similar. If you want to implement mutations for other patterns in the future, you will need the separate doc chapter anyway.
But of course it's up to you guys.

psfinaki (dismissed):

docs/mutations.md
richardwerkman: I've checked with stryker4s as they use a lot of pattern matching. They place these type of mutation in equality operators. So I'd say we keep in line with them.

yakivyusin: @richardwerkman Ok, thank you! I'll update the PR correspondingly.

yakivyusin: Docs are updated

docs/mutations.md
```diff
@@ -30,6 +30,8 @@ Do you have a suggestion for a (new) mutator? Feel free to create an [issue](htt
 | `<=` | `<` |
 | `==` | `!=` |
 | `!=` | `==` |
+| `and` | `or` |
+| `or` | `and` |
```
rouke-broersma, line 34: Actually shouldn't these be in the Logical Operators category?

richardwerkman:
We can also mutate `is` to `is not` and the other way around.

rouke-broersma (dismissed):
I don't think these mutations are different enough from the equality or logical Mutators to warrant it's own 'PatternMatching' category.

docs/mutations.md
```diff
@@ -17,7 +17,7 @@ Do you have a suggestion for a (new) mutator? Feel free to create an [issue](htt
 | `/` | `*` |
 | `%` | `*` |
 
-## Equality Operators (_equality_)
+## Equality Operators (_equality_, _patternMatching_)
```
richardwerkman, line 20: I also don't think we want to add patternMatching as a second name for these mutations

docs/mutations.md
richardwerkman:
```
| `and` | `or` |
| `or` | `and` |
```
Should be in Logical Operators

```
| `is` | `is not` |
| `is not` | `is` |
```

Should be in Equality Operators (if we add them in this PR)

src/Stryker.Core/Stryker.Core/Mutators/Mutator.cs
```diff
@@ -42,7 +42,9 @@ public enum Mutator
         [MutatorDescription("Null coalescing")]
         NullCoalescing,
         [MutatorDescription("Math methods")]
-        Math
+        Math,
+        [MutatorDescription("Pattern matching expressions")]
+        PatternMatching
```
richardwerkman, line 47: This is not a new mutator type. They fall under Logical and Equality operators

docs/mutations.md
yakivyusin: Done

docs/mutations.md
yakivyusin: Done

src/Stryker.Core/Stryker.Core/Mutators/Mutator.cs
yakivyusin: Done

richardwerkman (approved):
Thanks for this feature :) Sorry it took a while to review and merge but it will be released soon

richardwerkman:
Release is live 👍

PR #2321, opened 2022-11-17
docs: Update docs for Math mutator

yakivyusin:
* wrap methods with the code tag on mutations.md page
* add the Math mutator level to configuration.md


richardwerkman:
Thanks for the small fix :)

PR #2340, opened 2022-11-25
feat(solution context run): discover solution file

### LINKED ISSUE #2282: Discover solution file
Currently for a solution context run you need to pass the full path to the solution file.

We need to accept relative paths. Also we could discover the solution file from disk in the current directory. We need to check if exactly one solution file is found and *no* csproj file is found. In that case we can enable solution context.

Liam-Rougoor:
Closes #2282


src/Stryker.Core/Stryker.Core/Options/StrykerInputs.cs
```diff
@@ -126,7 +126,7 @@ public StrykerOptions ValidateAll()
                 MutationLevel = MutationLevelInput.Validate(),
                 DevMode = DevModeInput.Validate(),
                 MsBuildPath = MsBuildPathInput.Validate(_fileSystem),
-                SolutionPath = SolutionInput.Validate(_fileSystem),
+                SolutionPath = SolutionInput.Validate(basePath, _fileSystem),
```
rouke-broersma, line 129: How do we make sure this is only done if a project file is not found?

Liam-Rougoor: My implementation prioritizes solution files over project files by design, (possibly wrongly?) assuming solution files don't exist in project directories.

Liam-Rougoor: If no project filter or solution file are specified, I'd assume it should prioritize a solution file over (possibly multiple) project files.

rouke-broersma: You know what they say about assumptions :) 

We need to decide how to handle this and we should probably document the implemented behavior.

Liam-Rougoor: Solution context is prioritized for now, until it might become an issue. This prioritization is documented in [operating modes](https://github.com/stryker-mutator/stryker-net/blob/2282-discover-solution-files/docs/operating-modes.md).

Issue #2356, opened 2022-12-13
When setting the --msbuildPath option, the nuget restore should also use this msbuild to restore all the nuget packages


MischaVreeburg:
I'm trying to use [stryker.net](http://stryker.net/) to run against a large solution (85 projects) in azure devops pipelines. The projects are sdk style files, but target .NET 4.8 full framework, due to some dependencies.
On my local machine, this works fine, where I have vs2019 and vs2022 installed, but in the pipeline the [Stryker.NET](http://stryker.net/) run fails because it tries to build/restore the projects using VS2017 enterprise msbuild.
Both VS2017 Enterprise and VS2019 Enterprise are installed on the build machines. I've tried to supply the msbuild path via the --msbuild-path option, but still this will use the VS2017 enterprise msbuild for the nuget restore.

I'd like to see that the --msbuildpath option is also honored by the nuget restore and thus won't throw or otherwise misbehave when  VS2017 Enterprise is installed and a SDK style project targetting .NET48 and using a higher language level than 7.3 is being used, by correctly using the provided msbuild to restore the projects.

As an alternative Stryker.NET should always use the latest version of msbuild, not defaulting to VS2017 Enterprise, as there are currently 2 newer versions of the software out there (2019 and 2022)

Logging from the console illustrating the issue:
PS C:\Users\xxxxxxx> dotnet stryker --project "C:\git\xxxxx\Server\Foundation\Common\code\Foundation.Common.csproj" --solution "c:\git\xxxxx\Server\xxxxx.sln"  --reporter 'json' --reporter 'progress' --test-project "c:\git\xxxxx\server\Feature\xxxxxxxxxxxxxxxxx\test\xxxxxxxx.xxxxxxxxx.Test\xxxxxxxx.xxxxxxxxx.Test.csproj" --verbosity debug --msbuild-path **_"C:\Program Files\Microsoft Visual Studio\2022\Professional\MSBuild\Current\Bin\MSBuild.exe"_**
 ```
  _____ _              _               _   _ ______ _______
  / ____| |            | |             | \ | |  ____|__   __|
 | (___ | |_ _ __ _   _| | _____ _ __  |  \| | |__     | |
  \___ \| __| '__| | | | |/ / _ \ '__| | . ` |  __|    | |
  ____) | |_| |  | |_| |   <  __/ |    | |\  | |____   | |
 |_____/ \__|_|   \__, |_|\_\___|_| (_)|_| \_|______|  |_|
                   __/ |
[...]

PR #2359, opened 2022-12-16
feat(nugetrestore): use MsBuild option during nuget restore

MischaVreeburg:
Code to implement feature request https://github.com/stryker-mutator/stryker-net/issues/2356 (When setting the --msbuildPath option, the nuget restore should also use this msbuild to restore all the nuget packages)

Added 2 commits, 
one to implement the use of the msbuildPath cli option for the nuget restore 
the other one to change the default return of the MsBuildHelper to first look for the highest version of Visual Studio, only then trying to give Visual Studio 2017 Enterprise MsBuild path .

### DIFF
--- a/src/Stryker.Core/Stryker.Core/ToolHelpers/MsBuildHelper.cs
+++ b/src/Stryker.Core/Stryker.Core/ToolHelpers/MsBuildHelper.cs
@@ -1,9 +1,9 @@
-using Microsoft.Extensions.Logging;
-using Stryker.Core.Logging;
-using Stryker.Core.Testing;
 using System.Collections.Generic;
 using System.IO;
 using System.IO.Abstractions;
+using Microsoft.Extensions.Logging;
+using Stryker.Core.Logging;
+using Stryker.Core.Testing;
 
 namespace Stryker.Core.ToolHelpers
 {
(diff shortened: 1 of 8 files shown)


src/Stryker.Core/Stryker.Core/ToolHelpers/MsBuildHelper.cs
```diff
@@ -58,7 +54,10 @@ public string GetMsBuildPath(IProcessExecutor _processExecutor)
                     }
                 }
             }
-
+            if (File.Exists(@"C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\MSBuild.exe"))
```
richardwerkman, line 57: Now that I see this code again. Maybe this path should just be added to the fallbackLocations list. It will result in the same behavior but cleaner code.

MischaVreeburg: Good call, Ill fix it and upload the new version

richardwerkman:
Thanks for the fix! This will be released soon :)

MischaVreeburg:
That would be really  great.

richardwerkman:
Release is live 👍

Issue #2372, opened 2023-01-03
Have `highest-available` option for `target-framework`


Marvin-Brouwer:
Hi, 

I noticed that the documentation states the following:
![image](https://user-images.githubusercontent.com/5499778/210403321-f4932ddd-fff0-4c10-b5a9-3da3263661c6.png)

This makes sense to be random, however, I don't like that I have to update that setting in the config now every time I want to add a new runtime.
I have a multi-targeted project with a `net7` test-suite so I want `net7` stryker only.

I'd like to suggest some additional settings for the [`target-framework` setting](https://stryker-mutator.io/docs/stryker-net/configuration/#target-framework-string):
Consider projects with these targets:
```xml
<TargetFrameworks>net6.0;net7.0;netstandard2.1</TargetFrameworks>
```

- `highest-available` => This would look at the options and select the newest: `net7.0`
- `project-default` => This would mean the project visual studio would default to when you first open the sln: `net6.0`

The names are just a suggestion, at the end I'm just looking for a way to automatically select the highest framework available.

Alternatively, auto detecting the test-suite only has `net7` support would also help me out.
However, that just sounds more difficult.

rouke-broersma:
`This would mean the project visual studio would default to when you first open the sln`, don't really know what this means, there is no default target framework as far as I know. Highest available could work, just non-trivial to implement.

Marvin-Brouwer:
To clarify the `This would mean the project visual studio would default to when you first open the sln`.
If you've never opened a solution, so you don't have a `.vs` folder (I think it's actually about the `.suo` but that's in there)
Visual studio needs to select a version here:
![image](https://user-images.githubusercontent.com/5499778/210411826-bc64324a-4e18-4d67-9d3b-ad8eaf735c68.png)
I don't think it's documented anywhere, but it appears that it just looks at the `<TargetFrameworks>` node, splits it by `;` and selects the first it finds:
![image](https://user-images.githubusercontent.com/5499778/210411996-8c92e291-fedb-40a1-b229-34796414efb4.png)
I may be mistaken, but that's what it looks like to me.

I'd say that's a lot less trivial to implement than figuring out what the newest available version is.

rouke-broersma:
There's no guarantee we have that information in that order because we don't read the csproj itself. The csproj gets read by msbuild and we take the result from msbuild (through a 3rd party library) . If both msbuild and this library maintain the ordering that might work. That's a big if and undocumented behavior basically.

[...]

PR #2373, opened 2023-01-04
Update docs about TFM

psfinaki:
As per this thread: #2372

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -138,11 +138,11 @@ Valid language versions:
 
 ### `target-framework` <`string`>
 
-Default: randomly selected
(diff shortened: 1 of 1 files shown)


rouke-broersma:
We always selected the first, I think we say 'random' because we don't really know what 'first' means since it is processed by msbuild and then buildalyzer.

psfinaki:
Hmm fair. I just realized that either way it can be a bit confusing. But maybe it's better to leave things as they are, that's on you then

rouke-broersma:
As discussed we can't guarantee ordering so from our perspective the targetframework that's chosen is 'random'.

PR #2377, opened 2023-01-06
fix(no test projects): null-reference exception when no test projects are found

### LINKED ISSUE #2374: Null reference exception if no test project in solution
**Describe the bug**
If there is no test project in solution, Stryker throws null reference exception.

**Logs**
Version: 3.4.0
[09:20:36 INF] Identifying projects to mutate in /builds/apigateway/ApiGateway.sln. This can take a while.
[09:20:[40](http://gitlab/apigateway/-/jobs/51092#L40) INF] Found 1 source projects
[09:20:40 INF] Found 0 test projects
[09:20:[42](http://gitlab/apigateway/-/jobs/51092#L42) WRN] No test projects could be found for /builds/apigateway/src/ApiGateway/ApiGateway.csproj
[09:20:42 INF] Time Elapsed 00:00:06.1619361
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Stryker.Core.StrykerRunner.RunMutationTest(IStrykerInputs inputs, ILoggerFactory loggerFactory, IProjectOrchestrator projectOrchestrator)
   at Stryker.CLI.StrykerCli.RunStryker(IStrykerInputs inputs) in /_/src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs:line 93
   at Stryker.CLI.StrykerCli.<>c__DisplayClass10_0.<Run>b__0() in /_/src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs:line 68
   at McMaster.Extensions.CommandLineUtils.CommandLineApplication.<>c__DisplayClass1[43](http://gitlab/apigateway/-/jobs/51092#L43)_0.<OnExecute>b__0(CancellationToken _)
   at McMaster.Extensions.CommandLineUtils.CommandLineApplication.ExecuteAsync(String[] args, CancellationToken cancellationToken)
   at McMaster.Extensions.CommandLineUtils.CommandLineApplication.Execute(String[] args)
   at Stryker.CLI.StrykerCli.Run(String[] args) in /_/src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs:line 74
   at Stryker.CLI.Program.Main(String[] args) in /_/src/Stryker.CLI/Stryker.CLI/Program.cs:line 14
Aborted (core dumped)
ERROR: Job failed: exit code 134

**Expected behavior**
If there is no test project in solution, it should skip running instead of throwing exception like other dotnet tools such as dotnet test.

 - OS: Linux Alpine 3.16 x64
 - Type of project: Core
 - Framework Version: .NET 6
 - Stryker Version: 3.4.0

**Additional context**
We have many projects using our CI/CD pipeline. Unfortunately some of them do not have any test project. Therefore they fail at mutation testing stage and exit.

richardwerkman:
fixes #2374

### DIFF
--- a/src/Stryker.Core/Stryker.Core/StrykerRunner.cs
+++ b/src/Stryker.Core/Stryker.Core/StrykerRunner.cs
@@ -162,6 +162,11 @@ namespace Stryker.Core
         /// <returns>The root folder component</returns>
         private IProjectComponent AddRootFolderIfMultiProject(IEnumerable<IProjectComponent> projectComponents, StrykerOptions options)
         {
+            if (!projectComponents.Any())
+            {
+                throw new NoTestProjectsException();
+            }
+
             if (projectComponents.Count() > 1)
             {
                 var rootComponent = new Solution
--- a/src/Stryker.CLI/Stryker.CLI/Program.cs
+++ b/src/Stryker.CLI/Stryker.CLI/Program.cs
@@ -13,6 +13,11 @@ namespace Stryker.CLI
                 var app = new StrykerCli();
                 return app.Run(args);
             }
+            catch (NoTestProjectsException exception)
+            {
+                AnsiConsole.WriteLine(exception.Message);
+                return ExitCodes.Success;
+            }
             catch (InputException exception)
             {
                 AnsiConsole.MarkupLine("[Yellow]Stryker.NET failed to mutate your project. For more information see the logs below:[/]");
--- a/src/Stryker.Core/Stryker.Core/Exceptions/NoTestProjectsException.cs
+++ b/src/Stryker.Core/Stryker.Core/Exceptions/NoTestProjectsException.cs
@@ -1,0 +1,13 @@
+using System;
+using System.Text;
+
+namespace Stryker.Core.Exceptions
+{
+    /// <summary>
+    /// Represents error when no test projects are found in the solution or configured for stryker.
+    /// </summary>
+    public class NoTestProjectsException : Exception
+    {
+        public NoTestProjectsException() : base("No test projects found. Please add a test project to your solution or fix your stryker config.") { }
+    }
+}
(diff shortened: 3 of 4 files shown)


src/Stryker.Core/Stryker.Core/StrykerRunner.cs
```diff
@@ -162,6 +162,11 @@ private void AnalyseCoverage(StrykerOptions options)
         /// <returns>The root folder component</returns>
         private IProjectComponent AddRootFolderIfMultiProject(IEnumerable<IProjectComponent> projectComponents, StrykerOptions options)
         {
+            if (!projectComponents.Any())
```
rouke-broersma, line 165: Couldn't this also happen if there aren't any files?

richardwerkman: I mean technically yes. But then there'd be an empty solution or only a test project. Both make no sense. The "no test project" case seems the only reasonable one.

rouke-broersma: We could fail to resolve any files with buildalyzer, and then end up with no project components. The error then send you on the wrong path.

richardwerkman: I've tried really hard to get separate errors for when there are no test projects vs when there are no projects at all. But I can't get it to work without a huge refactor. 

Basically the issue is that we now skip the initialisation completely if we don't find a test project. So there is no MutationTestProcess created. The same happens when no project is found, because of course we don't create any MutationTestProcess then. But if we want to create a MutationTestProcess for a project without a test project we need to adapt the whole initialisation process to that. I've tried but it's too much work. We also can't throw the exception when we notice a project without any test projects. Because there might be another project that DOES have test project. 

tldr; I've tried but no luck. I say: we keep it like this

PR #2390, opened 2023-01-30
Adding break-on-initial-test-failure option to the config file

psfinaki:
As per discussion here: https://github.com/stryker-mutator/stryker-net/pull/2151#issuecomment-1406919865
This PR basically borrows from the initial commits in the PR above. I also tested this manually.


psfinaki:
Thanks! Will be cool to see a release with the new stuff :)

PR #2392, opened 2023-01-30
chore(BinaryExpressionMutator): Initialize _kindsToMutate statically and inline Mutator type

C-xC-c:
The function `GetMutatorType` works out the `Mutator` type for a `SyntaxKind` for each binary expression using some fairly expensive string methods. Since this information is known statically, we can change `_kindsToMutate` to hold the `Mutator` along with the `SyntaxKinds` for mutation, for a nice performace boost. On a simple expression like `a < b && a > c`, it's somewhere around 20% on my machine.

I've also initialized `_kindsToMutate` statically instead of in the constructor, which is how the rest of the mutators are, but this has no performance implications in the real world since only one instance of `BinaryExpressionMutator` is created.

```
|  Method |     Mean |   Error |  StdDev | Ratio |
|-------- |---------:|--------:|--------:|------:|
| Current | 999.5 ns | 2.26 ns | 2.12 ns |  1.00 |
|  Inline | 791.0 ns | 1.51 ns | 1.42 ns |  0.79 |
```


psfinaki (commented):
Could be nice to have some  real-life testing as well - but from the code point of view the benefit is clear.

psfinaki (approved):

PR #2406, opened 2023-02-11
fix(dashboard reporter): Actually print the hint in cyan

0xced:
Also simplify writing text with Spectre.Console by using the WriteLine method directly with a style. This avoids unnecessary markup parsing.


rouke-broersma:
🙊 😂

richardwerkman:
I've tested your PR and it looks like the same colors to me.

image

Thanks for fixing an embarrassing bug :)

PR #2419, opened 2023-02-22
Adding TFM command line option

psfinaki:
So this adds an option to specify TFM in the command line in addition to the config file.

Some context: in our repo, we are now changing the way we execute mutation testing. Instead of having one TFM value in the config and hoping that random selection works wherever this TFM is not available, we are going to have the information about the desired TFM elsewhere and use it when running Stryker for every particular project. 

This will save some time and make mutation testing results more deterministic.

I checked this manually for a little repro also :)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -141,7 +141,7 @@ Valid language versions:
 ### `target-framework` <`string`>
 
 Default: randomly selected
-Command line: `N/A`  
+Command line: `target-framework`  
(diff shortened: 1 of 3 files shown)


RussKie:
👍

richardwerkman:
Looking good :)
Will be merged soon

psfinaki:
Thanks @richardwerkman! Looking forward for the release.

PR #2429, opened 2023-02-26
feat(mutators): Add string.IsNullOrEmpty and string.IsNullOrWhiteSpace mutations

### LINKED ISSUE #1581: Mutator idea: string.IsNullOrEmpty
When using string.IsNullOrEmpty most of the times only null is tested. 

If we mutate `string.IsNullOrEmpty(x)` into `(x != null)` and `(x != string.Empty)` it's clear both are tested.

yakivyusin:
Closes #1581


docs/mutations.md
```diff
@@ -161,6 +161,11 @@ Do you have a suggestion for a (new) mutator? Feel free to create an [issue](htt
 | `$"foo {bar}"` | `$""` |
 | `@"foo"` | `@""` |
 | `string.Empty` | `"Stryker was here!"` |
+| `string.IsNullOrEmpty(x)` | `x != null` |
+| `string.IsNullOrEmpty(x)` | `x != ""` |
+| `string.IsNullOrWhiteSpace(x)` | `x != null` |
+| `string.IsNullOrWhiteSpace(x)` | `x != ""` |
+| `string.IsNullOrWhiteSpace(x)` | `x.Trim().Length != 0` |
```
psfinaki, line 168: TBH That looks a bit off here, what's the justification for having it?

src/Stryker.Core/Stryker.Core/Mutants/CsharpMutantOrchestrator.cs
```diff
@@ -51,7 +51,8 @@ public CsharpMutantOrchestrator(IEnumerable<IMutator> mutators = null, StrykerOp
                 new NullCoalescingExpressionMutator(),
                 new MathMutator(),
                 new SwitchExpressionMutator(),
-                new IsPatternExpressionMutator()
+                new IsPatternExpressionMutator(),
+                new StringIsNullOrMutator()
```
psfinaki, line 55: StringIsNullOrEmptyMutator? Also, I believe this should be just part of the StringEmptyMutator somehow.

docs/mutations.md
yakivyusin: Do you mean the `x.Trim().Length` mutation? This is check that the whitespace-only string is covered by unit tests. `Trim()` will return the zero-length string for the whitespace-only input.

This mutation also can be rewritten as `x.Trim() != ""` or using regexp.

psfinaki: Right. I think the latter-most form is clearer.

src/Stryker.Core/Stryker.Core/Mutants/CsharpMutantOrchestrator.cs
rouke-broersma: https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/Mutators/MutatorBase.cs#L11

:) If it makes more sense, you can implement IMutator directly.

docs/mutations.md
yakivyusin: Rewritten using regexp

src/Stryker.Core/Stryker.Core/Mutants/CsharpMutantOrchestrator.cs
yakivyusin: Refactored to `StringEmptyMutator`

docs/mutations.md
rouke-broersma: Mhm not sure I agree @psfinaki , `x.Trim() != ""` is probably way more clear than regex, to most people.

psfinaki: Damn I actually meant `x.Trim() != ""`🤦Probably my eyes naturally filter out anything about regex, anyway sorry for the confusioin.

yakivyusin: Refactored with `x.Trim()` )

PR #2448, opened 2023-03-22
fix(String mutations): Generate empty string mutations parenthesized

### LINKED ISSUE #2446: New empty string mutations break in negated expressions
**Describe the bug**
They work for
```if (string.IsNullOrEmpty(s)) {...}```
producing ```if (s!="") {...}```

but they don't work for
```!string.IsNullOrEmpty(s) {...}```
which produces ```(!s!="") {...}```
which causes compile errors and unkilled mutants.

**Expected behavior**
This mutation should be wrapped in extra parens hence producing valid mutants.

 - OS: Windows
 - Type of project: Core
 - Framework Version: .NET7
 - Stryker Version: 3.7.0

richardwerkman:
fixes #2446


Liam-Rougoor:
How about this one? https://github.com/stryker-mutator/stryker-net/pull/2447

PR #2452, opened 2023-03-31
fix(BlockMutator): BlockMutator itself adds 'return default' when block returns a value.

### LINKED ISSUE #2283: Block Mutator should report when it returns default value
When the BlockMutator removes a block that contains a return statement, it implicitly adds a statement that returns default. However, this is not shown in the output report. The code that is displayed in the report would actually result in a compile error in some cases. 

I would like the report to explicitly display when a 'return default' is added. 

**Additional context**
This:
CurrentScenario

Should look something like this:
NewScenario

Liam-Rougoor:
Closes #2283.

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/CsharpMutantOrchestratorTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/CsharpMutantOrchestratorTests.cs
@@ -200,7 +200,7 @@ private bool Out(out string test)
 	return true;
 }";
             string expected = @"void TestMethod()
-{if(StrykerNamespace.MutantControl.IsActive(0)){}else{if(StrykerNamespace.MutantControl.IsActive(19)){
+{if(StrykerNamespace.MutantControl.IsActive(0)){return default;}else{if(StrykerNamespace.MutantControl.IsActive(19)){
 	int i = 0;
 	if (i + 8 == 8)
 	{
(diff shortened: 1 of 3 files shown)


psfinaki:
I haven't read the full discussion but looks nice.

One question though - am I right that this doesn't take into consideration the fact that [default literals](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-version-history#c-version-71) are relatively new in the C# language? (since 7.1). How hard will it be to implement the older version of this (with explicit type, like `default(T)`).

Also, MSDN says 
> You can use default literal expressions in default value expressions when the target type can be inferred.

I cannot think what are the edge cases when the types actually **cannot** be inferred, but that's probably another argument for using explicit syntax.

Liam-Rougoor:
> One question though - am I right that this doesn't take into consideration the fact that [default literals](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-version-history#c-version-71) are relatively new in the C# language? (since 7.1). How hard will it be to implement the older version of this (with explicit type, like `default(T)`).

Good point.

I'm also considering to re-use the `EndingReturnEngine` somehow, which is now responsible for adding the default return under the hood. That one does use the explicit defaults.

Liam-Rougoor:
We are investigating whether this approach is the one we want, or whether we want to fix this issue some other way. Changing the PR to draft.

PR #2469, opened 2023-04-13
feat(reporting): add support for realtime reporting

### LINKED ISSUE #2461: Realtime integration in the HTML reporter in Stryker.NET
This issue depends on stryker-mutator/mutation-testing-elements#2434

---

# Realtime Reporting

Add support for realtime reporting in the HTML reporter in Stryker.NET.

## Implementation

1. When all mutants have been created, open the file in the browser with its initial state.
2. After opening the HTML file, use a `HttpListener` to listen to a certain endpoint with the `text/event-stream` header.
3. When a mutant is tested, send an event to the browser. This event should at least contain a mutant `id` and the new `status`.
   a. It might also be nice to put these mutants in a list, and output them when a client is connected. This could be used to prevent missing mutant data in the browser.
5. When mutation testing is done, send a different event so that the report knows it won't recieve any new updates. 
6. Finally write the complete version to disk (as we currently do).

Also, we should rename `NotRun` to `Pending` (see stryker-mutator/mutation-testing-elements#2424)

## Question(s)

1. Should this be enabled by default or should this be behind the `-o` flag?

xandervedder:
~~Waiting for https://github.com/stryker-mutator/mutation-testing-elements/issues/2434 to be completed~~

closes #2461


src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
```diff
@@ -234,6 +234,11 @@ public string ProjectVersion
         /// </summary>
         public bool BreakOnInitialTestFailure { get; set; }
 
+        /// <summary>
+        /// The port used when realtime reporting is enabled (`-o` flag).
+        /// </summary>
+        public int? Port { get; init; }
```
rouke-broersma, line 240: It this necessary as an option? Can we not just request any free high-range port?

xandervedder: Yeah I was already looking at this:
```csharp
static int FreeTcpPort()
{
  TcpListener l = new TcpListener(IPAddress.Loopback, 0);
  l.Start();
  int port = ((IPEndPoint)l.LocalEndpoint).Port;
  l.Stop();
  return port;
}
```

I suppose I should use this instead

src/Stryker.Core/Stryker.Core/Reporters/Html/Realtime/Events/SseEvent.cs
```diff
@@ -0,0 +1,16 @@
...
+public class SseEvent<T>
+{
+    public string EventName { get; init; }
+    public T Data { get; init; }
+
+    public override string ToString()
+    {
+        var data = JsonSerializer.Serialize(Data,
+            new JsonSerializerOptions { PropertyNamingPolicy = JsonNamingPolicy.CamelCase });
+        return $"event: {EventName}\ndata:{data}\n\n";
```
rouke-broersma, line 14: If you want these fields in your event, why not just make them actual fields in your event?

src/Stryker.Core/Stryker.Core/Reporters/Html/Realtime/Events/SseEvent.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace Stryker.Core.Reporters.Html.Realtime.Events;
+
+public class SseEvent<T>
+{
+    public string EventName { get; init; }
+    public T Data { get; init; }
+
+    public override string ToString()
+    {
+        var data = JsonSerializer.Serialize(Data,
```
rouke-broersma, line 12: It would be better to leave the transmission format up to the sender (either real time mutant handler or sseserver) and treat sseevent as a domain object

src/Stryker.Core/Stryker.Core/Reporters/Html/Realtime/SseServer.cs
```diff
@@ -0,0 +1,69 @@
...
+            var writer = new StreamWriter(response.OutputStream);
+            _writers.Add(writer);
+        }
+    }
+
+    public void SendEvent<T>(SseEvent<T> @event)
+    {
+        foreach (var writer in _writers)
+        {
+            writer.Write(@event.ToString());
```
rouke-broersma, line 58: If the format should be json, use json serialization here instead of directly calling ToString where the ToString implements the serialization.

src/Stryker.Core/Stryker.Core/Reporters/Html/Realtime/Events/SseEvent.cs
xandervedder: I'm not sure what you mean here 😕

src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
rouke-broersma: I also found that stackoverflow suggestion but as far as I can tell what you would be doing is request a port, then release the port, then hope the port is still available by the time you create your new listener.

I'm not sure if all that is necessary, I think you could use 0 for the port number in your current implementation, then check on the listener which port it received and getting that injected into the report.

src/Stryker.Core/Stryker.Core/Reporters/Html/Realtime/Events/SseEvent.cs
rouke-broersma: Why does SseEvent not include the properties event and data and let the json serializer serialize them, why do you add these fields manually in the ToString.

src/Stryker.Core/Stryker.Core/Reporters/Html/Realtime/SseServer.cs
```diff
@@ -0,0 +1,86 @@
...
+        Task.Run(ListenForConnectionsAsync);
+    }
+
+    private async Task ListenForConnectionsAsync()
+    {
+        while (true)
+        {
+            if (!_listener.IsListening)
+            {
+                return;
```
rouke-broersma, line 50: This seems like an error condition. When does this happen and should we log this and/or throw an exception?

src/Stryker.Core/Stryker.Core/Reporters/Html/Realtime/Events/SseEvent.cs
rouke-broersma: I realize that this is something I should have noticed in mutation testing elements, but I did not :D 
Just wondering if there's a technical reason or if this is just how you designed the event data.

src/Stryker.Core/Stryker.Core/Reporters/Html/Realtime/RealtimeMutantHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+    private readonly ISseServer _server;
+
+    public RealtimeMutantHandler(StrykerOptions options, ISseServer server = null)
+        => _server = server ?? new SseServer(options);
+
+    public void OpenSseEndpoint() => _server.OpenSseEndpoint();
+
+    public void CloseSseEndpoint()
+    {
+        _server.SendEvent(new SseEvent<string> { EventName = "finished", Data = "" });
```
rouke-broersma: Shouldn't this use the enum

src/Stryker.Core/Stryker.Core/Reporters/Html/Realtime/Events/SseEvent.cs
xandervedder: > I think we should document the different event's schemas in mutation-testing-elements.

This is something I definitely want to do in the near future 😄 .

> I still feel that abusing ToString is not the best way to go.

Maybe something like `ToPayload()` would fit better?

src/Stryker.Core/Stryker.Core/Reporters/Html/Realtime/SseServer.cs
xandervedder: I added this to satisfy sonarcloud, but I guess we could also put this condition in the `while (true) ...` loop

rouke-broersma: Either this is something that can happen, and we should handle this gracefully (throwing or logging) so the user knows that the realtime reporting is not working, or this is a non-issue and should be marked as false positive in sonarcloud.

src/Stryker.Core/Stryker.Core/Reporters/Html/Realtime/Events/SseEvent.cs
rouke-broersma: `Serialize` seem fitting to me.

I would suggest using a multiline interpolated string (https://stackoverflow.com/questions/33236592/multiline-c-sharp-interpolated-string-literal) to improve readability.

```
$@"
event: {EventName}
data:{data}
";
```

Or using a StringBuilder. 

The newline between messages should however be added by the sseserver.SendEvent since it's not really a part of the event itself.

src/Stryker.Core/Stryker.Core/Reporters/Html/Realtime/SseServer.cs
xandervedder: I think it's a non-issue, but I like the condition to include `_listener.IsListening`, so it will stop doing the loop once the listener the realtime reporting is done.

xandervedder:
By the way, since we use a random port now for listening, I could probably test this. But do we want that?

src/Stryker.Core/Stryker.Core/Reporters/Html/Realtime/SseServer.cs
rouke-broersma: Okay, makes sense.

rouke-broersma:
> By the way, since we use a random port now for listening, I could probably test this. But do we want that?

What would you propose to test? Receiving a port number?

xandervedder:
The actual implementation? But that would require to both run and listen to said server or we can just do `[ExcludeFromCodeCoverage]`

rouke-broersma:
> The actual implementation? But that would require to both run and listen to said server or we can just do `[ExcludeFromCodeCoverage]`

If we can test it we probably should ;)

src/Stryker.Core/Stryker.Core/Reporters/Html/Realtime/SseServer.cs
```diff
@@ -0,0 +1,78 @@
...
+            _writers.Add(writer);
+        }
+    }
+
+    public void SendEvent<T>(SseEvent<T> @event)
+    {
+        var serialized = @event.Serialize();
+        foreach (var writer in _writers)
+        {
+            writer.Write($"{serialized}\n\n");
```
rouke-broersma, line 67: Use Environment.NewLine

src/Stryker.Core/Stryker.Core/Reporters/Html/HtmlReporter.cs
```diff
@@ -0,0 +1,148 @@
...
+        }
+        else
+        {
+            _console.WriteLine(reportUri, green);
+        }
+
+        _console.WriteLine("You can open it in your browser of choice.", green);
+    }
+
+    private void OpenReportInBrowser(IReadOnlyProjectComponent reportComponent, string reportPath, TestProjectsInfo testProjectsInfo)
```
rouke-broersma, line 68: private methods should not be in-between public methods, move them to below the last public method

src/Stryker.Core/Stryker.Core/Reporters/Html/HtmlReporter.cs
```diff
@@ -0,0 +1,148 @@
...
+    {
+        WriteHtmlReport(reportPath, reportComponent, testProjectsInfo);
```
rouke-broersma, line 70: Why does OpenReportInBrowser also write the report to disk?

src/Stryker.Core/Stryker.Core/Reporters/Html/HtmlReporter.cs
```diff
@@ -0,0 +1,148 @@
...
+
+    public void OnAllMutantsTested(IReadOnlyProjectComponent reportComponent, TestProjectsInfo testProjectsInfo)
+    {
+        var reportPath = BuildReportPath();
+        WriteHtmlReport(reportPath, reportComponent, testProjectsInfo);
+
+        if (_options.ReportTypeToOpen == ReportType.Html)
+        {
+            _mutantHandler.CloseSseEndpoint();
+            return;
```
rouke-broersma, line 47: I don't see why we can't print the file location to the console anyway, the user might have closed the report in-between or something and then they can easily re-open it.

src/Stryker.Core/Stryker.Core/Reporters/Html/HtmlReporter.cs
xandervedder: Good question, I'll separate it.

PR #2491, opened 2023-04-28
feat: Record for each unit test which mutant they kill

### LINKED ISSUE #1784: Disable-bail available as a cli option
We should make disable-bail available as a cli option so that users can on-demand request a html report with all test information. Relevant after #1565

_Originally posted by @rouke-broersma in https://github.com/stryker-mutator/stryker-net/issues/1565#issuecomment-962469428_

### LINKED ISSUE #2489: Mark tests as killing if they fail during mutation test run.
Right now tests are never marked as killing, even if they fail.

Mutants should track which tests failed, and display in the HTML report. To do this correctly we should run all covering tests for each mutant, and not stop when the first test fails. We already have an option for this `disable-bail`, should this be enabled by default when using the HTML reporter? Or should we explain clearly in the docs how to use this feature correctly.

richardwerkman:
closes #2489
closes #1784


richardwerkman:
Somehow it looks like disable bail doesn't actually work. I have to look into this more.

JazzGlobal:
What's the status on this?

JazzGlobal:
> @JazzGlobal richard will finish this when they have time to do so

Of course. Was wondering if richard had found anything out after saying he would "look into this more". No rushing was intended.

rouke-broersma:
Cool! If you have any questions or if you need help with something you can reach us on slack or in the discussions as well 👌👌

rouke-broersma:
What doesn't work irt disable bail?
image

Issue #2470, opened 2023-04-14
Refactor all files to use file-scoped namespaces


xandervedder:
From:
```cs
namespace X.Y.Z 
{
    public class Foo 
    {
        // ...
    }
}
```

To:
```cs
namespace X.Y.Z;

public class Foo
{
    // ...
}
```

ysbakker:
Will pick this up during the hacktoberfest event, so we can avoid merge conflicts with existing PRs.

KonH:
Any updates about plans on that refactor? I also would like to make PR with that refactor, but there are still 9 open PRs which may leads to merge issues.

rouke-broersma:
@KonH unfortunately PRs were being generated faster than we could review them 😅 I think the potential of this change does not necessitate the disruption it would cause at this time unfortunately.

PR #2500, opened 2023-05-06
chore(solution): Refactor all files to use file-scoped namespaces (#2470).

jameswiseman76:
Refactor all files to use file-scoped namespaces stryker-net codebase to use C# 10 file-scoped namespaces

(https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-10#file-scoped-namespace-declaration)


rouke-broersma:
If you would be willing to, you could now recreate this PR and I'll accept it.

PR #2520, opened 2023-05-12
chore: Use file scoped namespaces (#2470) (And pace 'usings' within namespace declaration IDE0065)  

ghost:


src/Stryker.Core/Stryker.Core/Initialisation/ProjectFileReader.cs
```diff
@@ -1,147 +1,145 @@
+namespace Stryker.Core.Initialisation;
 using System.Collections.Generic;
 using System.Linq;
-using Buildalyzer;
+using global::Buildalyzer;
```
rouke-broersma, line 4: I've seen this in some places specifically for Buildalyzer, why?

jameswiseman76: So, looks like a side-effect of the namespace/using moving. I'd need to read exactly why, but removing it makes the exposed types of that namespace unavailable.

I'll hold fire on this and dig some more. It might be that moving the usings inside the namespace is a different PR :-)

rouke-broersma:
Whenever you get around to this again, if you still want to fix this, please create a new PR :)

PR #2524, opened 2023-05-12
chore: Fix long standing sonar issues

dupdob:
I have fixed most, if not all of sonar alerts. I say most because I cannot be sure from a PR as I am not able to see any full project analysis, so I may have missed a few.
Most of the adjustments were cosmetic, but I did some refactoring for the design to better match Sonar rules better. Did add also a bunch of `//NOSONAR` comments were there was no sensible way to make sonar happy AND the source code was indeed correct.


dupdob:
that was quick 👍

dupdob:
yeah, only 7 remainings

PR #2526, opened 2023-05-16
chore: Fix remaining sonar issues

dupdob:
hi
this should fix 6 out of the 7 remaining.
the last one is a todo that I think should remain. At least, I am not sure what it means, hence I can't deal with it 😄


rouke-broersma:
Thank you for this! I think the TODO is mine but I am not certain what I meant by it :D

Issue #2538, opened 2023-05-19
Unhandled exception when single file in multiple projects


CzBuCHi:
**Describe the bug**
I have file that is imported in every single project in solution (SharedAssemblyInfo.cs to be precise).
When running stryker on whole solution it ends up with this exception:

```
Unhandled exception. System.ArgumentException: An item with the same key has already been added. Key: c:\projects\dotnet\Centaur\5.0\_build\SharedAssemblyInfo.cs
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at Stryker.Core.Reporters.Json.JsonReport.AddTestFiles(TestProjectsInfo testProjectsInfo) in /_/src/Stryker.Core/Stryker.Core/Reporters/Json/JsonReport.cs:line 69
   at Stryker.Core.Reporters.Json.JsonReport..ctor(StrykerOptions options, IReadOnlyProjectComponent mutationReport, TestProjectsInfo testProjectsInfo) in /_/src/Stryker.Core/Stryk                                                   er.Core/Reporters/Json/JsonReport.cs:line 29
   at Stryker.Core.Reporters.Json.JsonReport.Build(StrykerOptions options, IReadOnlyProjectComponent mutationReport, TestProjectsInfo testProjectsInfo) in /_/src/Stryker.Core/Stryk                                                   er.Core/Reporters/Json/JsonReport.cs:line 32
   at Stryker.Core.Reporters.Html.HtmlReporter.WriteHtmlReport(String filePath, IReadOnlyProjectComponent reportComponent, TestProjectsInfo testProjectsInfo) in /_/src/Stryker.Core                                                   /Stryker.Core/Reporters/Html/HtmlReporter.cs:line 110
   at Stryker.Core.Reporters.Html.HtmlReporter.OnMutantsCreated(IReadOnlyProjectComponent reportComponent, TestProjectsInfo testProjectsInfo) in /_/src/Stryker.Core/Stryker.Core/Re                                                   porters/Html/HtmlReporter.cs:line 74
   at Stryker.Core.Reporters.BroadcastReporter.OnMutantsCreated(IReadOnlyProjectComponent reportComponent, TestProjectsInfo testProjectsInfo) in /_/src/Stryker.Core/Stryker.Core/Re                                                   porters/BroadcastReporter.cs:line 26
[...]

PR #2542, opened 2023-05-20
fix: exception when source files in more than one test project

dupdob:
Fix issue #2538


rouke-broersma:
@dupdob I don't know why but both of your recent PRs for some reason have an infinite wait in the unit test in the pipeline, I cannot reproduce this locally..

dupdob:
I have not idea what is happening. From what I understands looking at Azure logs, the tests don't even start...

dupdob:
Since PRs impact very different files, it looks like a flaky test situation, that is probably triggered by a change in timing, which can be due to many things: new tests, tests order change, upgrades/changes on azure side...

dupdob:
Well, my first guess was clearly spot on: the problem comes from SseServerTests: those are flaky and prone to deadlock.
But not sure how to get out of these...

rouke-broersma:
I'll ask @xandervedder to take a look

xandervedder:
I think what we might need to do is use a different http listener implementation (one that is easier to test), but that would require a new dependency. I also struggled a lot with this unfortunately.

Issue #2549, opened 2023-05-24
Stryker hangs for an empty code-test project pair


psfinaki:
**Describe the bug**
So yeah, it just hangs, after outputting this:

```
Version: 3.9.0

[16:57:49 INF] Analysis starting.
[16:57:59 INF] Found project C:\Users\xxx\Desktop\Repro\Code\Code.csproj to mutate.
[16:57:59 INF] Analysis complete.
[16:57:59 INF] Building test project C:\Users\xxx\Desktop\Repro\Tests\Tests.csproj (1/1)
[16:58:07 WRN] No test detected for project 'C:\Users\xxx\Desktop\Repro\Tests\Tests.csproj'. No cause identified.
[16:58:07 INF] Number of tests found: 0 for project C:\Users\xxx\Desktop\Repro\Code\Code.csproj. Initial test run started.
```

**Expected behavior**
Create an empty report or an error (not to crash the solution mode though).

**Additional context**
Note this is **a regression**, compared to at least 3.7.1, so we can't update. Prior to that, this was happening:
```
No test has been detected. Make sure your test project contains test and is 
compatible with VsTest.
```

Repro, for reference: [Repro.zip](https://github.com/stryker-mutator/stryker-net/files/11556157/Repro.zip)

dupdob:
thanks
I had a quick look: I do reproduce the problems. So Stryker no longer stops when no tests are found (new behavior, quite deliberate), but fails to detect the end of (useless) initial test session.
Should be easy to fix 😄

dupdob:
TIL that VsTest does not send any notifications if there is no test in the assembly, despite claiming otherwise in the log file.
And since it is the initial test session, there is no timeout set inside Stryker. As a reminder, Stryker adds an extra timeout logic on top of VsTest ones. This is due because: 
1. VsTest does not report time out situations, those look like normal runs, with some tests missing, or not (and they be missing for various reasons)
2. VsTest timeout logic may fail and the process may have to be killed.

That being said, I see two problems to fix here:
1. Stryker should not run test sessions when there is no test at all
2. Stryker should rely solely on VsTest when there is no timeout provided (as during initial tests)

psfinaki:
Damn VsTest is really evil!

PR #2550, opened 2023-05-25
fix(testrunner): fix scenario where vstest would hang if no tests were detected

dupdob:
Fix issue #2549 : Stryker now warns when no test are present.
Fix issue #2559 : Stryker now disables parallelism for MsTest
Also implement a revised strategy for handling hanged VsTest:
1. For initial test session: wait for VsTest to stop, then wait for the end notification (if not received yet). No failover available for the time being.
2. For other test sessions: 
      1. wait for the end notification: if it times out, this session is discarded
      2. then, wait for the end of the session: if it times out, we assume VsTest hanged during clean up phase, it will be recycled, but the session is discarded.
      1. if anything timed out, VsTest console is recycled
      1. if the test session is discarded, a new attempt is made
      1. an exception is thrown after 5 failed attempts

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -205,7 +205,7 @@ stages:
       parameters:
         testName: 'windows-netcore'
         workingDirectory: 'integrationtest/TargetProjects'
-        strykerCommands: --solution '$(Agent.BuildDirectory)\s\integrationtest\TargetProjects\IntegrationTestApp.sln'
+        strykerCommands: --solution 'IntegrationTestApp.sln'
     - task: DotNetCoreCLI@2
       displayName: 'Assert integration test results'
       inputs:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -197,6 +197,10 @@ stages:
       parameters:
         testName: 'windows-netframework'
         workingDirectory: 'integrationtest/TargetProjects/NetFramework/FullFrameworkApp.Test'
+    - script: $(Agent.BuildDirectory)/tools/dotnet-stryker
+      workingDirectory: 'integrationtest/TargetProjects/EmptyTestProject'
+      failOnStderr: false
+      continueOnError: true
     - task: DotNetCoreCLI@2
       displayName: 'Assert integration test results'
       inputs:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -203,9 +203,9 @@ stages:
       continueOnError: true
     - template: pipeline-templates/run-integration-test-steps.yml
       parameters:
-        testName: 'windows-solution'
+        testName: 'windows-netcore'
         workingDirectory: 'integrationtest/TargetProjects'
-        strykerCommands: --solution 'integrationtest/TargetProjects/IntegrationTestApp.sln'
+        strykerCommands: --solution '$(Agent.BuildDirectory)\s\integrationtest\TargetProjects\IntegrationTestApp.sln'
     - task: DotNetCoreCLI@2
       displayName: 'Assert integration test results'
       inputs:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -203,9 +203,9 @@ stages:
       continueOnError: true
     - template: pipeline-templates/run-integration-test-steps.yml
       parameters:
-        testName: 'linux-netcore'
+        testName: 'windows-solution'
         workingDirectory: 'integrationtest/TargetProjects'
-        strykerCommands: --solution '$(Agent.BuildDirectory)/s/integrationtest/TargetProjects/IntegrationTestApp.sln'
+        strykerCommands: --solution 'integrationtest/TargetProjects/IntegrationTestApp.sln'
     - task: DotNetCoreCLI@2
       displayName: 'Assert integration test results'
       inputs:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -197,10 +197,7 @@ stages:
       parameters:
         testName: 'windows-netframework'
         workingDirectory: 'integrationtest/TargetProjects/NetFramework/FullFrameworkApp.Test'
-    - script: $(Agent.BuildDirectory)/tools/dotnet-stryker
-      workingDirectory: 'integrationtest/TargetProjects/EmptyTestProject'
-      failOnStderr: false
-      continueOnError: true
+
     - task: DotNetCoreCLI@2
       displayName: 'Assert integration test results'
       inputs:
@@ -225,6 +222,10 @@ stages:
         command: test
         projects: '**/Validation.csproj'
         arguments: --filter Category=Solution
+    - script: $(Agent.BuildDirectory)/tools/dotnet-stryker
+      workingDirectory: 'integrationtest/TargetProjects/EmptyTestProject'
+      failOnStderr: false
+      continueOnError: true
 
   - job: MacOsTests
     displayName: Run tests on Mac OS
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -201,11 +201,6 @@ stages:
       workingDirectory: 'integrationtest/TargetProjects/EmptyTestProject'
       failOnStderr: false
       continueOnError: true
-    - template: pipeline-templates/run-integration-test-steps.yml
-      parameters:
-        testName: 'windows-netcore'
-        workingDirectory: 'integrationtest/TargetProjects'
-        strykerCommands: --solution 'IntegrationTestApp.sln'
     - task: DotNetCoreCLI@2
       displayName: 'Assert integration test results'
       inputs:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -201,6 +201,11 @@ stages:
       workingDirectory: 'integrationtest/TargetProjects/EmptyTestProject'
       failOnStderr: false
       continueOnError: true
+    - template: pipeline-templates/run-integration-test-steps.yml
+      parameters:
+        testName: 'linux-netcore'
+        workingDirectory: 'integrationtest/TargetProjects'
+        strykerCommands: --solution '$(Agent.BuildDirectory)/s/integrationtest/TargetProjects/IntegrationTestApp.sln'
     - task: DotNetCoreCLI@2
       displayName: 'Assert integration test results'
       inputs:
--- a/src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/VsTestMockingHelper.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/VsTestMockingHelper.cs
@@ -431,8 +431,7 @@ public class VsTestMockingHelper : TestBase
             }));
 
     protected Mock<IVsTestConsoleWrapper> BuildVsTestRunnerPool(StrykerOptions options,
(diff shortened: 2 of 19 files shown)


psfinaki:
Thanks for the fix! Maybe worth adding an integration test with an empty project?

rouke-broersma:
Looks like it's a bit too aggressive, it marks the integration tests as having no tests and/or hanging.

image

dupdob:
lol
sorry, checking that ASAP

dupdob:
ok, got it. I introduced a hard coded 2 seconds timeout to the initial tests.
Fixing it

dupdob:
should be ok now

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
```diff
@@ -215,63 +239,70 @@ void HandlerUpdate(object sender, EventArgs e)
...
+                    _logger.LogError($"{RunnerId}: Test session ended with code {strykerVsTestHostLauncher.ErrorCode}");
                 }
-                _vsTestFailed = true;
-
             }
-            else if (strykerVsTestHostLauncher.ErrorCode > 0)
+            else
             {
-                _logger.LogError($"{RunnerId}: Test session ended with code {strykerVsTestHostLauncher.ErrorCode}");
+                // hard wait
```
psfinaki, line 291: Are you sure this is a good idea? By far I don't have much understanding in the VSTest stuff, that's just paranoia from dev experience - having random problems because of hard waiting.

Maybe some very gracious timeout could be safer? Or, well, at least comment justifying the decision :)

dupdob: hummm
I think I will comment. But I would gladly hear what a gracious timeout could be?
Note that nothing have actually changed here, I am merely restoring the historical behavior, that has been lost because I introduced a grace delay, hence transforming a null value to a 2 seconds timeout.

dupdob: but it made me think of how the design should be improved

psfinaki: Well that probably depends on possible code flows that can lead here. I mean, I just wanted to share the fear that some other edge case project will fall here and things will hang forever. Not that I know what exactly.

What's the expected time to wait here, if you have any idea? For example, if usually it'd take a few seconds to wait, we could add a timeout of 5 mins and then throw something like "Hey, something weird is going on with the project X, report this on GitHub".

dupdob: yep, I reverted to the original implementation, which was delivering the expected results. I introduced changes as an attempt to work around the 'no test' problem, discovering later than nothing could be done. But I forgot to restore the original code.

dupdob:
I revised the VsTest timeout logic again (a good night sleep does wonder) as I realized it was needed for some edge cases.
Indeed, Stryker will no longer attempts to run a test session without any known tests in it, but it may still happen for some other reason, such as some test filter.

On other news, I removed everything related to 'VsTest dit not' error messages. Previous assumptions was that no event whatsoever meant VsTest did not run/start properly, but it happens also when there is no test to run in the assembly.
So the message was misleading

psfinaki:
The logic seems to be a bit more streamlined now, that's good :) 

But I still think an int test will be beneficial since this has broken a couple of times already.

dupdob:
Update: PR switched to draft as I have a reproduceable error in solution mode.
Which is cool: another problem will be fixed!

src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/VsTestMockingHelper.cs
```diff
@@ -130,19 +130,24 @@ protected SourceProjectInfo BuildSourceProjectInfo(IEnumerable<Mutant> mutants =
...
         => new(name, framework == TestFrameworks.xUnit ? _xUnitUri : _NUnitUri, _testAssemblyPath) { Id = new Guid(), DisplayName = displayName ?? name };
 
     private Microsoft.VisualStudio.TestPlatform.ObjectModel.TestCase FindOrBuildCase(string testResultId) => TestCases.FirstOrDefault(t => t.FullyQualifiedName == testResultId) ?? BuildCase(testResultId);
 
-    private static void MockTestRun(ITestRunEventsHandler testRunEvents, IReadOnlyList<TestResult> testResults,
+    private static Task MockTestRun(ITestRunEventsHandler testRunEvents, IReadOnlyList<TestResult> testResults,
         Microsoft.VisualStudio.TestPlatform.ObjectModel.TestCase timeOutTest = null) =>
         Task.Run(() =>
         {
+            if (testResults.Count == 0)
```
psfinaki, line 146: Just thinking out loud: from a design perspective, this feels like a bit too much of white-box testing - I would expect this to be a transparent implementation detail in the code itself.

src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/VsTestRunnerPoolTests.cs
```diff
@@ -156,6 +179,26 @@ public void DetectTimeout()
             result.TimedOutTests.IsEmpty.ShouldBeFalse();
         }
 
+        [Fact]
+        public void RetryHangedSession()
```
psfinaki, line 183: Okay sorry I cannot skip this anymore! _Hanged_ is used only for the method of execution, otherwise it's _hung_ :D

dupdob: yep, I did not remember the exact rule. Any useful synonym I could use instead (I am not fan of hung either)
:-)

src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/VsTestMockingHelper.cs
dupdob: I do not think this is white box testing, in the sense that the test verifies what happens inside the _sut_.
Here it is about better faking for VsTest: if you launch VsTest with an empty test assembly, you will not get any event at all. While I suspect this is a bug on VsTest part, we still need to handle this gracefully

Don't you think?

psfinaki: Yeah fair, I didn't look closely enough. Okay then!

src/Stryker.Core/Stryker.Core.UnitTest/TestRunners/VsTestRunnerPoolTests.cs
psfinaki: It could be _freeze_ probably, but I would live with whatever, just wanted to make a point :-P

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
```diff
@@ -153,14 +153,13 @@ private ICollection<Guid> TestCases(IReadOnlyList<Mutant> mutants, Dictionary<in
...
-            bool compressAll = true)
+        private TestRunResult BuildTestRunResult(IRunResults testResults, int expectedTests, int totalCountOfTests, bool compressAll = true)
         {
             var resultAsArray = testResults.TestResults.ToArray();
             var testCases = resultAsArray.Select(t => t.TestCase.Id).ToHashSet();
             var ranTestsCount = testCases.Count;
-            var timeout = !_aborted && ranTestsCount < expectedTests;
-            var ranTests = compressAll && ranTestsCount >= totalCountOfTests ? (ITestGuids)TestGuidsList.EveryTest() : new WrappedGuidsEnumeration(testCases);
+            var timeout = !_cancelled && ranTestsCount < expectedTests;
+            var ranTests = compressAll && totalCountOfTests>0 && ranTestsCount >= totalCountOfTests ? (ITestGuids)TestGuidsList.EveryTest() : new WrappedGuidsEnumeration(testCases);
```
psfinaki, line 164: So thinking out loud here - I guess at some point it could be useful to just write down all the known scenarios here, like:
- 1 source project x 1 test project
- 1 source project x many test projects
- many source projects x one test project
- many source projects x many test projects
- empty source projects
- empty test projects
- ... 

... and then put this in code (read - design patterns, interfaces, whatever). 

It seems like sooner or later in the flow most of these scenarios need special treatment hence I am under the impression that the related code is gradually turning into a bunch of if/else hacks and becoming a bit hard to follow.

My impression can be wrong, that's basically a general reminder to keep an eye on the design in the light of newly appearing edge cases and scenarious.

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
```diff
@@ -195,13 +194,41 @@ private ICollection<Guid> TestCases(IReadOnlyList<Mutant> mutants, Dictionary<in
...
+            Action<IRunResults> updateHandler = null)
+        {
+
+            if (tests.IsEveryTest && !_context.IsValidSourceList(sources))
+            {
+                _logger.LogWarning($"Test assembl{(sources.Count() <2 ? "y does" : "ies do")} not contain any test, skipping.");
+                return new RunEventHandler(_context.VsTests, _logger, RunnerId);
+            }
+
+            for (var attempt = 0; attempt < 5; attempt++)
```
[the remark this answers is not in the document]
dupdob: Done. 
There is no solid justifications, except VsTest having an history of time related bugs. But I think things get better, so I lowered the number of attempts to 3.

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
psfinaki: Well, I put my comment here because these lines just feel like we are starting to take too many circumstances into account, but generally it's unrelated :) 

But yeah I guess I am overthinking here a bit, eventually we might also just end up with some kind of wrapper on top of VsTest which would effectively carry the semantics of "this is how VsTest would work if it was designed properly and implemented without bugs", basically cutting corners and streamlining edge cases.

dupdob: Comment has been added.
VsTestRunner more or less acts as this VsTest wrapper. Maybe a better design will emerge or be found in the coming months. For my own part, I feel like I have a done a lot of redesign in the past year, and I want to focus on fixing things for the time being 😄

psfinaki: Sure, that's fine. I am just sharing thoughts here, no action required :D

azure-pipelines.yml
```diff
@@ -197,6 +197,10 @@ stages:
       parameters:
         testName: 'windows-netframework'
         workingDirectory: 'integrationtest/TargetProjects/NetFramework/FullFrameworkApp.Test'
+    - script: $(Agent.BuildDirectory)/tools/dotnet-stryker
```
rouke-broersma, line 200: I see the empty test project is dotnet6. The windows integration test is the slowest, it would be better to add this to the linux integration test phase as it's the quickest.

dupdob: seems ok

psfinaki:
Cool - thanks for merging and looking forward to the release to test this in the wild :)

PR #2557, opened 2023-06-01
chore: stabilize SSE tests

dupdob:
Clean up the design of SSE server tests so that they are no longer flaky

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Reporters/Html/Realtime/SseServerTest.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Reporters/Html/Realtime/SseServerTest.cs
@@ -1,21 +1,54 @@
 using System;
+using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
 using LaunchDarkly.EventSource;
+using Shouldly;
 using Stryker.Core.Reporters.Html.Realtime;
 using Stryker.Core.Reporters.Html.Realtime.Events;
 using Xunit;
+using Xunit.Sdk;
 
 namespace Stryker.Core.UnitTest.Reporters.Html.Realtime;
 
 public class SseServerTest : TestBase
 {
-    private readonly ISseServer _sut;
+    private readonly SseServer _sut;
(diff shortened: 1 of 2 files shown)


rouke-broersma:
@xandervedder could you review this?

dupdob:
crap: more stable, but still not fully stable

dupdob:
I increased the timeouts, to cover risks for slower build machines.

src/Stryker.Core/Stryker.Core.UnitTest/Reporters/Html/Realtime/SseServerTest.cs
```diff
@@ -1,21 +1,54 @@
...
+    private bool WaitConnection(int timeout)
+    {
+        var watch = new Stopwatch();
+        watch.Start();
+        lock (_lock)
+        {
+            while (!_connected && watch.ElapsedMilliseconds < timeout)
+            {
+                Monitor.Wait(_lock, Math.Max(timeout - (int)watch.ElapsedMilliseconds, 1));
+            }
```
xandervedder, line 44: Is the while loop necessary? What if we only do this:

```suggestion
            Monitor.Wait(_lock, Math.Max(timeout, 1));
```

Should be functionally equivalent, no?

src/Stryker.Core/Stryker.Core.UnitTest/Reporters/Html/Realtime/SseServerTest.cs
```diff
@@ -1,21 +1,54 @@
...
+    private void ClientConnected(object sender, EventArgs e)
+    {
+        lock (_lock)
+        {
+            _connected = true;
+            Monitor.Pulse(_lock);
+        }
+    }
+
+    private bool WaitConnection(int timeout)
```
xandervedder, line 35:
```suggestion
    private bool WaitForConnection(int timeout)
```

dupdob: done

PR #2577, opened 2023-06-19
docs: We don't mutate constants

psfinaki:
We don't mutate constants, right? Inspired by [this question](https://stackoverflow.com/q/76495189/3232646) BTW.


rouke-broersma:
You're right, thanks!

PR #2588, opened 2023-06-23
feat(baseline): Support large reports in azure file share

### LINKED ISSUE #2548: Using AzureFileStorage for baseline results in RequestEntityToLarge error
**Describe the bug**
I'm trying to test out the baseline feature with AzureFileStorage as the provider. The project I'm running is a rather large solution and the total report size amounts up to about 10 MB. At the end, I get a successful report but also get an error that the upload failed because the request entity is to large. I found that the maximum upload size for a file is 4 MB (https://learn.microsoft.com/en-us/troubleshoot/azure/general/request-body-large)  

**Logs**
```
[17:34:48 ERR] Azure File Storage upload failed with statuscode RequestEntityTooLarge and message: <?xml version="1.0" encoding="utf-8"?><Error><Code>RequestBodyTooLarge</Code><Message>The request body is too large and exceeds the maximum permissible limit.
RequestId:5b68d54f-001a-0031-048c-8defc5000000
Time:2023-05-23T15:34:46.9474761Z</Message><MaxLimit>4194304</MaxLimit></Error>
```

**Expected behavior**
I expect the upload to azure file storage to also work with reports bigger then 4MB

 - OS: Windows11
 - Type of project: Core
 -  core 6.0
 - Stryker Version 3.9.0

rouke-broersma:
fix #2548


src/Stryker.Core/Stryker.Core.UnitTest/Baseline/Providers/AzureFileShareBaselineProviderTests.cs
```diff
@@ -1,476 +1,223 @@
...
+
+            Mock.Get(subdirectoryClient).VerifyAll();
+            Mock.Get(subdirectoryClient).VerifyNoOtherCalls();
+
+            Mock.Get(fileClient).VerifyAll();
+            Mock.Get(fileClient).VerifyNoOtherCalls();
+        }
+
+        [Fact]
+        public async Task Save_Report()
```
richardwerkman, line 168: Does this mimic the upload of a large or small file? I think we need tests for both.

Issue #2585, opened 2023-06-22
Mutation tests fails on Ubuntu


RussKie:
**Describe the bug**

I'm bolting mutation tests on to a project. I expect that project contributors will generally work on Windows but the CI will run on Ubuntu.
The MT run successfully on Windows (locally), however those fail on an Ubuntu (both locally under WSL2 and on CI).

There are a bunch of errors like this:
```
Safe Mode! Stryker will try to continue by rolling back all mutations in method. This should not happen, please report this as an issue on github with the previous error message. (2b631064)
```

**Logs**
The logs can be downloaded [here](https://dev.azure.com/dnceng-public/public/_build/results?buildId=316957&view=artifacts&pathAsName=false&type=publishedArtifacts)
![image](https://github.com/stryker-mutator/stryker-net/assets/4403806/69e471b5-ab72-4067-90f2-d35938e53bf7)

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04/22.04
 - Type of project: .NET 8
 - Stryker Version: 3.9.2

**Additional context**

To repro locally:
```bash
git clone https://github.com/Azure/dotnet-extensions-experimental.git mt
cd ./mt
git switch mutationtests
./restore.sh
./build.sh --mutationTest --configuration Release --bl
```

rouke-broersma:
I can reproduce this locally, looks like most errors are caused by missing or inaccessible Types. Eg `'Throw' is inaccessible due to its protection level`. Maybe a referenced assembly or an msbuild target is missing during stryker reference resolving.

dupdob:
there is another issue: there is no explicit package dependency to a testrunner in some of the test projects. As such, initial tests fail on windows so Stryker stop there. They work on Linux, probably because the xUnit.runner is part of the default adapters in the distribution.

RussKie:
> I can reproduce this locally, looks like most errors are caused by missing or inaccessible Types. Eg `'Throw' is inaccessible due to its protection level`. Maybe a referenced assembly or an msbuild target is missing during stryker reference resolving. Maybe `InternalsVisibleToTest` isn't working correctly for some reason in stryker on linux.

Throw.cs is conditionally referenced in a project if `InjectSharedThrow = true`:
https://github.com/Azure/dotnet-extensions-experimental/blob/b28a1decf1b59b223ba4612e9ec557c8e98af371/eng/MSBuild/Shared.props#L2-L4
`InjectSharedThrow` is set to `true` unless already set:
[...]

PR #2595, opened 2023-07-03
fix: Improve Vstest support

dupdob:
Complete revision of the VsTest error strategy.
At this time, this PR is feature complete but at least one unit test is missing. I open it for discussing if needed.
Overall changes:
1. Test assemblies are run in dedicated test session to workaround VsTest bug
2. If a test session fails to report the end notification in due time, Stryker assumes this is a timeout
3. If a test session reports TranslationErrorException,Stryker assumes the session had an internal error and retries it (up to three times)
4. if VsTest hangs (fails to stop properly), it is recycled. This happens independently from the test session result
5. fix MsTest detection (it was not working at all), resulting in inccorect timeouts for MsTest assemblies


src/Stryker.CLI/Stryker.CLI/runtimeconfig.template.json
```diff
@@ -1,3 +1,3 @@
 {
-  "rollForwardOnNoCandidateFx": 2
-}
\ No newline at end of file
+ // "rollForwardOnNoCandidateFx": 2
```
rouke-broersma, line 2: Removing this causes a 1-1 dependency on dotnet runtime version and a error on missing runtime even if a newer and probably compatible runtime is available.

dupdob: oh shit. Thanks for pointing that out. I had to comment it  when I tried VsStudio 17.7 preview to look into the issue #2585, as I had errors due to conflicting settings regarding framework preferences. 
BTW I am keeping the preview version and removed the regular one. It looks to me like it obsoletes several things in the build architecture. For now, I can no longer build NFluent solution via MsBuild (but not problem from VS itself)

dupdob: new update: it turns out the issue was simpler: Stryker does not detect preRelease version, and it is the only version I kept. So it fell back to MsBuild 4, which was too old to support the project.
So I added detection of prerelease version (if normal fails) and ensure that the msbuild fallback message is displayed (information instead of debug)

src/Stryker.Core/Stryker.Core/CoverageAnalysis/CoverageAnalyser.cs
```diff
@@ -72,17 +72,25 @@ private void ParseCoverage(IEnumerable<CoverageRunResult> coverage, IEnumerable<
...
-                CoverageForThisMutant(mutant, mutationToResultMap, allTestsExceptTrusted, new TestGuidsList( dubiousTests), failedTests);
+                CoverageForThisMutant(mutant, mutationToResultMap, allTest, allTestsExceptTrusted, new TestGuidsList( dubiousTests), failedTests);
             }
         }
 
         private void CoverageForThisMutant(Mutant mutant,
             IReadOnlyDictionary<int, List<CoverageRunResult>> mutationToResultMap,
+            TestGuidsList everytest,
             TestGuidsList allTestsGuidsExceptTrusted,
             TestGuidsList dubiousTests,
```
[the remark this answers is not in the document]
psfinaki: I mean I can imagine a lot of dubious tests but... :D

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/RunEventHandler.cs
```diff
@@ -203,6 +230,11 @@ public bool Wait(int timeOut)
                     Monitor.Wait(_lck, Math.Max(0, (int)(timeOut-watch.ElapsedMilliseconds)));
                 }
 
+                slept = watch.ElapsedMilliseconds - timeOut > 30 * 1000;
```
psfinaki, line 233: Hmm why 30 mins? I mean, this is a configurable thing in an operated system... Shouldn't this info be somehow extracted from the OS?

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
```diff
@@ -190,108 +193,132 @@ private TestRunResult BuildTestRunResult(IRunResults testResults, int expectedTe
...
+                if (!tests.IsEveryTest && testsForAssembly.Count == 0)
+                {
+                    // skip empty assemblies
+                    continue;
+                }
+
+                _logger.LogTrace($"{RunnerId}: testing assembly {source}.");
+                RunVsTest(tests, source, runSettings, options, timeOut, runEventHandler);
+
+                if (_cancelled)
```
psfinaki, line 244: VSTestRunner is still mostly dark matter for me - yet I wonder, have you considered bringing `CancellationToken` here for a more native handling of cancellations? That could probably simplify testing as well.

docs/technical-reference/VsTest internal.md
```diff
@@ -0,0 +1,41 @@
...
+
+
+### Timeout
+VsTest supports time out since V 15.5. That is, one can provide a maximum execution time for the test session, 
+and VsTest will automatically stops the test session if it is still running when the time is elapsed.
+
+#### Known issues (As of VsTest 17.6,)
+- Cancelled test sessions may never end: VsTest may not report the end of a cancelled test session when more than one testhost process was required.
+Note that this also applies to session cancelled by the timeout logic. When this happen, the endsession event is not raised, but the task does end.
+- Crashed test sessions (StackOverflow, OutOfMemory....) may appear as timeout. Note the the (VsTest) logs will reflect the crash, but the VsTest server
```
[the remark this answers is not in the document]
dupdob: definitely a lol type

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/RunEventHandler.cs
dupdob: this is 30 seconds. it looks reasonable, any sleep of more than 30 seconds will be detected and invalidate this test run

dupdob: if Stryker waited 30 seconds more than the time out, it means either that the machine was really really busy (unlikely) or went to sleep (likely), so the timeout triggered as soon as the computer woke up (timeout are wall clock based).
Then we assume this may be a spurious timeout and it deserves a new try. I think any value below a few seconds may be risky, as a significant swap or context switch in a VM, or a GC, can lead to a few seconds of freeze.
On the other hand, if the user is next to his/her computer, there is a good chance he/she acts if/when it sees it going to sleep, meaning the computer maybe asleep for a few seconds before waking up.
I think the lowest sensible value is 10 seconds, and I think anything above a minute is too long.
hard to pick a good value.
Of course, having a definitive way to detect sleep mode would be better.

src/Stryker.Core/Stryker.Core/CoverageAnalysis/CoverageAnalyser.cs
```diff
@@ -72,17 +72,25 @@ private void ParseCoverage(IEnumerable<CoverageRunResult> coverage, IEnumerable<
                 }
             }
 
+
+            var allTest = TestGuidsList.EveryTest();
             var allTestsExceptTrusted = (trustedTests.Count == 0 && failedTests.Count == 0)? TestGuidsList.EveryTest()
```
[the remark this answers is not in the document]
dupdob: a trusted test is a test for which Stryker has total certainty of which mutation it covers. This happens when the coverage was captured via a dedicated test session (i.e. a test session where it was the only test ran).

src/Stryker.Core/Stryker.Core/CoverageAnalysis/CoverageAnalyser.cs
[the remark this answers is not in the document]
dupdob: a dubious test is a test for which we fail to capture the coverage. it happens with xUnit and some theories...
Stryker must assume these tests could cover any mutations

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
dupdob: I guess this is poor naming on my part. `_cancelled` is true when Stryker bails the test session because it reaches a conclusion for every mutation to be tested (in this session). It is used here so that Stryker does not start a test session with the next assembly in the list. It saves time (and avoid a few nasty bugs linked to extra notifications).

src/Stryker.Core/Stryker.Core/CoverageAnalysis/CoverageAnalyser.cs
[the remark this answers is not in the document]
psfinaki: Thanks!

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/RunEventHandler.cs
psfinaki: > this is 30 seconds

Sorry, brain fart :D

> Of course, having a definitive way to detect sleep mode would be better.

I see, so that is a heuristic basically. Well, probably that's the best we can do. I mean looks like there is [something for that](https://stackoverflow.com/questions/16747146/c-sharp-sleep-detection-powermodechanged) but this doesn't seem to be cross-platform and yeah well accessing system APIs might lead to all sorts of access rights issues and so on.

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
psfinaki: Ohhh I see. Maybe `_testSessionCanceled` then?

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/RunEventHandler.cs
dupdob: took a look at the doc. The events also need a message pump active; I think my hack is better as: 
1. generally available
2. identify what matters (a suspicious delay)

src/Stryker.CLI/Stryker.CLI/Logging/LoggingInitializer.cs
```diff
@@ -29,7 +28,7 @@ public void SetupLogOptions(IStrykerInputs inputs, IFileSystem fileSystem = null
             var logLevel = inputs.VerbosityInput.Validate();
             var logToFile = inputs.LogToFileInput.Validate(outputPath);
 
-            ApplicationLogging.ConfigureLogger(logLevel, logToFile, outputPath);
+            ApplicationLogging.ConfigureLogger(logLevel, logToFile, inputs.DevModeInput.Validate(), outputPath);
```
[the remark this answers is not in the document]
dupdob: well, I did use the 'new' dev_mode while working on this PR; so, no short plan for removing it. Just made it a better fit for nowadays use, at least from where I stand

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
dupdob: went for `_currentSessionCancelled'

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/RunEventHandler.cs
psfinaki: Yep that's what I say, this is fine then.

src/Stryker.CLI/Stryker.CLI/Logging/ApplicationLogging.cs
```diff
@@ -10,17 +10,17 @@ namespace Stryker.CLI.Logging
 {
     public static class ApplicationLogging
     {
-        private static ILoggerFactory _factory = null;
+        private static ILoggerFactory factory;
 
-        public static void ConfigureLogger(LogEventLevel logLevel, bool logToFile, string outputPath)
+        public static void ConfigureLogger(LogEventLevel logLevel, bool toFile, bool traceToFile, string outputPath)
```
rouke-broersma, line 15:
```suggestion
        public static void ConfigureLogger(LogEventLevel logLevel, bool logToFile, bool traceToFile, string outputPath)
```

src/Stryker.CLI/Stryker.CLI/Logging/ApplicationLogging.cs
```diff
@@ -10,17 +10,17 @@ namespace Stryker.CLI.Logging
...
         {
             LoggerFactory.AddSerilog(new LoggerConfiguration().MinimumLevel.Is(logLevel).Enrich.FromLogContext().WriteTo.Console().CreateLogger());
 
-            if (logToFile)
+            if (toFile)
```
rouke-broersma, line 19:
```suggestion
            if (logToFile)
```

src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestContextInformation.cs
```diff
@@ -148,6 +148,7 @@ private ConsoleParameters DetermineConsoleParameters(string runnerId)
                 return determineConsoleParameters;
             }
 
+            determineConsoleParameters.TraceLevel = Options.DevMode ? TraceLevel.Verbose : TraceLevel.Info;
```
rouke-broersma, line 151: Does vstest expose user information in Verbose mode that is not exposed in Info mode? Or what is the reason for using devmode here?

dupdob: not sure what you mean by this question...
Verbose expose more information about what is happening inside VsTest, and this was needed for me to understand the problem with it (VsTest).
As I see things, we know have `log-to-file` for when we need a user to share a log and `dev mode` for when a Stryker contributor needs to understand what happens in details.

rouke-broersma: I understand the change for stryker's own logs but this log level is only for vstest, I'm asking what the purpose is for a difference between trace and debug in vstest logs specifically.

dupdob: for vstest logs, there is no debug level. This is information or verbose

dupdob: verbose provides information regarding the exchange between the VsTest server and the test host instances

PR #2608, opened 2023-07-25
chore: An integration test for F#

psfinaki:
- This fixes a hard crash happening currently for F# projects and adds an integration test to keep an eye on that.
- The integration test basically just verifies no crash and only runs on Ubuntu to save the pipeline resources.
- Also, a couple of small things

### DIFF
--- a/integrationtest/ValidationProject/ValidateStrykerResults.cs
+++ b/integrationtest/ValidationProject/ValidateStrykerResults.cs
@@ -20,7 +20,7 @@ namespace IntegrationTests
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
                 var directory = new DirectoryInfo("../../../../TargetProjects/NetFramework/FullFrameworkApp.Test/StrykerOutput");
-                directory.GetFiles("*.json").ShouldNotBeNull("No reports available to assert");
+                directory.GetFiles("*.json", SearchOption.AllDirectories).ShouldNotBeEmpty("No reports available to assert");
 
                 var latestReport = directory.GetFiles(MutationReportJson, SearchOption.AllDirectories)
                     .OrderByDescending(f => f.LastWriteTime)
@@ -39,7 +39,7 @@ namespace IntegrationTests
         public void NetCore()
         {
             var directory = new DirectoryInfo("../../../../TargetProjects/NetCoreTestProject.XUnit/StrykerOutput");
-            directory.GetFiles("*.json").ShouldNotBeNull("No reports available to assert");
+            directory.GetFiles("*.json", SearchOption.AllDirectories).ShouldNotBeEmpty("No reports available to assert");
 
             var latestReport = directory.GetFiles(MutationReportJson, SearchOption.AllDirectories)
                 .OrderByDescending(f => f.LastWriteTime)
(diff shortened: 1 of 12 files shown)


integrationtest/ValidationProject/ValidateStrykerResults.cs
```diff
@@ -20,7 +20,7 @@ public void NetFullFramework()
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
                 var directory = new DirectoryInfo("../../../../TargetProjects/NetFramework/FullFrameworkApp.Test/StrykerOutput");
-                directory.GetFiles("*.json").ShouldNotBeNull("No reports available to assert");
+                directory.GetFiles("*.json", SearchOption.AllDirectories).ShouldNotBeEmpty("No reports available to assert");
```
psfinaki, line 23: This construction was never working - GetFiles would always return empty here because:
a) it just [doesn't return null](https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.getfiles?view=net-7.0#system-io-directory-getfiles(system-string))
b) it doesn't search recursively by default whereas the reports are beyond another level of folders (timestamps)

src/Stryker.Core/Stryker.Core/Compiling/FsharpCompilingProcess.cs
```diff
@@ -63,7 +63,6 @@ public CompilingProcessResult Compile(IEnumerable<ParsedInput> syntaxTrees, bool
                     _fileSystem.Directory.CreateDirectory(injectionPath);
                 }
 
-                _fileSystem.File.Copy(mutatedAssemblyPath, injectionPath, true);
```
psfinaki, line 66: This was dramatically crashing with an IOException - "the project cannot access the file". Which probably makes sense, since at least for the `mutatedAssemblyPath` was the same as `injectionPath`. 

Regardless, just removing this line works :D That said, this should be revisited in the future to align the whole thing with the current C# flow.

integrationtest/ValidationProject/ValidateStrykerResults.cs
```diff
@@ -53,12 +53,32 @@ public void NetCore()
             CheckReportTestCounts(report, total: 14);
         }
 
+        [Fact]
+        [Trait("Category", "FSharp")]
```
psfinaki, line 57: I don't really know how to categorize things here, I hope eventually we will run this on whatever platform and probably in the solution mode. At that point, we might need to reassess the whole test categorization.

psfinaki:
Okay folks so I managed to reproduce it on WSL, indeed things break on Linux (and MacOS respectively). Most probably it's related to some wrong `FSharpChecker` configuration we have in `FSharpCompilingProcess` but I am running out of ideas now. Some possibly related stuff: [this](https://github.com/dotnet/fsharp/issues/14250), [this](https://github.com/fsharp/FsAutoComplete/issues/1006), [this](https://github.com/dotnet/fsharp/issues/14289).

I'd get back to that since I'll be definitely accessing `FSharpCompilingProcess` in the future work. Meanwhile, would you mind having the integration test running on Windows in CI? As of now, this adds 30 secs pipeline time.

psfinaki:
Hm now seems like something happened to the pipeline - artifacts are not published. This seems to be unrelated...

rouke-broersma:
Looks like myget is gone: https://github.com/MyGet/MyGetDocs/issues/143 Crap

src/Stryker.Core/Stryker.Core/Initialisation/ProjectMutator.cs
```diff
@@ -46,6 +47,12 @@ private void EnrichTestProjectsWithTestInfo(InitialTestRun initialTestRun, TestP
                 var testFile = testProjectsInfo.TestFiles.SingleOrDefault(testFile => testFile.FilePath == unitTest.CodeFilePath);
                 if (testFile is not null)
                 {
+                    // hard stub - F# has a different syntax tree and would throw further down the line
```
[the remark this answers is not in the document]
PierrickBlons: Had the same issue when dealing with FSharp test project.
Here is another suggestion to fix the issue by filtering directly on the initial list to use only csharp files where the foreach is applied.

```csharp
var unitTests = initialTestRun.Result.VsTestDescriptions
                .Select(desc => desc.Case)
                .Where(unitTest => unitTest.CodeFilePath != null && unitTest.CodeFilePath.EndsWith(".cs"));
foreach (var unitTest in unitTests)
{
    ...
```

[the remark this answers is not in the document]
psfinaki: Yep, that's a less ugly way to do that, thanks for the suggestion. :)

Issue #2621, opened 2023-08-02
Unhandled exception in Core.Initialisation.ProjectOrchestrator.MutateProjects


tommysor:
**Describe the bug**
Unhandled exception during initialization.
This error does _not_ happen when using Stryker 3.8.2 (completes successfully).
Stryker 3.9.0 and 3.10.0 fail
```
Unhandled exception. System.InvalidOperationException: Sequence contains no elements
   at System.Linq.ThrowHelper.ThrowNoElementsException()
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source)
   at Stryker.Core.Initialisation.ProjectOrchestrator.MutateProjects(StrykerOptions options, IReporter reporters, ITestRunner runner)+MoveNext()
```

**Logs**
Console output (shortened) for success (v3.8.2), then failure (v3.10.0)
```
@tommysor ➜ /workspaces/SimplifiedSearch (main) $ dotnet stryker --reporter Dots
[.. snip ..]
Version: 3.8.2

A new version of Stryker.NET (3.10.0) is available. Please consider upgrading using `dotnet tool update -g dotnet-stryker`

[20:55:04 INF] Identifying projects to mutate in /workspaces/SimplifiedSearch/SimplifiedSearch.sln. This can take a while.
[20:55:11 INF] Found 1 source projects
[20:55:11 INF] Found 2 test projects
[20:55:18 INF] Found project /workspaces/SimplifiedSearch/src/SimplifiedSearch/SimplifiedSearch.csproj to mutate.
[20:55:22 INF] Total number of tests found: 114.
[20:55:22 INF] Initial testrun started.
[20:55:32 INF] 180 mutants created
[20:55:32 INF] Capture mutant coverage using 'CoverageBasedTest' mode.
[.. snip ..]
[20:55:38 INF] 59    total mutants are skipped for the above mentioned reasons
[20:55:38 INF] 121   total mutants will be tested
........S...............S.....................................................S.............S............................
[21:00:17 INF] Time Elapsed 00:05:12.9085227
[21:00:17 INF] The final mutation score is 95.90 %

@tommysor ➜ /workspaces/SimplifiedSearch (main) $ dotnet tool restore
Tool 'dotnet-stryker' (version '3.10.0') was restored. Available commands: dotnet-stryker
Restore was successful.

@tommysor ➜ /workspaces/SimplifiedSearch (main) $ dotnet stryker --reporter Dots --log-to-file
[.. snip ..]
[...]

PR #2624, opened 2023-08-05
Improve handling for SourceProjectName parameter

dupdob:
As per issue #2621, Stryker.Net versions V3.9 and V.3.10 crash when provided with an incorrect `SourceProjectName` parameter. This version improves the situation by:
1) adding support for relative path for `SourceProjectName`
2) throwing an `InputException` explaining that no project matches the `SourceProjectName` option.


rouke-broersma:
I hadn't realized we now supported running in solution mode but with a project filter. I would actually vote to remove that functionality because unless I'm missing something it doesn't make sense. You gain nothing except slower mutation test times because the whole solution is analysed unnecessarily. Instead I would suggest that in this case we throw a warning that the project name filter is unused. @richardwerkman what do you think

dupdob:
Fixed the remaining bug highlighted by integration test: project name filtering logic was done sequentially, so the second filtering logic would obligatory fail.

dupdob:
@rouke-broersma : I am not saying that supporting projectname filtering brings a lot of value; but it makes sense in my book from a configuration UX point of view:
1. Every configuration entries may be present simultaneously and this should be a valid file
2. If two or more entries lead to conflicting decisions, those should be described via a single enumeration
3. If this is not achievable, a priority must be applied, leading to one or more configuration entry(ies) to be ignored. An explicit warning should be displayed at run time
4. more generally, if an entry is ignored as the result of some other configuration entry(ies), a warning should be displayed about this too.
Following these principles have several benefits:
1. Adding new configuration entries do not result in breaking change
2. Users can try changing any single configuration entry without having to deal with complex dependency rules
There are drawbacks as well:
1. This puts some extra work on development side as it usually happens when the program is liberal in what it accept; on the other hand, it drives the code into a design where parameters are decoupled, which help streamline the architecture.
2. This increase the number of possible states for the configuration files, increasing the risk of edge case bugs; but as stressed previously, this also drives to a more decoupled design that should help deal with those edge cases.

Regarding that specific situation (`SourceProjectName`), I think Stryker.Net is moving toward a situation where users may choose between an additive or subtractive configuration mode: 
- Additive is using the Single Project mode, specifying additional test projects
- Subtractive is using Solution Mode and filtering out the project(s) you are not interested with

Adding support for glob or regexp for this parameter would make this approach interesting.
Assuming users want/need that (I think they may do).

richardwerkman:
> I hadn't realized we now supported running in solution mode but with a project filter.

@rouke-broersma I also wasn't aware of this. And I agree that I don't see the added value of this. We might want to consider removing the "support" for filtering for now.

@dupdob I've been testing your PR and it works on the `SimplifiedSearch` project. However. For me, our `master` branch also works on the project. I've reverted the project to before the "fix", so it has the relative path set as "project". So basically, I was unable to reproduce the issue. Maybe I'm doing something wrong tho... For now I'm hesitant merging this PR as (from my perspective) this doesn't fix anything. 

> Subtractive is using Solution Mode and filtering out the project(s) you are not interested with
 Adding support for glob or regexp for this parameter would make this approach interesting.

Yes this is the long term plan! But I wouldn't reuse the project option for that. It would be a separate option, but we need to make a design for how this would work.

dupdob:
Well
I was able to to reproduce it locally consistently, not sure why it works for you. I was also able to reproduce it with the unit test I added (you can copy paste it if you feel inclined to). 

Just use some crappy solution project name and it should fail

dupdob:
temporary switching back to draft. In progress:
1) improve test coverage
2) refactor the projectorchestratortest class to simplify test writing ==> this should ease the creation of unit test in the future

dupdob:
@richardwerkman , @rouke-broersma 
I am Ok to remove support for project filtering if this bothers you, as I feel the redesign of the tests will be bring present and future value.
This means disregarding (and logging a warning) when SourceProjectName is provided and in solution mode ?
Also, should I remove support for relative path names (for SourceProjectName) which is added by this PR ? this was the only feature change

rouke-broersma:
> @richardwerkman , @rouke-broersma 
> I am Ok to remove support for project filtering if this bothers you, as I feel the redesign of the tests will be bring present and future value.
> This means disregarding (and logging a warning) when SourceProjectName is provided and in solution mode ?
> Also, should I remove support for relative path names (for SourceProjectName) which is added by this PR ? this was the only feature change

Yes I prefer if the support is removed in solution mode with a warning when configured. And support for path resolution should also be removed. Improved tests are much appreciated!

dupdob:
@rouke-broersma, @richardwerkman: I have a new question: it turns out that project filtering WAS working previously, so disabling it altogether modifies integration tests. What was not working was relative path support.
So, what shall I do:

1. adjust the integration expected results (to the new behavior)
2. restore filtering (with relative path support)
3. restore as is (i.e. no change to filtering logic in solution mode)

I fell the first option is the most logical one, but I ask because this change of behavior may impact users

dupdob:
ok,
I moved the most interesting part to PR #2719, I close this one. Adjusting existing behavior will be easy thanks to the added tests. So decision can happen any time.

PR #2628, opened 2023-08-09
chore: Lightening up some abstractions

psfinaki:
This is a non-functional change on the road to F# support. However, it should bring minor benefits to the code anyway, see my notes in the PR.

Now, to dramatically simplify what's going on: we need to get rid of the Roslyn (`Microsoft.CodeAnalysis`) dependencies where they don't belong. F# has its own compiler so its core doesn't have a notion of Roslyn.

There will be more similar PRs, I want to keep them small.


src/Stryker.Core/Stryker.Core/MutantFilters/IgnoreBlockMutantFilter.cs
```diff
@@ -30,7 +30,7 @@ public IEnumerable<Mutant> FilterMutants(IEnumerable<Mutant> mutants, IReadOnlyF
             var mutantsToIgnore = blockMutants.Where(blockMutant => mutants.Any(
                 m => m != blockMutant
                 && !_inactiveStatuses.Contains(m.ResultStatus)
-                && blockMutant.Span.Value.Contains(m.Span.Value)
+                && blockMutant.Mutation.OriginalNode.Span.Contains(m.Mutation.OriginalNode.Span)
```
psfinaki, line 33: As seen here, `Span` properties was more harmful than helpful - direct access of a nullable's `Value` is a bad practice.

src/Stryker.Core/Stryker.Core/Mutants/Mutant.cs
```diff
@@ -46,10 +43,6 @@ public class Mutant : IReadOnlyMutant
 
         public string DisplayName => $"{Id}: {Mutation?.DisplayName}";
 
-        public int? Line => Mutation?.OriginalNode?.GetLocation().GetLineSpan().StartLinePosition.Line + 1;
```
psfinaki, line 49: Ideologically, that's just making core abstractions free of things that are used in some very specific places (a mutant filter and a reporter).

src/Stryker.Core/Stryker.Core/Reporters/Json/Location.cs
```diff
@@ -1,27 +1,34 @@
-using Microsoft.CodeAnalysis;
-
 namespace Stryker.Core.Reporters.Json
 {
+    public struct LocationDimensions
```
psfinaki, line 3: I realize this is pretty much yet another wrapper of the numbers here - I believe eventually things will streamline here a bit so hopefully this is an intermediary step.

src/Stryker.Core/Stryker.Core/Mutants/Mutant.cs
rouke-broersma: However you've now made more consumers responsible for translating the Roslyn Span to user facing Span. If anything I think we should promote Json.Location upwards so we have a proper abstraction for all user facing access of the code span where the span should be 1 instead of 0 indexed. What do you think?

src/Stryker.Core/Stryker.Core/Reporters/Json/Location.cs
rouke-broersma: Couldn't we solve the same problem with overloading for the correct type to use?

src/Stryker.Core/Stryker.Core/Mutants/Mutant.cs
psfinaki: Well this might be what we want. F# has its own text span, from the FSharp.Compiler.Test assembly. Now, the difference is quite subtle, basically it's about counting things from 0 vs counting things from 1.

Therefore, we could either generify code based on this type or do some conversions here and there (this is doable). As of now, I am not yet sure which way will produce better code hence I am doing some steps that are supposed to be somewhat beneficial anyway.

src/Stryker.Core/Stryker.Core/Reporters/ClearTextTreeReporter.cs
```diff
@@ -82,7 +82,8 @@ public void OnAllMutantsTested(IReadOnlyProjectComponent reportComponent, TestPr
                         _ => $"[Red][[{mutant.ResultStatus}]][/]",
                     };
 
-                    var mutantNode = fileNode.AddNode(status + $" {mutant.Mutation.DisplayName} on line {mutant.Line}");
+                    var mutantLine = mutant.Mutation.OriginalNode.GetLocation().GetLineSpan().StartLinePosition.Line + 1;
```
richardwerkman, line 85: Isn't it better to make the Mutation class generic and solve this in there? Because this line is still C# specific. If we add a `GetLine()` method to `Mutation` we could just get the line here, both for C# and F#.

psfinaki: @richardwerkman this is the end goal. It's just damn hard to do right away :D
I mean, that's what I did back in a day but then most of the data structures had to become double-generic and like 50% of the code base changed. 

But eventually we'll get there and some things will streamline. I hope.

psfinaki: This whole thing feels like Jenga game - I am poking different pieces and taking whatever I can take without crashing the whole thing right away :D

richardwerkman: Yeah I know what you mean. We need to split stryker into smaller pieces to make changes easier! It's on our backlog ;)

src/Stryker.Core/Stryker.Core/Mutants/Mutant.cs
richardwerkman: We will probably need something like a `CSharpMutant` and `FSharpMutant` right? I'm not sure if the PR in its current form contributes towards this goal.

PR #2657, opened 2023-08-29
Enable filtering null-conditional method invocations by name

### LINKED ISSUE #2653: Null-conditional method calls are not filtered out by `ignore-methods` setting.
**Describe the bug**
Method calls using null-conditional invocation syntax are not filtered out by `ignore-methods` configuration setting.

**Expected behavior**
Method calls using null-conditional invocation syntax are filtered out by `ignore-methods` configuration like normal method invocations.

**Desktop:**
 - OS: Windows
 - Type of project: library
 - Framework Version: .net 7.0
 - Stryker Version: 3.10.0

**Additional context**
Probably because the Roslyn syntax tree looks quite different:
```cs
// regular invocation:
InvocationExpression(
    MemberAccessExpression(
        SyntaxKind.SimpleMemberAccessExpression,
        IdentifierName("value"),
        IdentifierName("ToString")))))}))

// null-conditional invocation:
ConditionalAccessExpression(
    IdentifierName("value"),
    InvocationExpression(
        MemberBindingExpression(
            IdentifierName("ToString"))))))}))
```

pentp:
Fixes #2653

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/IgnoredMethodMutantFilterTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/IgnoredMethodMutantFilterTests.cs
@@ -83,6 +83,48 @@ public class IgnoredMethodMutantFilter_NestedMethodCalls
         }
 
         [Theory]
+        [InlineData("Range", false)]
+        [InlineData("Where", false)]
+        [InlineData("ToList", true)]
+        [InlineData("", false)]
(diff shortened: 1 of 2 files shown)


dupdob:
thanks for the PR. It clarifies the issue: the problem is specific to the statement mutator, and the fact that existing unit test are no longer adapted.

richardwerkman:
Thanks for the contribution! This seems like the perfect fix for the issue. I'll merge and this should be a part of the next version of Stryker.

PR #2661, opened 2023-08-29
chore: add suggestion for file-scoped namespaces to editorconfig

### LINKED ISSUE #2470: Refactor all files to use file-scoped namespaces

ysbakker:
- Refactor entire solution to file-scoped namespaces
- Bump language version of data collector project and also refactor to file-scoped namespaces

Relates to #2470, will be fixed in a later change.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -107,6 +107,8 @@ csharp_style_var_for_built_in_types = true:suggestion
 csharp_style_var_when_type_is_apparent = true:warning
 csharp_style_var_elsewhere = true:suggestion
 
+csharp_style_namespace_declarations = file_scoped:suggestion
+
 # Default naming styles
 dotnet_naming_rule.async_members_should_be_pascal_case_suffixed_with_async.severity = suggestion
 dotnet_naming_rule.async_members_should_be_pascal_case_suffixed_with_async.style = pascal_case_suffixed_with_async


ysbakker:
Any git experts know if this messes up the git blame? It doesn't appear to in my IDE, but if it does I'd rather not merge this and take credit for the *entire Stryker project* with a refactoring action that literally took 5 minutes.

ysbakker:
Also enforcing this in the editorconfig might be a good idea:
https://stackoverflow.com/questions/69486362/editorconfig-control-file-scoped-namespace-declaration

This rule doesn't appear to work in Rider though, can someone check if visual studio supports this?
`csharp_style_namespace_declarations = file_scoped:warning`

pentp:
> Any git experts know if this messes up the git blame? It doesn't appear to in my IDE, but if it does I'd rather not merge this and take credit for the _entire Stryker project_ with a refactoring action that literally took 5 minutes.

From git perspective this counts as an edit for basically all lines in the solution. It's possible to mitigated this by adding the commit to [.git-blame-ignore-revs](https://stackoverflow.com/questions/70435937/git-blame-with-optional-ignorerevsfile), but this needs manual setup for each user and isn't perfect.

IMHO the benefits of the newer namespace style are not worth the massive change to all existing files. Enforcing it for new files makes sense.

ysbakker:
> > Any git experts know if this messes up the git blame? It doesn't appear to in my IDE, but if it does I'd rather not merge this and take credit for the _entire Stryker project_ with a refactoring action that literally took 5 minutes.
> 
> From git perspective this counts as an edit for basically all lines in the solution. It's possible to mitigated this by adding the commit to [.git-blame-ignore-revs](https://stackoverflow.com/questions/70435937/git-blame-with-optional-ignorerevsfile), but this needs manual setup for each user and isn't perfect.
> 
> IMHO the benefits of the newer namespace style are not worth the massive change to all existing files. Enforcing it for new files makes sense.

Agreed.

Dug into the editorconfig rule, and it seems like Rider doesn't support some rules (https://youtrack.jetbrains.com/issue/RIDER-51400/Support-roslyn-properties-in-editorconfig). So I could add `csharp_style_namespace_declarations = file_scoped:suggestion`, but this will likely only work in visual studio. I'll revert my other changes and await feedback on this.

rouke-broersma:
> 

I would appreciate the editorconfig change!

ysbakker:
> @ysbakker we discussed the change.
> 
> We would like to simply get this over with, so we propose that you do perform the refactor, but perhaps not when we have 14 open pull requests. Are you going to join hacktoberfest event this year? If so we can try to get all existing PRs merged and do the refactor early in the day of the event. We should then also add the .git-blame-ignore-revs file so at least on github the blame is usable. There's your first two PRs already done ;)
> 
> This means that for now we would like the editorconfig change. After the big bang refactor we should also make the rule enforced instead of suggested.

Yes, I will likely be joining this year. Sounds like a plan!

PR #2663, opened 2023-08-30
chore: Add mutation score badge to readme

richardwerkman:
This was long overdue


README.md
```diff
@@ -1,3 +1,4 @@
+[![Mutation testing badge](https://img.shields.io/endpoint?style=flat&url=https%3A%2F%2Fbadge-api.stryker-mutator.io%2Fgithub.com%2Fstryker-mutator%2Fstryker-net%2Fmaster)](https://dashboard.stryker-mutator.io/reports/github.com/stryker-mutator/stryker-net/master)
```
rouke-broersma, line 1: I would prefer to see it after the unit test badge, at least after the nuget info.

richardwerkman: Maybe that's better yes. I just followed stryker-js as an example

PR #2645, opened 2023-08-18
Decoupling file excludability from mutant spans

psfinaki:
Phew... Okay so this is a somewhat bigger steps towards the F# support. 

**Motivation**

F# has its own understanding of text documents (btw that's sad). Roslyn (in this context C#) generally counts all characters from zero whereas F# generally uses lines and columns. Of course one is convertible to the other, it's more about what's native to the compiler. Type-wise, it's `Microsoft.CodeAnalysis.Text.TextSpan` versus `FSharp.Compiler.Text.Range`.

Sooner or later we would need native span arithmetics for mutants hence one approach is to create a definition of mutant span on the user level (= configuration level) and then translate them to compiler definitions of spans based on the file type.

**Implementation**

This PR
- Makes `FilePattern` language-agnostic
- Splits the concept of excludability to its own place since that concept has some usage on its own
- Adds some span (range) arithmetics for F#
- Hopefully brings some collateral improvements

**TO DO**
- tests
- basic parameter validation for new types
- docs
- see comments in the PR

PR #2681, opened 2023-09-23
chore: Breaking down FilePattern

psfinaki:
Part of [this PR](https://github.com/stryker-mutator/stryker-net/pull/2645).

Consider this file pattern: `!MyFolder/MyService.cs{10..100}`. It has three parts:
- ! mark, indicating file exclusion
- file path (`MyFolder/MyService.cs`)
- mutant spans notation (`{10..100}`)

The `FilePattern` currently handles this by immediately transforming the string into file path Glob and Roslyn-specific `TextSpan`s.

The PR changes this by adding an intermediary phase of this transformation. It adds the `ExcludableString` class which has a notion of three parts of the string but doesn't convert the mutant spans part to the `TextSpan`s yet.

This has two following advantages:
- We are going to handle the mutant spans part differently for C# and F#
- We don't need Roslyn-specific `TextSpan`s for the `DiffIgnoreChangesInput`, there we only look at the exclusion and glob

Along the way, this simplifies a few tests and lightens up the `FilePattern` class.


src/Stryker.Core/Stryker.Core/ExcludableString.cs
```diff
@@ -0,0 +1,43 @@
...
+                Glob = Glob.Parse(normalized);
+                MutantSpans = string.Empty;
+            }
+        }
+
+        public bool IsExcluded { get; }
+
+        public Glob Glob { get; }
+
+        public string MutantSpans { get; }
```
rouke-broersma, line 41: I feel like MutantSpans might give the wrong impression. Also it's one span right, so shouldn't be plural? Or am I missing something?

src/Stryker.Core/Stryker.Core/ExcludableString.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Text.RegularExpressions;
+using DotNet.Globbing;
+
+namespace Stryker.Core
+{
+    public readonly struct ExcludableString
+    {
+        private static readonly Regex _mutantSpansRegex = new("(\\{(\\d+)\\.\\.(\\d+)\\})+$");
+
+        public ExcludableString(string s)
```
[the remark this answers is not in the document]
rouke-broersma: That might still be confusing. Can't we move all the parse logic into this class and call it something like ExclusionPatternParser or something? Currently part of the parsin is moved here but some remains in FilePattern.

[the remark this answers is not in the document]
psfinaki: Yeah, `ExclusionPattern` is better. I didn't like the name but didn't want to say this loud to not bias you :D 

Now, as for moving the other part of the parsing logic here - that would kind of defeat the purpose. Because we want to parse this thing - `{10..20}{20..30}` - differently for F# and C#. Different parsing techniques and different resulting "hard types" at the end (`TextSpan` for C# and `Range` for F#).

src/Stryker.Core/Stryker.Core/ExcludableString.cs
psfinaki: Regex is noone's cup of tea (maybe apart from Stephen Toub). 

So, there might be several spans, e.g. `{10..19}{20..30}`. [Here in the tests](https://github.com/stryker-mutator/stryker-net/blob/03f991307924b67d14791b7ee42d1f0f1c2c8717/src/Stryker.Core/Stryker.Core.UnitTest/Options/FilePatternTests.cs#L37-L44), you try all sorts of stuff :D And then there is a bunch of logic to reduce, split and intersect the spans.

Now, of course my work in terms of F# context would also simplify if we decide to just remove this. But just from the user perspective, Stryker comments are much more intuitive and powerful, indeed I cannot see a scenario when I would prefer specifying spans over Stryker comments.

psfinaki: So it goes down to what you decide and when. If you plan to remove this in the near future, this would kind of eliminate the need of this change. Like, the reasoning I have in the PR description will not apply anymore, FilePattern will reduce to ExclusionPattern, would be used to for the both relevant inputs, and then - eventually - C# files will just assume default full-file `TextSpan`s and F# files will assume default full-file `Range`s.

psfinaki: Actually, that would simplify so much stuff than can even volunteer to do that, since I have already spent a lot of time working in this area. But I need your green light :D

src/Stryker.Core/Stryker.Core/ExcludableString.cs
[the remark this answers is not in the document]
rouke-broersma: I meant up until but not including the hard types. The first part is still just string matching right?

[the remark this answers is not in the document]
psfinaki: Ohh right, yeah, that we can do I think.

src/Stryker.Core/Stryker.Core/ExcludableString.cs
rouke-broersma: We need to discuss this first but I'm all for it so I think it's likely we'll do this. The only issue is that since we're removing the feature it's a breaking change and should be in the next major version. I created the issue for it and I'l try to discuss this with @richardwerkman shortly but otherwise we can talk it over during hacktoberfest :)

psfinaki: Alright! Well in that case this should be ready for review then and can be taken as a step to the option removal in the future.

rouke-broersma: We'll do this later.

Issue #931, opened 2020-02-06
Failed to build the mutated assembly due to unrecoverable error: error CS0518: Predefined type 'System.Attribute' is not defined or imported


NanoRossi:
**Describe the bug**
The above error occurs when running dotnet stryker with one of my test projects,

**Logs**

[10:50:30 INF] Started initial build using dotnet build
[10:50:41 INF] Initial build successful
[10:51:03 INF] Using testrunner VsTest
[10:51:03 INF] Total number of tests found: 57
[10:51:03 INF] Initial testrun started
[10:51:06 INF] Using 9219 ms as testrun timeout
[10:51:11 ERR] Failed to build the mutated assembly due to unrecoverable error: error CS0518: Predefined type 'System.Attribute' is not defined or imported
[10:51:11 ERR] An error occurred during the mutation test run

General Build Failure detected.

[10:51:11 INF] Time Elapsed 00:00:50.3424653

Mobrockers:
Could you add the additional info asked in the bug template? Like what dotnet version you're using and what stryker version you're using?

Mobrockers:
I'm gonna go ahead and close this until you provide the above information. Feel free to reply and I'll re-open the issue.

Rjae:
Stryker: 0.16.1 (beta)
.NET Core: 3.1.101

I get this for previously successful builds.  For example...
Succeeded: https://github.com/appshapes-org/logging-console/actions/runs/84933566
Failed: https://github.com/appshapes-org/logging-console/actions/runs/84981650

Mobrockers:
@Rjae Awesome, thanks!

So does this happen intermittently at random or is there any kind of pattern you can detect?
Does this only happen in the build or also on your own systems?

Rjae:
It does happen locally as well. I have not tried to dig in to why yet, but I will try a version bump this morning and report results here. Thanks!

Rjae:
@Mobrockers the version upgrade seems to have resolved the issue. Thanks!

Mobrockers:
Cool! If you happen to notice that it's still failing sometimes please let us know

MaurizioPz:
I'm having the same problem on Stryker 0.18.0
The project is an NUnit test project and the main one is .NET Framework v4.6.1 (Windows)

These are the logs

[15:00:53 INF] Identifying project to mutate.
[15:01:04 INF] The project *** will be mutated.
[15:01:14 INF] Analysis complete.
[...]

PR #2683, opened 2023-09-26
Fix issue for Buildalyzer couldn't fin sourcefiles

tnoiretisa:
For specific solution, when csproj is not "standalone" buildable without solution file, we need to get to Buildalyzers the solution file path

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/TargetFrameworkResolutionTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/TargetFrameworkResolutionTests.cs
@@ -25,7 +25,10 @@ namespace Stryker.Core.UnitTest.Initialisation
             analyzerManagerMock
                 .Setup(m => m.GetProject(It.IsAny<string>()))
                 .Returns(projectAnalyzerMock.Object);
-
+            analyzerManagerMock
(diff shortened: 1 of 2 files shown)


dupdob:
hi
I am not sure I understand what problem this PR is supposed to fix. Is there a related issue ? If not, could you please create one, so we can better track the problem

tnoiretisa:
> hi I am not sure I understand what problem this PR is supposed to fix. Is there a related issue ? If not, could you please create one, so we can better track the problem

Hi, i'm try to add more details on this new issue https://github.com/stryker-mutator/stryker-net/issues/2685

JackSteel97:
@tnoiretisa Could this also be the fix for #931?

tnoiretisa:
> @tnoiretisa Could this also be the fix for #931?

Logs on #931 don't throw warn on buildalyzer, so i don't know if bugs have the same origin :/

rouke-broersma:
@tnoiretisa could you test #2690?

It's more or less your change but refactored to always use the solution file when it's available

tnoiretisa:
@rouke-broersma it's works 🙂

rouke-broersma:
> @rouke-broersma it's works 🙂

thanks for confirming and thanks for finding this issue.

PR #2689, opened 2023-09-28
Adjust IgnoredMethod filter to more consistent results

dupdob:
Adjust ignored method mutant filter so that it matches #2669


src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/IgnoredMethodMutantFilterTests.cs
```diff
@@ -14,13 +16,57 @@ namespace Stryker.Core.UnitTest.MutantFilters
...
         }
 
+        // filter should support documented cases
+        [Theory]
+        [InlineData("IgnoredMethod(true);", "true", true)]
+        [InlineData("x = IgnoredMethod(true);", null, true)]
+        [InlineData("var x = IgnoredMethod(true);", null, true)]
+        [InlineData("while (x == IgnoredMethod(true));", "==", false)]
+        [InlineData("IgnoredMethod()++;", "IgnoredMethod()++", false)]
+        [InlineData("x==1 ? IgnoredMethod(true): IgnoredMethod(false);", null, true)]
```
richardwerkman, line 35: What is this actually testing? Which mutation is ignored? The equality mutation or the boolean mutation?

richardwerkman: The equality mutation should not be ignored. The boolean mutations should be ignored.

src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/IgnoredMethodMutantFilterTests.cs
```diff
@@ -14,13 +16,57 @@ namespace Stryker.Core.UnitTest.MutantFilters
...
+        [InlineData("SomeMethod(true).IgnoredMethod(false);", "true", true)]
```
richardwerkman, line 36: I don't think this should be ignored. The true isn't part of the ignored method invocation

richardwerkman (changes requested):
I think this is a bit different from the expected behavior.

src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/IgnoredMethodMutantFilterTests.cs
dupdob: There is a type here. It should read: `[InlineData("x==1 ? IgnoredMethod(true): IgnoredMethod(false);", "==", true)]`.
I.e. the equality is mutated.
As said in the discussion: ternary expression mutation should be ignored if true and false expression (mutations) are to be ignored. If `IgnoredMethod` is somme logging method, this result in a unkillable mutant.
One could also note that 
`x!=1 ? IgnoredMethod(true): IgnoredMethod(false);` is semantically equivalent to `x==1 ? IgnoredMethod(false): IgnoredMethod(true);`. And the second one should be ignored because they alter the parameter. And yes, this is not a an actual mutation since I changed the parameters to both invocations, but I do not think this is important for the demonstration.

The underlying principle is that: if a method's mutation is to be ignored, one can assume that its (direct or side) effects cannot be tested, so all invocations are assumed to semantically equivalent.

May I suggest we continue discussing this in the discussion in #2669

src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/IgnoredMethodMutantFilterTests.cs
dupdob: Actually, the return value of `SomeMethod` is part of the invocation. What about this example:
`GetLogger(logAsync: true).LogMessage("logtext");` (named parameter for clarity). It will be hard to have discrimating test for this mutation: `GetLogger(logAsync: false).LogMessage("logtext");`

Of course, it does not mean in any way that `GetLogger` invocations are systematically ignored. Just when use as a chain invocation of `LogMessage`

src/Stryker.Core/Stryker.Core/MutantFilters/IgnoredMethodMutantFilter.cs
```diff
@@ -28,18 +29,25 @@ public sealed class IgnoredMethodMutantFilter : IMutantFilter
...
-                ExpressionStatementSyntax invocationExpressionStatementSyntax => IsPartOfIgnoredMethodCall(invocationExpressionStatementSyntax.Expression, options, false),
+                ConditionalExpressionSyntax conditionalExpression => IsPartOfIgnoredMethodCall(conditionalExpression.WhenTrue, options, false) && IsPartOfIgnoredMethodCall(conditionalExpression.WhenFalse, options, false),
+
+                ExpressionStatementSyntax expressionStatement => IsPartOfIgnoredMethodCall(expressionStatement.Expression, options, false),
+
+                AssignmentExpressionSyntax assignmentExpression =>  IsPartOfIgnoredMethodCall(assignmentExpression.Right, options, false),
+
+                LocalDeclarationStatementSyntax localDeclaration => localDeclaration.Declaration.Variables.All(v => IsPartOfIgnoredMethodCall(v.Initializer?.Value, options, false)),
+  
+                BlockSyntax {Statements.Count: >0 } block => block.Statements.All(s=> IsPartOfIgnoredMethodCall(s, options, false)),
```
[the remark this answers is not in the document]
Liam-Rougoor: Why is this one resolved?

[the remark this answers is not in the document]
dupdob: I usually commit those kind of change requests... not sure what happened. It was the same for all other minor requests. I guess I tried to batch changes or something and it failed or I forget to go through
thanks for pointing this

src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/IgnoredMethodMutantFilterTests.cs
richardwerkman: But we also mutate properties and other external values that could be used in `IgnoredMethod` right? I don't see that as part of the invocation of the ignored method

dupdob: It feels like you are over generalizing here. What this PR does is ignoring any mutations that solely impact the invocation of an ignored method in as an attempt to provide a more consistent and explainable experience.
It addresses known limitations with current logic, such as the ternary operator mutation `x==1 ? IgnoredMethod(true): IgnoredMethod(false);` while not addressing every one of them, such as :
```
if (loggingEnabled)
   Log(..);
```
where Stryker still mutations the `if` condition while it should not.

dupdob:
On a general note
I stand by my proposal for the unit test cases I present here, but I am ok to alter them to match your expectations. The idea is to get some flexibility in this filter to better match what people are doing with it.

I do feel the OoB experience of Stryker is lacking on the _ignore_ front. Stryker should by default ignore some usual suspects such as logging, tracing, assertions and async control and do it in a way that makes sense to users.

richardwerkman (approved):

src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/IgnoredMethodMutantFilterTests.cs
richardwerkman: I see your point of view. I'm fine with merging this now. If we see issues for this in the future we can always improve this further.

PR #2699, opened 2023-10-04
chore: Improve analyzer loading related error logging + minor cleanup …

dupdob:
Improve logging when Stryker fails to load a generator/Analyzer for build.


dupdob:
Can't improve code coverage, as it is code executed by Roslyn logic. Not sure there is a proper way to mock this, nor if this would have any value

Issue #1196, opened 2020-09-09
Improve integration tests to check mutation spans


richardwerkman:
**Is your feature request related to a problem? Please describe.**
We have seen problems with the spans of mutations after a release. 

**Describe the solution you'd like**
We would like to improve the integration tests to also check the spans of the mutations. Currently we only check if the number of mutations is correct.

We could do this by looping through all mutations an check if they are logically placed. For example, a span inside a using statement is unexpected as we don't mutate usings. The same goes for xml comments and outside of methods or fields. If the span of one or more mutations is unexpected the integration test should fail.  That way we know before we release stryker.

**Describe alternatives you've considered**
We can also check the exact span for each mutation. But this would be a lot of work and with each change the integration test would fail.

dupdob:
Note that we have unit tests for span verification. 
It may be easier to add more test cases to these. 
Envoyé de mon iPhone

richardwerkman:
Yes we do have unit tests. But the issue here is that the span can get lost in a lot of places. It could go wrong in any of the orchestrators. Or a mutator itself. Or mutationcontext. By using the integration test we are sure the spans are correct in all of our code.

PR #2714, opened 2023-10-13
`Feat` | Blacklist for mutation span kinds

Andreas02-dev:
Check whether the location of the mutation is in a `Using` or in a `Comment`, if so then the mutation location will not be shown correctly and we should fail the run.

Related to #1196


integrationtest/ValidationProject/ValidateStrykerResults.cs
```diff
@@ -11,6 +15,25 @@ namespace IntegrationTests
 {
     public class ValidateStrykerResults
     {
+        private readonly ReadOnlyCollection<SyntaxKind> _blacklistedSyntaxKindsForMutating =
```
richardwerkman, line 18: Please add whitelisting for nodes inside of a MethodDeclarationSyntax or PropertyDeclarationSyntax

Andreas02-dev: Will do, we should also add SyntaxKind.ConstructorDeclaration and SyntaxKind.FieldDeclaration to the whitelist.

integrationtest/ValidationProject/ValidateStrykerResults.cs
```diff
@@ -107,10 +139,31 @@ public void SolutionRun()
 
             var report = JsonConvert.DeserializeObject<JsonReport>(strykerRunOutput);
 
+            CheckMutationKindsValidity(report);
```
richardwerkman, line 142: Make this method part of the CheckReportMutantCounts and rename it to CheckReportMutants

Andreas02-dev:
@richardwerkman Is there a reason why `CheckReportTestCounts` isn't called in the `NetFullFramework` test?

richardwerkman:
Congrats on your first contribution! :)

PR #2723, opened 2023-10-14
feat(baseline): Validate mutual exclusivity of explicit --since and --with-baseline options

### LINKED ISSUE #2721: Validate since and with-baseline mutually exclusive
At the moment, it is possible to use the --since and --with-baseline flags at the same time, but we want this to be mutually exclusive.

We want to add the following validation:

- If --with-baseline is passed, --since may not be passed.

For both of these scenarios, we want to throw an `InputException`.

 In the near future, we want to make these options into explicit commands. See: #2124,  #2554. This way, the user needs to explicitly choose which one to use. But this requires some work to be done to the cli/command setup first.

danihengeveld:
Closes #2721

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -482,7 +482,7 @@ The report name is based on the current branch name or the [project-info.version
 Set the diffing target on the command line by passing a committish with the since flag.
 Set the diffing target in the config file by setting the [since target](#sincetarget-committish) option.
 
-*\* The with-baseline and since options are mutually exclusive. This feature implicitly enables the [since](#since-flag-committish) feature.*
+*\* The baseline and since features are mutually exclusive. This feature implicitly enables the [since](#since-flag-committish) feature for now.*
 
 ### `baseline.enabled` <`flag`>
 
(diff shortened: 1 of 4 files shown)


docs/configuration.md
```diff
@@ -482,7 +482,7 @@ The report name is based on the current branch name or the [project-info.version
 Set the diffing target on the command line by passing a committish with the since flag.
 Set the diffing target in the config file by setting the [since target](#sincetarget-committish) option.
 
-*\* This feature automatically enables the [since](#since-flag-committish) feature.*
+*\* The with-baseline and since options are mutually exclusive. This feature implicitly enables the [since](#since-flag-committish) feature.*
```
rouke-broersma, line 485:
```suggestion
*\* The baseline and since features are mutually exclusive. This feature implicitly enables the [since](#since-flag-committish) feature for now.*
```

src/Stryker.Core/Stryker.Core/Options/Inputs/SinceInput.cs
```diff
@@ -10,6 +12,11 @@ public bool Validate(bool? withBaseline)
         {
             if (withBaseline.IsNotNullAndTrue())
             {
+                if (SuppliedInput.HasValue)
+                {
+                    throw new InputException("The since and with-baseline features are mutually exclusive.");
```
rouke-broersma, line 17:
```suggestion
                    throw new InputException("The since and baseline features are mutually exclusive.");
```

PR #2724, opened 2023-10-14
feat(core): add implicit array init mutation

### LINKED ISSUE #1024: Feature request: Mutate implicit array initializer
When mutating an implicit array initializer it results in a compile error.

``` cs
var myArray = new [] { 1, 2 };
```
Mutated to:
``` cs
var myArray = new [] { }; //error
```
We could mutate it if we change the array to a typed array:

``` cs
var myArray = new int[] { };
```
For this to work we have to retrace the type from the initialization values. We always have to be sure what Type the array should be or it results in other errors like:

``` cs
double[] myArray = new int[] { }; // error
```

We could also choose to not mutate implicit arrays but these mutations seem valuable to me.

coffeeneer:
This resolves #1024 and adds mutations for implicit array initialization.


coffeeneer:
This would be great if default actually would be empty array, but default of int[] is null.

image

PR #2729, opened 2023-10-14
fix(input validation): Only validate baseline configuration when baseline is enabled

### LINKED ISSUE #2675: Baseline configuration should not be validated when baseline is disabled
**Describe the bug**
When baseline is disabled but you have configuration such as set the baseline provider to azurefileshare stryker fails to run due to a validation error on the azure fileshare sas. However when baseline is disabled the sas should not be validated since it won't be used.

BasdeBruynInfoSupport:
fix #2675


danihengeveld:
Maybe add some tests for `StrykerInputs` in the `StrykerInputsTests`?

PR #2730, opened 2023-10-14
chore: Update FSharp.Compiler.Service

psfinaki:
Still fighting with the latest version, but this is at least a major update.
Contains a minor fix to the pdb path detection, it was always wrong but earlier just ignored by the F# compiler.


src/Directory.Packages.props
```diff
@@ -8,7 +8,7 @@
     <PackageVersion Include="YamlDotNet" Version="13.7.0" />
     <PackageVersion Include="Grynwald.MarkdownGenerator" Version="3.0.106" />
     <PackageVersion Include="LibGit2Sharp" Version="0.27.2" />
-    <PackageVersion Include="FSharp.Compiler.Service" Version="41.0.7" />
+    <PackageVersion Include="FSharp.Compiler.Service" Version="42.7.100-preview.22427.1" />
```
rouke-broersma, line 11: Just checking it's on purpose that we're updating to a preview package?

PR #2731, opened 2023-10-14
chore: add semantic model to mutators

### LINKED ISSUE #2720: [SPIKE] Investigate the cost of resolving semantic models for all syntax trees.
So far we have denied the need for semantic models as we assumed the cost of resolving the semantic models outweighs the value. However we are not sure what the exact cost is for the resolution.

I'd like to investigate what the exact cost is for resolving the semantic models on a big solution. If the cost is low (a few milliseconds) we can implement this and use the semantic models in our mutators. This would greatly improve our capabilities as we are now often blocked by the unknown types in a mutator.

A good starting point could be the [visual studio performance profiler](https://learn.microsoft.com/en-us/visualstudio/profiling/beginners-guide-to-performance-profiling?view=vs-2022) or some other tool to record performance.

richardwerkman:
fixes #2720


src/Stryker.Core/Stryker.Core/MutationTest/FsharpMutationProcess.cs
```diff
@@ -18,7 +19,7 @@ public class FsharpMutationProcess : IMutationProcess
         private readonly ILogger _logger;
         private readonly IFileSystem _fileSystem;
         private readonly StrykerOptions _options;
-        private readonly BaseMutantOrchestrator<FSharpList<SynModuleOrNamespace>> _orchestrator;
+        private readonly BaseMutantOrchestrator<FSharpList<SynModuleOrNamespace>, SemanticModel> _orchestrator;
```
rouke-broersma, line 22: This is a Roslyn type and does not work for fsharp.

src/Stryker.Core/Stryker.Core/MutationTest/IMutationProcess.cs
```diff
@@ -1,3 +1,5 @@
+using Microsoft.CodeAnalysis;
```
rouke-broersma, line 1: Interface should be language agnostic

src/Stryker.Core/Stryker.Core/Mutants/FsharpMutantOrchestrator.cs
```diff
@@ -11,7 +12,7 @@
 namespace Stryker.Core.Mutants
 {
     /// <inheritdoc/>
-    public class FsharpMutantOrchestrator : BaseMutantOrchestrator<FSharpList<SynModuleOrNamespace>>
+    public class FsharpMutantOrchestrator : BaseMutantOrchestrator<FSharpList<SynModuleOrNamespace>, SemanticModel>
```
rouke-broersma, line 15: We shouldn't introduce Roslyn types to the fsharp orchestrator

richardwerkman: I've switched to object as a placeholder type (as fsharp doesn't yet use the semantic model)

Issue #2748, opened 2023-11-03
In solution mode, the since filter does not correctly detect test projects


MischaVreeburg:
**Describe the bug**
When running the since filter on a multi project solution, in solution mode, the unit test projects are not correctly identified. This results in a filter run with incorrect results.

**Expected behavior**
The unit test projects are correctly identified

**Desktop (please complete the following information):**
 - OS: window
 - .NET Framework 48
 - Framework Version [e.g. core 2.2]
 - Stryker Version 3.10

PR #2749, opened 2023-11-03
Correctly add test projects to strykerOptions TestProjects in solution mode for use in filters

MischaVreeburg:
The Since filter bases the testproject detection on the strykerOptions TestProjects property. This is currently not being filled in solution mode.

This PR fixes this issue and addresses issue: https://github.com/stryker-mutator/stryker-net/issues/2748


rouke-broersma:
Thank you for contributing a potential fix for this problem. This has been superseded by #2928

Issue #2750, opened 2023-11-03
Stryker does not enrich test project with test info for C#  .NET Framework unit tests


MischaVreeburg:
**Describe the bug**
Stryker is unable to correctly enrich test project data with test info data from VsTest for C# .NET Framework, in order to report to the user which test killed which mutant.

**Expected behavior**
Stryker correctly enriches test project data with test info data for C# .NET Framework, in order to report to the user which test killed which mutant.

**Desktop (please complete the following information):**
 - OS: [e.g. Windows]
 - Type of project [framework, core, xamarin]
 - Framework Version [e.g. core 2.2]
 - Stryker Version [e.g. beta 0.10.0]

**Additional context**

dupdob:
this is a known issue with vstest, I am pretty sure there is already an issue for this but I can't find it right now.

rouke-broersma:
Duplicate of #2458

PR #2751, opened 2023-11-03
Fix the problem that the test projects are not being enriched by VsTest.

MischaVreeburg:
For the enrichment, an alternative datasource has been chosen along with some clever parsing. It now falls back for C# files to the previously scanned test source files and tries to map the test full name with the name in the file and uses that info to get the test file path, along with the method declaration.

This PR fixes issue: https://github.com/stryker-mutator/stryker-net/issues/2750

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Initialisation/ProjectMutator.cs
+++ b/src/Stryker.Core/Stryker.Core/Initialisation/ProjectMutator.cs
@@ -1,3 +1,4 @@
+using System;
 using System.IO;
 using System.Linq;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
@@ -30,7 +31,7 @@ namespace Stryker.Core.Initialisation
         {
             var process = _injectedMutationTestProcess ?? new MutationTestProcess(input, options, reporters,
                 new MutationTestExecutor(input.TestRunner));
-            
+
             // Enrich test projects info with unit tests
             EnrichTestProjectsWithTestInfo(input.InitialTestRun, input.TestProjectsInfo);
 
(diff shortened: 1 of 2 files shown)


dupdob:
Thanks for this PR. This is not a review but a bunch of general comments:
1) you need to add unit test for this. I suggest isolating the source code searching/parsing algorithm so writing test is easier
2) note that this approach provides only partial support: unit tests can be renamed via framework specific attributes and theory like tests may include data as part of their name.

Overall, I think this is a good move, but I would like feedback from the core team regarding if this is an approach they support (i.e. this simply improving and supporting it until a fix from VsTest teams is made available)

rouke-broersma:
@dupdob sounds good to me, I have no objection and I agree with your review feedback as well.

src/Stryker.Core/Stryker.Core/Initialisation/ProjectMutator.cs
```diff
@@ -60,7 +69,49 @@ private void EnrichTestProjectsWithTestInfo(InitialTestRun initialTestRun, TestP
...
+                        var nodesInSpan = testFile.SyntaxTree.GetRoot().DescendantNodes(lineSpan);
+                        var node = nodesInSpan.First(n => n is MethodDeclarationSyntax);
+                        testFile.AddTest(unitTest.Id, unitTest.FullyQualifiedName, node);
+                    }
+                    else
+                    {
+                        _logger.LogDebug(
+                            "Could not locate unit test in any testfile. This should not happen and results in incorrect test reporting.");
+                    }
                 }
```
dupdob, line 115: Could be please move this logic to a dedicated class ? at least dedicated methods?

This code needs unit tests.

dupdob (changes requested):
The test detection logic should be moved to a specific class, or methods.
You need to add unit tests for this

dupdob:
Closing as PR is abandoned

PR #2753, opened 2023-11-03
fix(since): Retest all mutants covered by tests for which we cannot determine the filepath

### LINKED ISSUE #2752: The SinceMutantFilter does not mark the unit test as changed if for any reason the TestFilePath is empty
**Describe the bug**
When the SinceMutantFilter is active and there are tests that do not have a TestFilePath, then these are never marked as having been changed and needing a rescan.

**Expected behavior**
All covering tests that don't have a TestFilePath are marked as needing a rescan because Stryker is unable to determine if the file is part of the changeset and it is better to err on the side of caution in this case.

 - OS: Windows
 - Type of project framework
 - Framework Version : Net Framework 48
 - Stryker Version 3.10

MischaVreeburg:
All covering unit tests that don't have a TestFilePath are marked as pending.

Added shortcircuit for mutants having no covering tests

fix #2752


src/Stryker.Core/Stryker.Core/MutantFilters/SinceMutantFilter.cs
```diff
@@ -120,10 +120,13 @@ private IEnumerable<Mutant> ResetMutantStatusForChangedTests(IEnumerable<Mutant>
 
             foreach (var mutant in mutants)
             {
+                if (mutant.CoveringTests.IsEmpty || mutant.CoveringTests.Count == 0)
```
rouke-broersma, line 123: please expand this to a full if

src/Stryker.Core/Stryker.Core/MutantFilters/SinceMutantFilter.cs
```diff
@@ -120,10 +120,13 @@ private IEnumerable<Mutant> ResetMutantStatusForChangedTests(IEnumerable<Mutant>
...
+                    continue;
```
rouke-broersma, line 124:
```suggestion
                if (mutant.CoveringTests.IsEmpty || mutant.CoveringTests.Count == 0)
                {
                    continue;
                }
```

rouke-broersma:
Thank you for this fix and my apologies for not getting around to this for so long 🙏

PR #2783, opened 2023-11-29
Remove the 'please open an issue' part of safe mode error message.

dupdob:
In the past couple of years, all reported issues where non fixable. 
I suggest we keep the warning and remove the 'please open an issue' part.


dupdob:
Thanks for the idea.
New version downgrades the message to information (note that compile error is still a warning) and it explains that all mutations in method/accessor/constructor will be flagged as compile error.

dupdob:
YTF is it in draft ?
my mistake 😄

Issue #2799, opened 2023-12-15
dotnet8: process failed to connect to vstest.console (linux ssl issue)


draptik:
**Describe the bug**
Running `dotnet stryker` fails with error message `process failed to connect to vstest.console`.

**Logs**
Logs after running the command `dotnet stryker --verbosity trace --log-to-file` from the same folder as the `sln` file.

```txt
2023-12-15T13:28:55.0486719+01:00  [INF] Logging enabled at level Trace (8a18af78)
[...]

PR #2810, opened 2023-12-25
docs: add openssl workaround to 'known issues'

draptik:
This PR describes a workaround for OpenSSL to the "Known Issues" section which might happen on operating systems with newer OpenSSL implementations. 

https://github.com/stryker-mutator/stryker-net/issues/2799#issuecomment-1868758607

### DIFF
--- a/docs/technical-reference/VsTest internal.md
+++ b/docs/technical-reference/VsTest internal.md
@@ -39,3 +39,12 @@ Note that this also applies to session cancelled by the timeout logic. When this
 session will still wait for the end session event
 - Providing an incorrect TestSettings file will result in nothing happening, without any error report. This happens if the file is corrupted as well as with a negative timeouts
 - Some settings will be reported in the logs as not recognized or obsolete, but they are still needed by some test adapters
+- The error message `process failed to connect to vstest.console` might indicate that the OS is providing a different version of OpenSSL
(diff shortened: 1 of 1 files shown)


docs/technical-reference/VsTest internal.md
```diff
@@ -39,3 +39,12 @@ Note that this also applies to session cancelled by the timeout logic. When this
 session will still wait for the end session event
 - Providing an incorrect TestSettings file will result in nothing happening, without any error report. This happens if the file is corrupted as well as with a negative timeouts
 - Some settings will be reported in the logs as not recognized or obsolete, but they are still needed by some test adapters
+- The error message `process failed to connect to vstest.console` might indicate that the OS is providing a different version of OpenSSL
+  than what VsTest expects. F. ex. Arch provides OpenSSL3, whereas VsTest expects OpenSSL1. A more
```
Liam-Rougoor, line 43: I don't think `F. ex.` is a common abbreviation for for example. Use `e.g.` or spell it out completely instead.

Liam-Rougoor, line 43:
```suggestion
  than what VsTest expects. For example, Arch provides OpenSSL3, whereas VsTest expects OpenSSL1. A more
```

Liam-Rougoor: I don't think `F. ex.` is a common abbreviation for for example. Use `e.g.` or spell it out completely instead.

PR #2811, opened 2023-12-30
fix: Make sure to select valid IAnalyzerResult from Buildalyzer

0xced:
Just like #1900 but for the InputFileResolver.

This was not an issue until https://github.com/daveaglick/Buildalyzer/pull/198 was reverted in https://github.com/daveaglick/Buildalyzer/commit/8e85a15e324a65f9da5033f3d8a2320c76dda378 for the Buildalyzer 6.0.1 release. Before this change in Buildalyzer, the empty target framework was sorted last but since v6.0.1 it's sorted first instead.

Also, improve the NotSupportedException message when the language is undefined because `System.NotSupportedException: Specified method is not supported` is a terrible message.


rouke-broersma:
Thanks!

PR #2813, opened 2023-12-31
docs(contributing): Update debug log instruction

tommysor:
Existing instruction does not work (I assume the argument have been deprecated and removed).

Possibly this should be expanded with one more change as the instruction in `CONTRIBUTING.md` is similar to, but different from this:
https://github.com/stryker-mutator/stryker-net/blob/70adc60caad193ed6942ee970ea5373003ade4e4/.github/ISSUE_TEMPLATE/bug_report.md?plain=1#L13-L13


rouke-broersma:
Thanks!

Issue #2825, opened 2024-01-08
Cannot inject mutation error appeared in 3.13


cfuehrmann:
**Describe the bug**
A "cannot inject mutation" error has appeared in version 3.13. The error did not occur with 3.12.1. The _console_ output of `dotnet stryker --log-to-file` is here:

[CannotInjectMutation.log](https://github.com/stryker-mutator/stryker-net/files/13861065/CannotInjectMutation.log)

The (main?) log file written by Stryker is here:

[log-20240108.txt](https://github.com/stryker-mutator/stryker-net/files/13861095/log-20240108.txt)

**Expected behavior**
Stryker should succeed on the .NET project in question.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Type of project: library
 - Framework Version .NET 6 (using PolySharp to enable C# 11)
 - Stryker Version: 3:13

**Additional context**
I have a .NET 6 solution containing a dozen projects, on several of which I run Stryker. Since the update from Stryker 3.12 to 3.13, one of
 those projects causes Stryker to fail with "cannot inject mutation". Apparently when Stryker tries to mutate a `.First()` into a `.FirstOrDefault()`. Possibly important: The mutated statements are inside an anonymous function inside a LINQ `Select`:

```
			.Select(group =>
			{
				var employeePerson = group.First().TimeCorrection!.Employee.Person;
				return new WorkflowTimeCorrectionsReadResponse
				{
					EmployeeFirstName = employeePerson?.FirstName,
```

dupdob:
thank you for opening an issue for this. I may have a workaround for you: could add a Stryker comments above the problematic line ?
such as :
`// Stryker disable once: all` or `// Stryker disable once: Linq`

The mutant will still be generated but no attempt to inject it should happen. It will be reported as `Ignored`.
once you have pinpointed the offending line(s) could you please share the whole method so we can reproduce the problem and fix it properly ?

CrouseN:
I've encountered a similar issue with .NET 6 and Flurl. The error is "Cannot inject mutation '$""' in 'IsSuccessStatusCode' because we cannot find the original code." and has a stack trace similar to the OP. It is interesting that this is taking that bool (IsSuccessStatusCode) from the System.Net.Http.HttpResponseMessage and trying to change it to a string. I hope this helps in investigating this issue. Thank you for the workaround.

dupdob:
@CrouseN : thanks for the report. Could you please share the method where you encounter the issue ?
[...]

PR #2826, opened 2024-01-09
Fix mutation leak with SuppressNullableWarningSytanx constructs (!.)

dupdob:
Fix issue #2825


richardwerkman:
Fix looks good! If it's confirmed this solves the issue I'm fine with merging this

dupdob:
There is already a (new) unit test that does not pass on head but (obviously) does in my PR. So it does fix an existing bug.
Shall I ask the author of the issue to try this branch on their own project for confirmation?

Issue #2827, opened 2024-01-10
Stryker throwing ArgumentException (key conflict) after running tests in a project


mstoicmicrosoft:
**Describe the bug**
/_/src/Stryker.CLI/Stryker.CLI/Program.cs:line 14 is throwing an Unhandled exception. System.ArgumentException: An element with the same key but a different value already exists. Key: 'CS8002' after running several tests in projects. 

**Logs**
Console output error:
Unhandled exception. System.ArgumentException: An element with the same key but a different value already exists. Key: 'CS8002'
   at System.Collections.Immutable.ImmutableDictionary`2.HashBucket.Add(TKey key, TValue value, IEqualityComparer`1 keyOnlyComparer, IEqualityComparer`1 valueComparer, KeyCollisionBehavior behavior, OperationResult& result)
   at System.Collections.Immutable.ImmutableDictionary`2.AddRange(IEnumerable`1 items, MutationInput origin, KeyCollisionBehavior collisionBehavior)
   at System.Collections.Immutable.ImmutableDictionary`2.AddRange(IEnumerable`1 pairs, Boolean avoidToHashMap)
   at System.Collections.Immutable.ImmutableDictionary`2.AddRange(IEnumerable`1 pairs)
   at System.Collections.Immutable.ImmutableDictionary.ToImmutableDictionary[TKey,TValue](IEnumerable`1 source, IEqualityComparer`1 keyComparer, IEqualityComparer`1 valueComparer)
   at System.Collections.Immutable.ImmutableDictionary.ToImmutableDictionary[TKey,TValue](IEnumerable`1 source)
   at Stryker.Core.Initialisation.Buildalyzer.IAnalyzerResultExtensions.GetDiagnosticOptions(IAnalyzerResult analyzerResult) in /_/src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultExtensions.cs:line 187
   at Stryker.Core.Initialisation.Buildalyzer.IAnalyzerResultCSharpExtensions.GetCompilationOptions(IAnalyzerResult analyzerResult) in /_/src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultCSharpExtensions.cs:line 13
[...]

PR #2828, opened 2024-01-11
fix: Stryker fails if project has conflicting settings regarding warnings

dupdob:
fix issue #2827

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultExtensions.cs
+++ b/src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultExtensions.cs
@@ -183,7 +183,16 @@ public static class IAnalyzerResultExtensions
         var warningsNotAsErrorsString = analyzerResult.GetPropertyOrDefault("WarningsNotAsErrors");
         var warningsNotAsErrors = warningsNotAsErrorsString is not null ? warningsNotAsErrorsString.Split(";").Where(x => !string.IsNullOrWhiteSpace(x)).ToDictionary(x => x, x => ReportDiagnostic.Warn) : new Dictionary<string, ReportDiagnostic>();
 
-        var diagnosticOptions = noWarn.Concat(warningsAsErrors).Concat(warningsNotAsErrors);
+        // merge settings,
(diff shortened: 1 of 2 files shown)


src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultExtensions.cs
```diff
@@ -183,7 +183,16 @@ internal static string GetAssemblyOriginatorKeyFile(this IAnalyzerResult analyze
         var warningsNotAsErrorsString = analyzerResult.GetPropertyOrDefault("WarningsNotAsErrors");
         var warningsNotAsErrors = warningsNotAsErrorsString is not null ? warningsNotAsErrorsString.Split(";").Where(x => !string.IsNullOrWhiteSpace(x)).ToDictionary(x => x, x => ReportDiagnostic.Warn) : new Dictionary<string, ReportDiagnostic>();
 
-        var diagnosticOptions = noWarn.Concat(warningsAsErrors).Concat(warningsNotAsErrors);
+        // merge settings,
```
rouke-broersma, line 186: I think we should at the very least document this behavior, but preferably find the exact logic used by msbuild which apparently accepts this configuration and copy that. With this implementation it seems like we're making assumptions about the severity of the diagnostic.

My personal preference when we can't find what msbuild does would be to treat warningsAsErrors as more imporant than the reverse, so we should process that last.

dupdob: agree it should be documented. My ideal is that it followed what actually happens, but I fail to find any documentation about this (it seems to be undocumented).
The problem with prioritizing warnaserror above the errors is that the project may not compile at all without any mutations. So I have chosen a more tolerant approach

rouke-broersma: Actually never mind, this is safer because the user can always decide that they had intended for the mutant to be killed by the error and they would then know they made a configuration mistake instead of the mutant being silently killed when they did not intend it.

PR #2845, opened 2024-02-14
fix: Improve build, restore and test

dupdob:
Misc fixes:

1. Drop use of now obsolete VsTest async APIs (but keep some async behaviour to deal with VsTest freezes)
2. Ensure project analysis is done in the same way between Project and Solution mode
3. Improve readability of the log dump of Buildalyzer results
4. Log Stryker version number at start
5. Restore packages and retry build when it fails
6. Respect MaxConcurrency settings when doing project analysis
7. Correct logic when failing to get MsBuild version. Not sure if this can happen, but existing logic logged an error and tried to keep on while using the error message as a version number.
8. Enlarge search perimeter for nuget.exe (from drive root) when not found in path
9. Log an error when failing to restore packages
10. Adjust VsTest settings to prevent freezes with latest versions (force using an external test host)
11. Add a logging workflow for DataCollector and log an error when failing to retrieve MutantControl info
12. No longer tests the project in a solution against all test projects (reduce needless run time)

other changes:
`ProcessExecutor.Start` method now longer alters the path used for execution, some uses have been adjusted accordingly. Note that this may fix some surprising bugs in non standard configuration.
~~Note: this PR is in draft mode as I try to fix some new VsTest freeze scenarios whit small test projects~~


dupdob:
🙏

rouke-broersma:
> 🙏 

Thanks for the improvements 😁

PR #2863, opened 2024-03-01
feat(mutators): Add default parameter mutator.

### LINKED ISSUE #2848: Mutator for default arguments.
Default parameters could not be mutated in the method declaration due to the cause of compile errors. However, since the semantic model was introduced, we could mutate method invocations that make use of default parameters.  

I'd like to see a mutator that mutates method invocations that use default parameters. 

**Additional context**
An example:
```
void myMethodWithDefaultParameter(string defaultParam = "Hello Stryker!"){
...
}
```
When the method is called like this: 
`myMethodWithDefaultParameter();`

I'd expect a mutation like:
`myMehodWithDefaultParameter("");`

Liam-Rougoor:
Closes #2848.

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Mutants/ICsharpMutantOrchestrator.cs
+++ b/src/Stryker.Core/Stryker.Core/Mutants/ICsharpMutantOrchestrator.cs
@@ -1,0 +1,31 @@
+﻿using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
+using Stryker.Core.Mutators;
+
+namespace Stryker.Core.Mutants;
+
+public interface ICsharpMutantOrchestrator
+{
+    IEnumerable<IMutator> Mutators { get; }
+    MutantPlacer Placer { get; }
+    bool MustInjectCoverageLogic { get; }
+    ICollection<Mutant> Mutants { get; set; }
+    int MutantCount { get; set; }
+
+    /// <summary>
+    /// Recursively mutates a single SyntaxNode
+    /// </summary>
+    /// <param name="input">The current root node</param>
+    /// <returns>Mutated node</returns>
+    SyntaxNode Mutate(SyntaxNode input, SemanticModel semanticModel);
+
+    /// <summary>
+    /// Returns true if the new mutant is a duplicate of a mutant already listed in Mutants.
+    /// </summary>
+    bool IsMutantDuplicate(Mutant newMutant, Mutation mutation);
+
+    /// <summary>
+    /// Gets the stored mutants and resets the mutant list to an empty collection
+    /// </summary>
+    IReadOnlyCollection<Mutant> GetLatestMutantBatch();
+}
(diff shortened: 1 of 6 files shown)


src/Stryker.Core/Stryker.Core/Mutants/ICsharpMutantOrchestrator.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
+using Stryker.Core.Mutators;
+
+namespace Stryker.Core.Mutants;
+
+public interface ICsharpMutantOrchestrator
```
richardwerkman, line 7:
```suggestion
public interface ICSharpMutantOrchestrator
```

src/Stryker.Core/Stryker.Core/Mutators/DefaultParameterMutator.cs
```diff
@@ -0,0 +1,94 @@
...
+                    OriginalNode = node,
+                    ReplacementNode = node.WithArgumentList(newArgumentList),
+                    DisplayName = $"Default parameter mutation: {parameterName}",
+                    Type = Mutator.DefaultParameter
+                };
+            }
+        }
+    }
+
+    public override MutationLevel MutationLevel => MutationLevel.Standard;
```
richardwerkman, line 75: First of all, I would place this at the top of the class, not here. Like in our other mutator classes. And I think we should debate whether this should be a level standard mutation. This could add a lot of mutations.

src/Stryker.Core/Stryker.Core/Mutators/DefaultParameterMutator.cs
```diff
@@ -0,0 +1,94 @@
...
+    public DefaultParameterMutator(ICsharpMutantOrchestrator orchestrator, StrykerOptions options)
+    {
+        _orchestrator = orchestrator;
+        _options = options;
+    }
+
+    public override IEnumerable<Mutation> ApplyMutations(InvocationExpressionSyntax node, SemanticModel semanticModel)
+    {
+        var methodSymbol = (IMethodSymbol)semanticModel.GetSymbolInfo(node).Symbol;
+        if (methodSymbol is null)
```
richardwerkman, line 25: I think we should check whether the symbol is internal or external. We might not want to mutate optional parameters of external APIs like `System.Linq`. @rouke-broersma what do you think?

richardwerkman (changes requested):
Nice! But we might need to think about some things before merging this

src/Stryker.Core/Stryker.Core/Mutants/ICsharpMutantOrchestrator.cs
Liam-Rougoor: The original class is also called `CsharpMutantOrchestrator`..
Do you want me to keep it as is, or do you want me to rename both the interface and the class to `CSharp...`?

rouke-broersma: Would be best to fix the casing

Liam-Rougoor:
Merging with master was a bit of a hassle, it was easier to recreate #3024

Issue #2867, opened 2024-03-04
Stryker.NET encountered a compile error in Stryker.Core.InjectedHelpers.MutantControl.cs (at 14:15) with message: Non-nullable field '[...]' must contain a non-null value when exiting constructor


tristanevain:
**Describe the bug**
On one specific (but not exotic) .net 8 project, Stryker fails and repeatedly prints the error in the title. It happens on Windows machines but not when running in an Azure pipeline.
Other projects of the same solution run fine.

The flow of repeating, and similar errors ends with the following message:

> Stryker.NET could not compile the project after mutation. This is probably an error for Stryker.NET and not your project. Please report this issue on github with the previous error message.

**Logs**

Note that the place in the file is not always the same. Neither is the name of the faulty field.

> [09:17:33 WRN] Stryker.NET encountered a compile error in Stryker.Core.InjectedHelpers.MutantControl.cs (at 14:15) with message: Non-nullable field '_coveredMutants' must contain a non-null value when exiting constructor. Consider declaring the field as nullable. (Source code: static MutantControl()
>         {
>             InitCoverage();
>         })
> [09:17:33 INF] Safe Mode! Stryker will flag mutations in MutantControl as compile error.
> [09:17:33 WRN] Stryker.NET encountered a compile error in Stryker.Core.InjectedHelpers.MutantControl.cs (at 14:15) with message: Non-nullable field '_coveredStaticdMutants' must contain a non-null value when exiting constructor. Consider declaring the field as nullable. (Source code: static MutantControl()
>         {
>             InitCoverage();
>         })
> [09:17:33 INF] Safe Mode! Stryker will flag mutations in MutantControl as compile error.
> [09:17:33 WRN] Stryker.NET encountered a compile error in Stryker.Core.InjectedHelpers.MutantControl.cs (at 14:15) with message: Non-nullable field 'envName' must contain a non-null value when exiting constructor. Consider declaring the field as nullable. (Source code: static MutantControl()
>         {
>             InitCoverage();
>         })
> [09:17:33 INF] Safe Mode! Stryker will flag mutations in MutantControl as compile error.
> [09:17:33 WRN] Stryker.NET encountered a compile error in Stryker.Core.InjectedHelpers.MutantControl.cs (at 53:45) with message: Converting null literal or possible null value to non-nullable type. (Source code: System.Environment.GetEnvironmentVariable("ActiveMutation"))
[...]

PR #2869, opened 2024-03-04
fix: Ensure MutationControl compiles with 'nullable' enabled

dupdob:
Ensure MutationControl class, which is injected into mutated code, compiles without warnings with 'nullable' option on while remaining compatible with C# 2
Fix issue #2867


rouke-broersma:
@dupdob while I think it's good to have proper nullable code anyway, wouldn't we want to explicitly disable nullable for the injected code? `#nullable disable` Or does that not work on older target frameworks?

dupdob:
you guess right: it was the first thing I attempted. The pragma raises an error for pre CSharp 8

PR #2903, opened 2024-04-07
feat: adding mutation score to json reporter

phmonte:
### Feature:
Adding mutation score to json reporter.

### Description:
The markdown and html reporters have the mutation score, using the same logic, adding the mutation score to the json reporter.


rouke-broersma:
The json reporter implements the mutation report schema, and adheres to this schema. The schema does not contain a score, instead anything using the report should calculate the result at presentation time. This decision was made so that if for some reason the calculation changes (more states get added or some other reason) the json will still be valid without having to be regenerated. To make this easier libraries have been created for js/ts and scala/java that implement the algorithm. Such a library does not yet exist for dotnet but can be of there's sufficient interest.

phmonte:
@rouke-broersma, thanks for the explanation, I'll close this PR.
I'm doing this work manually so as not to copy the markdown reporter string, I intend to create the metrics for dotnet.

Issue #2929, opened 2024-05-02
Wrong Logging usage of unexpected test case in initial test run leads to stryker crash


Dirk-Peters-BF:
**Describe the bug**
Stryker run is aborted due to crash when a specific logging call in `VsTestRunner.InitialTest` Line 60 is used, when a new unexpected test case is detected in the intial test run.

**Logs**
         System.FormatException: Input string was not in a correct format. Failure to parse near offset 132. Expected an ASCII digit.
00:02:36      at System.Text.CompositeFormat.Parse(String format)
00:02:36      at Microsoft.Extensions.Logging.LogValuesFormatter..ctor(String format)
00:02:36      at Microsoft.Extensions.Logging.FormattedLogValues.<>c.<.ctor>b__9_0(String f)
00:02:36      at Microsoft.Extensions.Logging.FormattedLogValues..ctor(String format, Object[] values)
00:02:36      at Microsoft.Extensions.Logging.LoggerExtensions.Log(ILogger logger, LogLevel logLevel, EventId eventId, Exception exception, String message, Object[] args)
00:02:36      at Microsoft.Extensions.Logging.LoggerExtensions.Log(ILogger logger, LogLevel logLevel, String message, Object[] args)
00:02:36      at Microsoft.Extensions.Logging.LoggerExtensions.LogWarning(ILogger logger, String message, Object[] args)
00:02:36      at Stryker.Core.TestRunners.VsTest.VsTestRunner.InitialTest(IProjectAndTests project) in /_/src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs:line 60
00:02:36      at Stryker.Core.TestRunners.VsTest.VsTestRunnerPool.<>c__DisplayClass12_0.<InitialTest>b__0(VsTestRunner runner) in /_/src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunnerPool.cs:line 62
00:02:36      at Stryker.Core.TestRunners.VsTest.VsTestRunnerPool.RunThis[T](Func`2 task) in /_/src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunnerPool.cs:line 103
00:02:36      at Stryker.Core.TestRunners.VsTest.VsTestRunnerPool.InitialTest(IProjectAndTests project) in /_/src/Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunnerPool.cs:line 62
[...]

PR #2931, opened 2024-05-03
fix: Wrong Logging usage (#2929)

Dirk-Peters-BF:
- fix combination of string interpolation with parameters in logging call in Stryker.Core/Stryker.Core/TestRunners/VsTest/VsTestRunner.cs
- harmonize logging calls:
  * remove string interpolation
  * use parameter name instead of index
  * use constant log messages


rouke-broersma:
Thank you and sorry for the delay!

PR #2941, opened 2024-05-22
chore(testrunner): Add abstractions to decouple from VSTest

MarcZw:
# Why?
Stryker should be able to support other testrunners apart from VSTest. 
This is *not possible* in the current implementation of Stryker, because of the tight coupling with VSTest. 
Therefore some abstractions needed to be created to facilitate the decoupling from VSTest. 

# Solution
The current solution to decouple is to introduce a new project, Stryker.Shared. 
This project contains abstractions to facilitate other testrunners, without breaking the current implementation. 

**Note** this is a temporary solution. There are plans for an architectural overhaul, to decouple Stryker even more.


rouke-broersma:
Implemented by #3064

PR #2968, opened 2024-06-19
Integer mutators

### LINKED ISSUE #2849: Mutator for numbers
Similar to the `StringMutator`, we could mutate numbers.

I'd like to see a mutator for numbers. Example mutations could be:
| Original | Mutation |
| --------- | ---------- |
| var x = 10; | var x = 0; |
| var x = 10; | var x = -10; |

With semantic model, we could add some mutations on numeric variables as well:
| Original | Mutation |
| --------- | ---------- |
| var x = myNumericVariable; | var x = -myNumericVariable; | 
> This example might be bad if myNumericVariable equals 0..

YVbakker:
This pull request will add the following mutators:
- Integer nullification mutator
- Integer negation mutator

### Integer nullification mutator

Changes numeric literal expressions of type integer into 0. 0 is not mutated.

Examples:

- `var x = 10` → `var x = 0`
- `var x = -10` → `var x = 0`

### Integer negation mutator

Changes numeric literal expressions of type integer into the negated form. 0 cannot be negated.

Examples:

- `var x = 10` → `var x = -10`
- `var x = -10` → `var x = 10`

closes #2849

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/CsharpMutantOrchestratorTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/CsharpMutantOrchestratorTests.cs
@@ -13,67 +13,75 @@ public class CsharpMutantOrchestratorTests : MutantOrchestratorTestsBase
     [Fact]
     public void ShouldNotMutateEmptyInterfaces()
     {
-        var source = @"using System;
-using System.Collections.Generic;
-using System.Text;
-namespace StrykerNet.UnitTest.Mutants.TestResources
-{
-    interface TestClass
-    {
-        int A { get; set; }
-        int B { get; set; }
-        void MethodA();
-    }
-}";
+        const string Source = """
+                              using System;
+                              using System.Collections.Generic;
+                              using System.Text;
+                              namespace StrykerNet.UnitTest.Mutants.TestResources
+                              {
+                                  interface TestClass
+                                  {
+                                      int A { get; set; }
+                                      int B { get; set; }
+                                      void MethodA();
+                                  }
+                              }
+                              """;
 
-        var expected = @"using System;
-using System.Collections.Generic;
-using System.Text;
-namespace StrykerNet.UnitTest.Mutants.TestResources
-{
-    interface TestClass
-    {
-        int A { get; set; }
-        int B { get; set; }
-        void MethodA();
-    }
-}";
-        ShouldMutateSourceToExpected(source, expected);
+        const string Expected = """
+                                using System;
+                                using System.Collections.Generic;
+                                using System.Text;
+                                namespace StrykerNet.UnitTest.Mutants.TestResources
+                                {
+                                    interface TestClass
+                                    {
+                                        int A { get; set; }
+                                        int B { get; set; }
+                                        void MethodA();
+                                    }
+                                }
+                                """;
+        ShouldMutateSourceToExpected(Source, Expected);
     }
 
     [Fact]
     public void ShouldMutateDefaultImplementationInterfaces()
     {
-        var source = @"using System;
-using System.Collections.Generic;
-using System.Text;
-namespace StrykerNet.UnitTest.Mutants.TestResources
-{
-    interface TestClass
-    {
-        int A { get; set; }
-        int B { get; set; }
-        void MethodA() {
-            var three = 1 + 2;
-        }
-    }
-}";
+        const string Source = """
+                              using System;
+                              using System.Collections.Generic;
+                              using System.Text;
+                              namespace StrykerNet.UnitTest.Mutants.TestResources
+                              {
+                                  interface TestClass
+                                  {
+                                      int A { get; set; }
+                                      int B { get; set; }
+                                      void MethodA() {
+                                          var three = 1 + 2;
+                                      }
+                                  }
+                              }
+                              """;
 
-        var expected = @"using System;
-using System.Collections.Generic;
-using System.Text;
-namespace StrykerNet.UnitTest.Mutants.TestResources
-{
-    interface TestClass
-    {
-        int A { get; set; }
-        int B { get; set; }
-        void MethodA() {if(StrykerNamespace.MutantControl.IsActive(0)){}else{
-            var three = (StrykerNamespace.MutantControl.IsActive(1)?1 - 2:1 + 2);
-        }}
-    }
-}";
-        ShouldMutateSourceToExpected(source, expected);
+        const string Expected = """
+                                using System;
+                                using System.Collections.Generic;
+                                using System.Text;
+                                namespace StrykerNet.UnitTest.Mutants.TestResources
+                                {
+                                    interface TestClass
+                                    {
+                                        int A { get; set; }
+                                        int B { get; set; }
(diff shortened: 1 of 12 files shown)


src/Stryker.Core/Stryker.Core/Mutators/IntegerNegationMutator.cs
```diff
@@ -0,0 +1,36 @@
...
+            ReplacementNode = SyntaxFactory.LiteralExpression(SyntaxKind.NumericLiteralExpression, SyntaxFactory.Literal(replacementValue)),
+            DisplayName = "Number negation mutation",
+            Type = Mutator.Number
+        };
+    }
+
+    private static bool IsNumberLiteral(LiteralExpressionSyntax node)
+    {
+        var kind = node.Kind();
+        return kind == SyntaxKind.NumericLiteralExpression && node.Parent is EqualsValueClauseSyntax;
```
dupdob, line 33: why limiting this mutation to assignment ? it seems that it could be uses within any expression (excluding indexed access `[...]` ). It should at least also support method arguments.

src/Stryker.Core/Stryker.Core/Mutators/IntegerNullificationMutator.cs
```diff
@@ -0,0 +1,32 @@
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Stryker.Core.Mutants;
+
+namespace Stryker.Core.Mutators;
+
+public class IntegerNullificationMutator : MutatorBase<LiteralExpressionSyntax>
```
dupdob, line 9: Why having two separate mutators ?

src/Stryker.Core/Stryker.Core/Mutators/IntegerNegationMutator.cs
```diff
@@ -0,0 +1,36 @@
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Stryker.Core.Mutants;
+
+namespace Stryker.Core.Mutators;
+
+public class IntegerNegationMutator : MutatorBase<LiteralExpressionSyntax>
```
dupdob, line 9: why having two separate mutators?

src/Stryker.Core/Stryker.Core/Mutators/IntegerNullificationMutator.cs
```diff
@@ -0,0 +1,32 @@
...
+                DisplayName = "Integer nullification mutation",
+                Type = Mutator.Number
+            };
+        }
+    }
+
+    private static bool IsNumberLiteral(LiteralExpressionSyntax node)
+    {
+        var kind = node.Kind();
+        return kind == SyntaxKind.NumericLiteralExpression && node.Parent is EqualsValueClauseSyntax;
```
dupdob, line 30: why limiting this mutation to assignment ? it seems that it could be uses within any expression. It should at least also support method arguments.

src/Stryker.Core/Stryker.Core.UnitTest/Mutants/CsharpMutantOrchestratorTests.cs
```diff
@@ -13,67 +13,75 @@ public class CsharpMutantOrchestratorTests : MutantOrchestratorTestsBase
...
-{
-    interface TestClass
-    {
-        int A { get; set; }
-        int B { get; set; }
-        void MethodA();
-    }
-}";
-        ShouldMutateSourceToExpected(source, expected);
+        const string Source = """
```
dupdob, line 16: Any specific reason fro switching to raw strings?
I do not feel this has significant benefits here

dupdob (changes requested):
thanks for this PR. I let you answer inline code questions.
You need to adjust integration test results, since adding a mutator changes them

YVbakker:
> thanks for this PR. I let you answer inline code questions. You need to adjust integration test results, since adding a mutator changes them

Thanks for your feedback, I'll get back to it this week

src/Stryker.Core/Stryker.Core/Mutators/IntegerNullificationMutator.cs
YVbakker: I figured they can possibly be conflicting, and so I thought it'd be a good idea to add them separately. 
When mutating `var x = 10` for instance, would this be `var x = 0`? `var x = -10`?
Maybe I'm all wrong about this and it doesn't work this way at all: if this is not an issue it could be a single mutator.

src/Stryker.Core/Stryker.Core/Mutators/IntegerNullificationMutator.cs
YVbakker: great idea! initially, I didn't have a constraint on the node parent at all so all occurrences would be mutated. I then constrained it to at least work for the example in the original issue, and I kinda left it open for suggestions if other places should receive this mutation as well

dupdob: limitation should only be placed if one detect an actual issue. But most (if not all) risky constructs, such as constants or attributes, are already identified and dealt with.

src/Stryker.Core/Stryker.Core.UnitTest/Mutants/CsharpMutantOrchestratorTests.cs
YVbakker: I see what you mean, however, I think this improves readability. Also, Microsoft [recommends](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/strings/#raw-string-literals) to use them when dealing with pre-formatted text. I think it's more suitable for source code snippets as those used in this test file.

dupdob: I u de it comes from a good place. But this kind of change drags a huge, invisible cost: PRs

Any PR referencing the previous version will be painful to merge. 
I am not sure the benefits outweigh this drawback. 
Been there, done that.
We will need to check with the core team what they feel about this.

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/IntegerNegationMutatorTests.cs
```diff
@@ -0,0 +1,51 @@
...
+using Stryker.Core.Mutators;
+using System.Linq;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Xunit;
+
+namespace Stryker.Core.UnitTest.Mutators;
+
+public class IntegerNegationMutatorTests : TestBase
+{
+    [Fact]
```
rouke-broersma, line 12: We have recently migrated all unit tests from xunit to mstest, sorry! Could you migrate your unit tests to mstest?

YVbakker:
I have been exceptionally busy due to the holidays, but I did read your further feedback and I will incorporate it

PR #2985, opened 2024-07-18
fix(real-time-reporting): Properly handle when browser is closed during mutation testing

### LINKED ISSUE #2517: Real time reporting can crash stryker when a client disconnects right before or during a crash
**Describe the bug**

https://github.com/stryker-mutator/stryker-net/blob/c116bd9bc4c3dc030dc5025873810f2341945fa0/src/Stryker.Core/Stryker.Core/Reporters/Html/Realtime/SseServer.cs#L68

Disconnected writes should be removed from the list of writers and some kinda exception handling should catch when a client disconnects in-flight.

dupdob:
Fix #2517.


dupdob:
Crap: the added unit test does in fact cover the new code, but I cannot make this test works properly on the CI pipeline due to the unpredictable behavior of `HttpClient`.
It obviously works like a charm on my machine...

PR #2988, opened 2024-07-19
DRAFT testrunner

richardwerkman:


dupdob:
quick architecture question: it looks like the test projects are loaded into the current process, right?
then, how can we deal with different framework versions and platforms?

richardwerkman:
@dupdob this is all very early stage and as you see contains lots of merge conflicts. We will split this work into smaller PRs because this isn't easily reviewable. This branch is the outcome of an internship and will be used as inspiration for the final implementation.

To answer your question, we don't know yet. We will have to investigate that once the feature is in a state where we can test that. The current implementation is only single threaded and only works for a single project (as far as I know).

rouke-broersma:
> quick architecture question: it looks like the test projects are loaded into the current process, right? then, how can we deal with different framework versions and platforms?

At some point we will have to think about spawning new processes for these cases I think.

Issue #1211, opened 2020-09-17
Consider using an ioc container to ease readability and testability


g7ed6e:
**Describe the bug**
When debugging stryker i noticed multiples instanciation of the `GitInfoProvider`. Also so many ctor args are declared with default null values to allow unit testing.

**Expected behavior**
Using an ioc container would allow to remove default null parameters in ctor parameters, newing implementation in ctor body and have finegrained control over object lifetimes

Mobrockers:
Hi! This is legacy from when stryker was not yet such a bug project. You are right that we should probably revisit and decide if it's time for dependency injection.

g7ed6e:
> Hi! This is legacy from when stryker was not yet such a bug project. You are right that we should probably revisit and decide if it's time for dependency injection.

I think you meant _big_ project ;)

Mobrockers:
Potato potato :)

g7ed6e:
Awesome!

PR #2992, opened 2024-07-23
feat(mutating): constant patterns

### LINKED ISSUE #2951: List pattern syntax mutation
The following code (in the switch expression) should be mutated but is not:

``` cs
    public void ListPatterns()
    {
        decimal balance = 0m;
        string[][] transactions = new[]
        {
            new[] { "1", "DEPOSIT", "100.00" },
        };
        foreach (string[] transaction in transactions)
        {
            balance += transaction switch
            {
            // "DEPOSIT" should be mutated here
            [_, "DEPOSIT", _, var amount] => decimal.Parse(amount),
            [_, "WITHDRAWAL", .., var amount] => -decimal.Parse(amount),
            [_, "INTEREST", var amount] => decimal.Parse(amount),
            [_, "FEE", var fee] => -decimal.Parse(fee),
                _ => throw new InvalidOperationException($"Record {string.Join(", ", transaction)} is not in the expected format!"),
            };
            Console.WriteLine($"Record: {string.Join(", ", transaction)}, New balance: {balance:C}");
        }
    }
```

We should introduce a new orchestrator for the syntax type `ListPatternSyntax`.

dupdob:
Main (unique?) visible impact is the ability to mutate strings that are part of case in switch expressions.
Note that this allows new mutators (that remain to be defined as of today).
Fix #2951

Not related: 
- Wrap Spectre Console for usage with CommandLineUtils. This increase overall consistency and fixes a flaky test.
- Mutant IDs generation is in a dedicated component
- Restore stable and known mutant IDs for unit tests

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Mutants/BaseMutantOrchestrator.cs
+++ b/src/Stryker.Core/Stryker.Core/Mutants/BaseMutantOrchestrator.cs
@@ -4,7 +4,6 @@ using Stryker.Core.Options;
 
 namespace Stryker.Core.Mutants;
 
-
 public abstract class BaseMutantOrchestrator
 {
     public readonly StrykerOptions Options;
@@ -23,7 +22,6 @@ public abstract class BaseMutantOrchestrator
     public ICollection<Mutant> Mutants { get; set; }
 
     protected int GetNextId() => _idProvider.NextId();
-
 
     /// <summary>
     /// Gets the stored mutants and resets the mutant list to an empty collection
--- a/src/Stryker.Core/Stryker.Core/Mutants/BaseMutantOrchestrator.cs
+++ b/src/Stryker.Core/Stryker.Core/Mutants/BaseMutantOrchestrator.cs
@@ -1,50 +1,51 @@
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
-using System.Threading;
 using Stryker.Core.Options;
 
-namespace Stryker.Core.Mutants
+namespace Stryker.Core.Mutants;
+
+
+public abstract class BaseMutantOrchestrator
 {
-    public abstract class BaseMutantOrchestrator
+    public readonly StrykerOptions Options;
+    private readonly IProvideId _idProvider;
+
+    protected BaseMutantOrchestrator(StrykerOptions options)
     {
-        public readonly StrykerOptions Options;
-
-        public bool MustInjectCoverageLogic =>
-            Options != null && Options.OptimizationMode.HasFlag(OptimizationModes.CoverageBasedTest) &&
-            !Options.OptimizationMode.HasFlag(OptimizationModes.CaptureCoveragePerTest);
-
-        public ICollection<Mutant> Mutants { get; set; }
-
-        private static int _mutantCount;
-
-        protected static int MutantCount => _mutantCount;
-
-        protected static void IncreaseMutantCount() => Interlocked.Increment(ref _mutantCount);
-
-        protected BaseMutantOrchestrator(StrykerOptions options) => Options = options;
-
-        /// <summary>
-        /// Gets the stored mutants and resets the mutant list to an empty collection
-        /// </summary>
-        public virtual IReadOnlyCollection<Mutant> GetLatestMutantBatch()
-        {
-            var tempMutants = Mutants;
-            Mutants = new Collection<Mutant>();
-            return (IReadOnlyCollection<Mutant>)tempMutants;
-        }
+        Options = options;
+        _idProvider = Options.MutantIdProvider ?? new BasicIdProvider();
     }
+
+    public bool MustInjectCoverageLogic =>
+        Options != null && Options.OptimizationMode.HasFlag(OptimizationModes.CoverageBasedTest) &&
+        !Options.OptimizationMode.HasFlag(OptimizationModes.CaptureCoveragePerTest);
+
+    public ICollection<Mutant> Mutants { get; set; }
+
+    protected int GetNextId() => _idProvider.NextId();
+
 
     /// <summary>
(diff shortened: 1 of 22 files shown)


src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
```diff
@@ -224,7 +225,13 @@ public class StrykerOptions
     /// Instruct Stryker to break execution when at least one test failed on initial run.
     /// </summary>
     public bool BreakOnInitialTestFailure { get; set; }
-    
+
+
+    /// <summary>
+    /// Get/set the mutation id provider
+    /// </summary>
+    public IProvideId MutantIdProvider {get; set;}
```
richardwerkman, line 233: Why is this placed in the options? This doesn't seem the right place to me. Options are settings that can be altered by the user.

src/Stryker.Core/Stryker.Core/Mutants/BaseMutantOrchestrator.cs
```diff
@@ -1,50 +1,51 @@
...
-        protected static int MutantCount => _mutantCount;
+    public bool MustInjectCoverageLogic =>
+        Options != null && Options.OptimizationMode.HasFlag(OptimizationModes.CoverageBasedTest) &&
+        !Options.OptimizationMode.HasFlag(OptimizationModes.CaptureCoveragePerTest);
 
-        protected static void IncreaseMutantCount() => Interlocked.Increment(ref _mutantCount);
+    public ICollection<Mutant> Mutants { get; set; }
 
-        protected BaseMutantOrchestrator(StrykerOptions options) => Options = options;
+    protected int GetNextId() => _idProvider.NextId();
```
richardwerkman, line 25: I like this change! This enables us to revert the orchestrator tests to check for mutant ids right?

src/Stryker.Core/Stryker.Core/Mutants/MutationStore.cs
```diff
@@ -160,8 +171,9 @@ public bool StoreMutations(IEnumerable<Mutant> store)
     /// <returns>a syntax expression with the mutations included </returns>
     public ExpressionSyntax Inject(ExpressionSyntax mutatedNode, ExpressionSyntax sourceNode)
     {
-        if (_pendingMutations.Peek().Control == MutationControl.MemberAccess)
+        if (_injectionBlockCounter>0 || _pendingMutations.Peek().Control == MutationControl.MemberAccess)
```
richardwerkman, line 174:
```suggestion
        if (_injectionBlockCounter > 0  ||  _pendingMutations.Peek().Control == MutationControl.MemberAccess)
```

richardwerkman (dismissed):
Thanks for the great work again :)

I see why you put the IdProvider in the options but I don't like it. Can't we pass it to the orchestrator in another way?

dupdob:
> Thanks for the great work again :)
> 
> I see why you put the IdProvider in the options but I don't like it. Can't we pass it to the orchestrator in another way?

I am not completely satisfied with this either, but it turns out I had to add it as a parameter to a lot of classes.
I then realized that `Options` is the proper vehicle for this: options is the repository for all customizable/variable behavior.

And we could have a customizable ID generation logic, with two or three options:
1. Technical ID: as a user, I don't care for the ID, I let Stryker decide. This would be as of today
2. Reproducible IDs: as as user, I want IDs to be stable across several runs (on the same code base version). An improved version of previous logic. It could be that each project's IDs would be in specific ranges, such as 0+Id, 1000000+Id, 2000000+Id...
3. Stable IDs: as a user, I want IDs to be stable across code base version. Those should be alphanumeric, using this pattern: _class.member.id_, eg. `MyClass.MyMethed.12`.

This option would clearly be in `Options` and converted to the proper strategy class during project init.

Makes sense?

src/Stryker.Core/Stryker.Core/Options/StrykerOptions.cs
dupdob: I am not completely satisfied with this either, but it turns out I had to add it as a parameter to a lot of classes.
I then realized that `Options` is the proper vehicle for this: options is the repository for all customizable/variable behavior.

And we could have a customizable ID generation logic, with two or three options:
1. Technical ID: as a user, I don't care for the ID, I let Stryker decide. This would be as of today
2. Reproducible IDs: as as user, I want IDs to be stable across several runs (on the same code base version). An improved version of previous logic. It could be that each project's IDs would be in specific ranges, such as 0+Id, 1000000+Id, 2000000+Id...
3. Stable IDs: as a user, I want IDs to be stable across code base version. Those should be alphanumeric, using this pattern: _class.member.id_, eg. `MyClass.MyMethed.12`.

This option would clearly be in `Options` and converted to the proper strategy class during project init.

Makes sense?

src/Stryker.Core/Stryker.Core/Mutants/BaseMutantOrchestrator.cs
dupdob: yes. Loosing IDs for unit tests was a problem for me: it was harder to verify I got the proper mutations when looking at the unit test error messages.
And I realized this loss was quite problematic indeed.

dupdob:
>I see why you put the IdProvider in the options but I don't like it. Can't we pass it to the orchestrator in another way?
Do you see this as a blocking point, @richardwerkman ?I had another look, it basically requires changing the constructor signature for all high level classes (ProjectOrchestrator, ProjectMutator, MutationTestProcess, CSharpMutationProcess.
Or maybe hack way via MutationTestInput for limited damages...

richardwerkman:
@dupdob After discussing this some further it's not a blocking issue. Let's keep it in the options for now since it requires the least amount of code. Also your point is valid that this could become an option in the future. 

This is a sign that we should start looking into #1211 as that would make passing values like this one around possible.

PR #3000, opened 2024-07-26
fix: Disable debug trace

### LINKED ISSUE #2375: Suppress Assertion failed dialog when mutant uses invalid value verified by System.Diagnostics.Debug.Assert.
Mutating code that uses `System.Diagnostics.Debug.Assert( ... )` will show modal 'Assertion failed: abort / retry / ignore?' dialog  
example: string mutation on `string arg` will trigger dialog for `Debug.Assert(arg != "")`

No 'Assertion failed' dialogs shown.

1. remove all `Debug.Assert` calls from my code, but then i loose (my) sanity checks.
2. replace all `Debug.Assert(expr)` with `if(!expr) throw Exception()`, but then i will have useless checks in release code.

**My proposal solution**
1. Replace all `Debug.Assert(expr)` with `if(!expr) throw new Exception()` when mutating code. (This should also mark mutant as killed)

dupdob:
Make Stryker compatible with debug assertions logic, so that a mutant is considered as killed if it result in a failed assertion.

Disable `Debug.Assert` report dialogs. Failed assertions result in an argument exception raised, ensuring this are recognized as failed tests by Stryker.

Fixes #2375

Note: coverage reporting is wrong. For some reason, Sonar does not capture any coverage data for the DataCollector. I have added a dedicated test that does cover all testable lines.

### DIFF
--- a/integrationtest/TargetProjects/NetCore/NetCoreTestProject.NUnit/Constructs.cs
+++ b/integrationtest/TargetProjects/NetCore/NetCoreTestProject.NUnit/Constructs.cs
@@ -49,7 +49,7 @@ namespace NetCoreTestProject.NUnit
         [Test]
         [TestCaseSource(nameof(ObjectSource))]
         // all test cases refer to the same test
-        public void TestAgeIndirectObject(StrykerComments sut)
+        public void TestAgeIndirectObject(KilledMutants sut)
         {
             Console.WriteLine($"ObjectSource test:{sut.Age}");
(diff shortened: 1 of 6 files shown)


Liam-Rougoor (commented):
Thanks for the PR, it looks good. I do think this case could benefit from an integration test as well.

dupdob:
I added an integration test, see [here](https://github.com/stryker-mutator/stryker-net/blob/09243bc123b5461c8625f47b640afaf66d8b0ef1/integrationtest/TargetProjects/NetCore/NetCoreTestProject.XUnit/VariousTests.cs#L7) and adjusted test expectations accordingly.

PR #3024, opened 2024-08-30
feat(mutators): Add default parameter mutator

### LINKED ISSUE #2848: Mutator for default arguments.

Liam-Rougoor:
Closes #2848 
Remake of #2863


richardwerkman:
Can we add a test case for this in `CSharpMutantOrchestratorTests`? And/or add a mutation for this in the integration test project. So we're sure the mutations are placed correctly

Liam-Rougoor:
Don't forget to add the new mutator to the docs

PR #3028, opened 2024-09-03
feat(ignore mutation): Add support for ignoring mutations in code using mutiline string comments

### LINKED ISSUE #3008: Stryker comments inside multiline comments or at the ends of lines of code are ignored
**Describe the bug**
I have a method which returns null from an else-if block. Because Stryker has the method return `default` when the else-if block is removed, the behavior under that mutation is identical to the expected behavior (i.e., the method returns null), so the mutant can't be killed. This issue is described in #2000. In that issue @dupdob suggests using `// Stryker disable once block` to ignore the mutant. However, I ran into some challenges getting this to work with my company's coding style.

Because our style is to put curly braces on the same line as an if/else-keyword, I would like to be able to disable the Block mutation in this manner:
```csharp
  public string? ReturnDefaultFromElseIf(bool? value) {
    if (value == true) {
      return TrueStr;
    } else if (value == null) /* Stryker disable once Block */ {
      return null;
    } else {
      return FalseStr;
    }
  }
```

However, Stryker comments that appear in multiline comments, or that appear after/amidst code on the same line, appear to be disregarded. Additionally, if a "disable once" comment appears anywhere before the `if` keyword, then the mutant is disabled for the last else block as well (since its syntax is a child of the if-statement), unless I have a second comment to restore it. I made various attempts to get Stryker to do what I wanted, which you can read through in the source code below.

<details>
<summary>DisableBlockIssueDemo.cs</summary>

```csharp
namespace LibraryProject;

public class DisableBlockIssueDemo {
  private const string TrueStr = "True";
  private const string FalseStr = "False";

  /*
   * Here is a method similar to the one I am trying to work with.
   */

  public string? ReturnDefaultFromElseIf(bool? value) {
    if (value == true) {
      return TrueStr;
    } else if (value == null) {
      return null;
    } else {
      return FalseStr;
    }
  }

  /*
   * In general, for the examples below, my expectation is that Stryker would ignore
   * the Block mutation for the else-if block but not the else block.
   * (Though this expectation is stronger for some of these examples than others.)
   */

  // Does not work - Comment is out of scope.
  public string? ReturnDefaultFromElseIf1(bool? value) {
    if (value == true) {
      return TrueStr;
    // Stryker disable once Block
    } else if (value == null) {
      return null;
    } else {
      return FalseStr;
    }
  }

  // Does not work - Stryker ignores multiline comments.
  public string? ReturnDefaultFromElseIf2(bool? value) {
    if (value == true) {
      return TrueStr;
    } /* Stryker disable once Block */ else if (value == null) {
      return null;
    } else {
      return FalseStr;
    }
  }

  // Does not work - Stryker ignores multiline comments.
  public string? ReturnDefaultFromElseIf3(bool? value) {
    if (value == true) {
      return TrueStr;
    } else /* Stryker disable once Block */ if (value == null) {
      return null;
    } else {
      return FalseStr;
    }
  }

  // Does not work - Stryker ignores multiline comments.
  // (This is the approach that I think would be the easiest to have Stryker work with, while being reasonably readable.)
  public string? ReturnDefaultFromElseIf4(bool? value) {
    if (value == true) {
      return TrueStr;
    } else if (value == null) /* Stryker disable once Block */ {
      return null;
    } else {
      return FalseStr;
    }
  }

  // Does not work - Comment appears after start of block.
  public string? ReturnDefaultFromElseIf5(bool? value) {
    if (value == true) {
      return TrueStr;
    } else if (value == null) { // Stryker disable once Block
      return null;
    } else {
      return FalseStr;
    }
  }

  // Does not work - Stryker ignores comments in trailing trivia (see Visual Studio's Syntax Visualizer).
  public string? ReturnDefaultFromElseIf6(bool? value) {
    if (value == true) {
      return TrueStr;
    } else if (value == null) // Stryker disable once Block
    {
      return null;
    } else {
      return FalseStr;
    }
  }

  // WORKS, but separates the curly brace from the if condition.
  public string? ReturnDefaultFromElseIf7(bool? value) {
    if (value == true) {
      return TrueStr;
    } else if (value == null)
    // Stryker disable once Block
    {
      return null;
    } else {
      return FalseStr;
    }
  }

  // Does not work - Stryker disables the mutant for both the else-if block and the else block.
  // (This is because it is a child of the if-statement; see Visual Studio's Syntax Visualizer)
  public string? ReturnDefaultFromElseIf8(bool? value) {
    if (value == true) {
      return TrueStr;
    } else
    // Stryker disable once Block
    if (value == null) {
      return null;
    } else {
      return FalseStr;
    }
  }

  // Does not work - Stryker ignores comments in trailing trivia.
  public string? ReturnDefaultFromElseIf9(bool? value) {
    if (value == true) {
      return TrueStr;
    } // Stryker disable once Block
    else if (value == null) {
      return null;
    } else {
      return FalseStr;
    }
  }

  // Does not work - Stryker disables the mutant for both the else-if block and the else block.
  public string? ReturnDefaultFromElseIf10(bool? value) {
    if (value == true) {
      return TrueStr;
    }
    // Stryker disable once Block
    else if (value == null) {
      return null;
    } else {
      return FalseStr;
    }
  }

  /*
   * The examples below try to mitigate the fact that Stryker still
   * disables the mutation for the else block if the disable comment
   * appears before the "if" keyword of the else-if portion.
   */

  // Does not work - Restore comment is outside the scope of the else block.
  public string? ReturnDefaultFromElseIf11(bool? value) {
    if (value == true) {
      return TrueStr;
    }
    // Stryker disable once Block
    else if (value == null) {
      return null;
    // Stryker restore Block
    } else {
      return FalseStr;
    }
  }

  // Does not work - Stryker ignores comments in trailing trivia.
  public string? ReturnDefaultFromElseIf12(bool? value) {
    if (value == true) {
      return TrueStr;
    }
    // Stryker disable once Block
    else if (value == null) {
      return null;
    } // Stryker restore Block
    else {
      return FalseStr;
    }
  }

  // WORKS, but separates the else keywords from the curly braces.
  // (This is the least ugly solution I've found so far.)
  public string? ReturnDefaultFromElseIf13(bool? value) {
    if (value == true) {
      return TrueStr;
    }
    // Stryker disable once Block
    else if (value == null) {
      return null;
    }
    // Stryker restore Block
    else {
      return FalseStr;
    }
  }
}
```

</details>

<details>
<summary>DisableBlockIssueDemoTests.cs</summary>

```csharp
namespace LibraryProject.Test;

public class DisableBlockIssueDemoTests {
  private DisableBlockIssueDemo exampleClass = new DisableBlockIssueDemo();

  public static IEnumerable<object?[]> ReturnDefaultFromElseIfInputs => [
    [true, "True"],
    //[false, "False"], // For demonstration purposes: Disabled so that mutations on else-branches survive.
    [null, null]
  ];

  [Theory, MemberData(nameof(ReturnDefaultFromElseIfInputs))]
  public void ReturnDefaultFromElseIf_ReturnsCorrectValue(bool? input, string? expectedOutput) {
    Assert.Equal(expectedOutput, exampleClass.ReturnDefaultFromElseIf(input));
  }

  [Theory, MemberData(nameof(ReturnDefaultFromElseIfInputs))]
  public void ReturnDefaultFromElseIf1_ReturnsCorrectValue(bool? input, string? expectedOutput) {
    Assert.Equal(expectedOutput, exampleClass.ReturnDefaultFromElseIf1(input));
  }

  [Theory, MemberData(nameof(ReturnDefaultFromElseIfInputs))]
  public void ReturnDefaultFromElseIf2_ReturnsCorrectValue(bool? input, string? expectedOutput) {
    Assert.Equal(expectedOutput, exampleClass.ReturnDefaultFromElseIf2(input));
  }

  [Theory, MemberData(nameof(ReturnDefaultFromElseIfInputs))]
  public void ReturnDefaultFromElseIf3_ReturnsCorrectValue(bool? input, string? expectedOutput) {
    Assert.Equal(expectedOutput, exampleClass.ReturnDefaultFromElseIf3(input));
  }

  [Theory, MemberData(nameof(ReturnDefaultFromElseIfInputs))]
  public void ReturnDefaultFromElseIf4_ReturnsCorrectValue(bool? input, string? expectedOutput) {
    Assert.Equal(expectedOutput, exampleClass.ReturnDefaultFromElseIf4(input));
  }

  [Theory, MemberData(nameof(ReturnDefaultFromElseIfInputs))]
  public void ReturnDefaultFromElseIf5_ReturnsCorrectValue(bool? input, string? expectedOutput) {
    Assert.Equal(expectedOutput, exampleClass.ReturnDefaultFromElseIf5(input));
  }

  [Theory, MemberData(nameof(ReturnDefaultFromElseIfInputs))]
  public void ReturnDefaultFromElseIf6_ReturnsCorrectValue(bool? input, string? expectedOutput) {
    Assert.Equal(expectedOutput, exampleClass.ReturnDefaultFromElseIf6(input));
  }

  [Theory, MemberData(nameof(ReturnDefaultFromElseIfInputs))]
  public void ReturnDefaultFromElseIf7_ReturnsCorrectValue(bool? input, string? expectedOutput) {
    Assert.Equal(expectedOutput, exampleClass.ReturnDefaultFromElseIf7(input));
  }

  [Theory, MemberData(nameof(ReturnDefaultFromElseIfInputs))]
  public void ReturnDefaultFromElseIf8_ReturnsCorrectValue(bool? input, string? expectedOutput) {
    Assert.Equal(expectedOutput, exampleClass.ReturnDefaultFromElseIf8(input));
  }

  [Theory, MemberData(nameof(ReturnDefaultFromElseIfInputs))]
  public void ReturnDefaultFromElseIf9_ReturnsCorrectValue(bool? input, string? expectedOutput) {
    Assert.Equal(expectedOutput, exampleClass.ReturnDefaultFromElseIf9(input));
  }

  [Theory, MemberData(nameof(ReturnDefaultFromElseIfInputs))]
  public void ReturnDefaultFromElseIf10_ReturnsCorrectValue(bool? input, string? expectedOutput) {
    Assert.Equal(expectedOutput, exampleClass.ReturnDefaultFromElseIf10(input));
  }

  [Theory, MemberData(nameof(ReturnDefaultFromElseIfInputs))]
  public void ReturnDefaultFromElseIf11_ReturnsCorrectValue(bool? input, string? expectedOutput) {
    Assert.Equal(expectedOutput, exampleClass.ReturnDefaultFromElseIf11(input));
  }

  [Theory, MemberData(nameof(ReturnDefaultFromElseIfInputs))]
  public void ReturnDefaultFromElseIf12_ReturnsCorrectValue(bool? input, string? expectedOutput) {
    Assert.Equal(expectedOutput, exampleClass.ReturnDefaultFromElseIf12(input));
  }

  [Theory, MemberData(nameof(ReturnDefaultFromElseIfInputs))]
  public void ReturnDefaultFromElseIf13_ReturnsCorrectValue(bool? input, string? expectedOutput) {
    Assert.Equal(expectedOutput, exampleClass.ReturnDefaultFromElseIf13(input));
  }
}
```

</details>

**Logs**
[log-20240808.txt](https://github.com/user-attachments/files/16552193/log-20240808.txt)

How Stryker treats a couple of the approaches shown in the source code above:

image

image

**Expected behavior**
Stryker comments that appear in multiline comments, or that appear after some code on the same line, should be applied the same as comments that appear on a line of their own.

**Desktop:**
 - OS: Windows
 - Type of project: Microsoft.NET.Sdk
 - Framework Version: net8.0
 - Stryker Version: 4.1.0

**Additional context**
Looking at the Stryker source code, the issue appears to be two-fold.

First, while multi-line comments [are included in the comment search](https://github.com/stryker-mutator/stryker-net/blob/e12cd88551e9b61fa2bc34900e655581f848d272/src/Stryker.Core/Stryker.Core/Mutants/CsharpNodeOrchestrators/CommentParser.cs#L72), it is impossible for them to match [the regex pattern](https://github.com/stryker-mutator/stryker-net/blob/e12cd88551e9b61fa2bc34900e655581f848d272/src/Stryker.Core/Stryker.Core/Mutants/CsharpNodeOrchestrators/CommentParser.cs#L14) that is used to detect Stryker comments, since that pattern only accepts comments that start with `//`.

Second, [only comments from leading trivia are considered](https://github.com/stryker-mutator/stryker-net/blob/e12cd88551e9b61fa2bc34900e655581f848d272/src/Stryker.Core/Stryker.Core/Mutants/CsharpNodeOrchestrators/CommentParser.cs#L70). As such, any comments that appear as trailing trivia (e.g., amid code or at the end of a line after some code) will not be included. In theory, I think it should be possible to include the trailing trivia of the prior SyntaxToken in addition to the leading trivia of the SyntaxNode in question (see [the implementation of SyntaxNode.GetLeadingTrivia](https://github.com/dotnet/roslyn/blob/43b0b05cc4f492fd5de00f6f6717409091df8daa/src/Compilers/Core/Portable/Syntax/SyntaxNode.cs#L1063C13-L1063C72), [SyntaxNode.GetFirstToken](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.syntaxnode.getfirsttoken?view=roslyn-dotnet-4.7.0), and [SyntaxToken.GetPreviousToken](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.syntaxtoken.getprevioustoken?view=roslyn-dotnet-4.7.0)).

It sounds like at some point [there was a deliberate decision to not support multiline comments](https://github.com/stryker-mutator/stryker-net/issues/1570#issuecomment-850263388). However, it sounds like that decision was in part due to the trailing/leading trivia issue, which I believe is surmountable (see previous paragraph).

dupdob:
Add support for the so called multiline comments format (`/* comment */ `) for Stryker comments.
Updated the doc and provided some detailed explanation for `once` option. Note that the Stryker comment must still be on a single line.

Multiline comments should work exactly line single line comments when put on a dedicated line.

Stryker attempts to associate comments within the line with the **previous** syntax node, that is:

`if (condition) {/* Stryker comment*/;....}`: this comment will apply to the whole syntax block
`if (/* Stryker comment*/ condition) {....}`: this comment will apply to the whole condition expression

Stryker comments may not be detected if they are within a syntax node
`if (cond/* Stryker comment*/ition) {....}`: will have no effect
`if (condition /* Stryker comment*/) {....}`: will have no effect

Fixes #3008

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
+++ b/src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
@@ -197,29 +197,5 @@ namespace Stryker.Core.Helpers
                 return (child.Parent is not AnonymousFunctionExpressionSyntax function || function.ExpressionBody != child)
                        && (child.Parent is not LocalFunctionStatementSyntax localFunction || localFunction.ExpressionBody != child);
             } ).Any(predicate);
-
-        /// <summary>
-        /// Find the syntax node or token that is just before the given node.
-        /// </summary>
-        /// <param name="node"></param>
-        /// <returns>previous node or token. Null if non exists.</returns>
-        public static SyntaxNodeOrToken GetPreviousSyntaxNode(this SyntaxNode node)
-        {
-            SyntaxNodeOrToken previous = node.Parent;
-            if (previous == null)
-            {
-                return null;
-            }
-
-            foreach(var child in previous.ChildNodesAndTokens())
-            {
-                if (child == node)
-                {
-                    return previous;
-                }
-                previous = child;
-            }
-            return null;
-        }
     }
 }
--- a/src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
+++ b/src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
@@ -198,13 +198,28 @@ namespace Stryker.Core.Helpers
                        && (child.Parent is not LocalFunctionStatementSyntax localFunction || localFunction.ExpressionBody != child);
             } ).Any(predicate);
 
-        public static bool HasAMemberBindingExpression(this ExpressionSyntax expression) =>
-            expression switch
+        /// <summary>
+        /// Find the syntax node or token that is just before the given node.
+        /// </summary>
+        /// <param name="node"></param>
+        /// <returns>previous node or token. Null if non exists.</returns>
+        public static SyntaxNodeOrToken GetPreviousSyntaxNode(this SyntaxNode node)
+        {
+            SyntaxNodeOrToken previous = node.Parent;
+            if (previous == null)
             {
-                MemberBindingExpressionSyntax => true,
-                MemberAccessExpressionSyntax memberAccess => memberAccess.Expression.HasAMemberBindingExpression(),
-                InvocationExpressionSyntax invocation => invocation.Expression.HasAMemberBindingExpression(),
-                _ => false
-            };
+                return null;
+            }
+
+            foreach(var child in previous.ChildNodesAndTokens())
+            {
+                if (child == node)
+                {
+                    return previous;
+                }
+                previous = child;
+            }
+            return null;
+        }
     }
 }
--- a/src/Stryker.Core/Stryker.Core/Mutants/CsharpNodeOrchestrators/CommentParser.cs
+++ b/src/Stryker.Core/Stryker.Core/Mutants/CsharpNodeOrchestrators/CommentParser.cs
@@ -11,8 +11,8 @@ namespace Stryker.Core.Mutants.CsharpNodeOrchestrators;
 
 internal static class CommentParser
 {
-    private static readonly Regex Pattern = new("^\\s*\\/\\/\\s*Stryker", RegexOptions.Compiled | RegexOptions.IgnoreCase, TimeSpan.FromMilliseconds(20));
-    private static readonly Regex Parser = new("^\\s*\\/\\/\\s*Stryker\\s*(disable|restore)\\s*(once|)\\s*([^:]*)\\s*:?(.*)$", RegexOptions.Compiled | RegexOptions.IgnoreCase, TimeSpan.FromMilliseconds(20));
+    private static readonly Regex Pattern = new("^\\s*(\\/\\/\\s*Stryker(.*))|(\\/\\*\\s*Stryker(.*[^\\*][^\\\\])\\*\\/\\s*)$", RegexOptions.Compiled | RegexOptions.IgnoreCase, TimeSpan.FromMilliseconds(20));
+    private static readonly Regex Parser = new("^\\s*(disable|restore)\\s*(once|)\\s*([^:]*)\\s*:?(.*)$", RegexOptions.Compiled | RegexOptions.IgnoreCase, TimeSpan.FromMilliseconds(20));
     private static readonly ILogger Logger = ApplicationLogging.LoggerFactory.CreateLogger("CommentParser");
 
     private static MutationContext ParseStrykerComment(MutationContext context, Match match, SyntaxNode node)
@@ -72,12 +72,17 @@ internal static class CommentParser
                                  t.IsKind(SyntaxKind.MultiLineCommentTrivia)).Select(t => t.ToString()))
         {
             // perform a quick pattern check to see if it is a 'Stryker comment'
-            if (!Pattern.Match(commentTrivia).Success)
+            var strykerCommentMatch = Pattern.Match(commentTrivia);
+            if (!strykerCommentMatch.Success)
             {
                 continue;
             }
 
-            var match = Parser.Match(commentTrivia);
+            // now we can extract actual command
+            var isSingleLine = strykerCommentMatch.Groups[1].Success;
+            var command = isSingleLine ? strykerCommentMatch.Groups[2].Value : strykerCommentMatch.Groups[4].Value;
+
+            var match = Parser.Match(command);
             if (match.Success)
             {
                 // this is a Stryker comments, now we parse it
--- a/src/Stryker.Core/Stryker.Core/Mutants/CsharpNodeOrchestrators/CommentParser.cs
+++ b/src/Stryker.Core/Stryker.Core/Mutants/CsharpNodeOrchestrators/CommentParser.cs
@@ -65,7 +65,7 @@ internal static class CommentParser
         return context.FilterMutators(disable, filteredMutators, match.Groups[OnceGroup].Value.ToLower() == "once", comment);
     }
 
-    public static MutationContext ParseNodeComments(SyntaxNode node, MutationContext context)
+    public static MutationContext ParseNodeLeadingComments(SyntaxNode node, MutationContext context)
     {
         foreach (var commentTrivia in node.GetLeadingTrivia()
                      .Where(t => t.IsKind(SyntaxKind.SingleLineCommentTrivia) ||
--- a/src/Stryker.Core/Stryker.Core/Mutants/CsharpNodeOrchestrators/CommentParser.cs
+++ b/src/Stryker.Core/Stryker.Core/Mutants/CsharpNodeOrchestrators/CommentParser.cs
@@ -1,10 +1,10 @@
 using System;
 using System.Linq;
+using System.Text;
 using System.Text.RegularExpressions;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.Extensions.Logging;
-using Stryker.Core.Helpers;
 using Stryker.Core.Logging;
 using Stryker.Core.Mutators;
 
@@ -12,8 +12,8 @@ namespace Stryker.Core.Mutants.CsharpNodeOrchestrators;
 
 internal static class CommentParser
 {
-    private static readonly Regex Pattern = new("^\\s*(\\/\\/\\s*Stryker(.*))|(\\/\\*\\s*Stryker(.*[^\\*][^\\\\])\\*\\/\\s*)$", RegexOptions.Compiled | RegexOptions.IgnoreCase, TimeSpan.FromMilliseconds(20));
-    private static readonly Regex Parser = new("^\\s*(disable|restore)\\s*(once|)\\s*([^:]*)\\s*:?(.*)$", RegexOptions.Compiled | RegexOptions.IgnoreCase, TimeSpan.FromMilliseconds(20));
+    private static readonly Regex Pattern = new("^\\s*(\\/\\/\\s*Stryker(.*))|(\\/\\*\\s*Stryker(.*[^\\*][^\\\\])\\*\\/\\s*)$", RegexOptions.Compiled | RegexOptions.IgnoreCase, TimeSpan.FromMilliseconds(200));
+    private static readonly Regex Parser = new("^\\s*(disable|restore)\\s*(once|)\\s*([^:]*)\\s*:?(.*)$", RegexOptions.Compiled | RegexOptions.IgnoreCase, TimeSpan.FromMilliseconds(200));
     private static readonly ILogger Logger = ApplicationLogging.LoggerFactory.CreateLogger("CommentParser");
 
     private static MutationContext ParseStrykerComment(MutationContext context, Match match, SyntaxNode node)
@@ -68,28 +68,24 @@ internal static class CommentParser
 
     public static MutationContext ParseNodeLeadingComments(SyntaxNode node, MutationContext context)
     {
-        // parse comment that is on a dedicated line before the syntax node
-        var firstResult = node.GetLeadingTrivia()
-            .Where(t => t.IsKind(SyntaxKind.SingleLineCommentTrivia) ||
-                        t.IsKind(SyntaxKind.MultiLineCommentTrivia))
-            .Select(t => ProcessComment(node, context, t.ToString()))
-            .FirstOrDefault( t => t!= null);
+        var processedComments = node.GetFirstToken(true).GetPreviousToken(true).TrailingTrivia
+            .Union(node.GetLeadingTrivia())
+            .Where(t => t.IsKind(SyntaxKind.MultiLineCommentTrivia) || t.IsKind(SyntaxKind.SingleLineCommentTrivia))
+            .Select(t => (ProcessComment(node, context, t.ToString()), t)).Where(t => t.Item1 != null).ToList();
 
-        if (firstResult != null)
-        {
-            return firstResult;
-        }
-        // identify the previous syntax node if possible
-        var previousSyntaxNode = node.GetPreviousSyntaxNode();
-        if (previousSyntaxNode == null)
+        if (processedComments.Count == 0)
         {
             return context;
         }
-        // parse any trailing comment, assuming the user intend it to apply to this node.
-        return previousSyntaxNode.GetTrailingTrivia()
-            .Where(t => t.IsKind(SyntaxKind.MultiLineCommentTrivia))
-            .Select(t => ProcessComment(node, context, t.ToString()))
-            .FirstOrDefault( t => t!= null) ?? context;
+        if (processedComments.Count > 1)
+        {
+            var errorMessage = new StringBuilder().Append(  
+                $"Multiple Stryker comments at {node.GetLocation().GetMappedLineSpan().StartLinePosition}, {node.SyntaxTree.FilePath}. Only the first one will be used");
+            errorMessage.Append(string.Join(Environment.NewLine, processedComments.Select(c => c.Item2.ToString())));
+            Logger.LogWarning(errorMessage.ToString());
+
+        }
+        return processedComments[0].Item1;
     }
 
     private static MutationContext ProcessComment(SyntaxNode node, MutationContext context, string commentTrivia)
--- a/src/Stryker.Core/Stryker.Core/Mutants/CsharpNodeOrchestrators/CommentParser.cs
+++ b/src/Stryker.Core/Stryker.Core/Mutants/CsharpNodeOrchestrators/CommentParser.cs
@@ -68,14 +68,17 @@ internal static class CommentParser
 
     public static MutationContext ParseNodeLeadingComments(SyntaxNode node, MutationContext context)
     {
-        // parse comment that is on a dedicated line before the syntx node
-        if (node.GetLeadingTrivia()
+        // parse comment that is on a dedicated line before the syntax node
+        var firstResult = node.GetLeadingTrivia()
             .Where(t => t.IsKind(SyntaxKind.SingleLineCommentTrivia) ||
-                        t.IsKind(SyntaxKind.MultiLineCommentTrivia)).Select(t => t.ToString()).Any(commentTrivia => ProcessComment(node, ref context, commentTrivia)))
-        {
-            return context;
-        }
+                        t.IsKind(SyntaxKind.MultiLineCommentTrivia))
+            .Select(t => ProcessComment(node, context, t.ToString()))
+            .FirstOrDefault( t => t!= null);
 
+        if (firstResult != null)
+        {
+            return firstResult;
+        }
         // identify the previous syntax node if possible
         var previousSyntaxNode = node.GetPreviousSyntaxNode();
         if (previousSyntaxNode == null)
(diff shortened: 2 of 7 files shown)


src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
```diff
@@ -198,13 +198,28 @@ public static bool ContainsNodeThatVerifies(this SyntaxNode node, Func<SyntaxNod
...
+            foreach(var child in previous.ChildNodesAndTokens())
+            {
+                if (child == node)
+                {
+                    return previous;
+                }
+                previous = child;
+            }
+            return null;
+        }
```
nwalker-et, line 223: When I wrote the original issue, I had imagined that something like `node.GetFirstToken(includeZeroWidth: true).GetPreviousToken(includeZeroWidth: true)` would suffice for a method like this. What led you to go with this logic instead?

src/Stryker.Core/Stryker.Core/Mutants/CsharpNodeOrchestrators/CommentParser.cs
```diff
@@ -65,32 +66,55 @@ private static MutationContext ParseStrykerComment(MutationContext context, Matc
...
-            }
-
-            var match = Parser.Match(commentTrivia);
-            if (match.Success)
-            {
-                // this is a Stryker comments, now we parse it
-                context = ParseStrykerComment(context, match, node);
-                break;
-            }
+            return context;
```
nwalker-et, line 76: I notice that this skips the rest of the logic if any leading trivia are present. Is that intentional? Would we want to allow for cases where, for whatever reason, someone is using both trailing trivia on the previous token and leading trivia on the current token?

This would certainly be a strange way for someone to write their code, but I would still expect Stryker's comment parsing logic to handle it as one would expect, unless supporting it would break other stuff:
```csharp
public bool DoSomeLogic(int a, int b)
{   // Stryker disable arithmetic
    // Stryker disable equality
    return a + b < 10;
}
```

src/Stryker.Core/Stryker.Core/Mutants/CsharpNodeOrchestrators/CommentParser.cs
```diff
@@ -65,32 +66,55 @@ private static MutationContext ParseStrykerComment(MutationContext context, Matc
...
-                node.SyntaxTree.FilePath);
+        // identify the previous syntax node if possible
+        var previousSyntaxNode = node.GetPreviousSyntaxNode();
+        if (previousSyntaxNode == null)
+        {
+            return context;
         }
+        // parse any trailing comment, assuming the user intend it to apply to this node.
+        previousSyntaxNode.GetTrailingTrivia()
+            .Where(t => t.IsKind(SyntaxKind.MultiLineCommentTrivia)).Select(t => t.ToString()).
```
nwalker-et, line 87: Should we include single-line comments here as well, or is there a reason we wouldn't want to support that scenario?

nwalker-et:
I'm also having a hard time getting the new tests to pass when I run them locally, though it seems like they're passing in the CI/CD pipeline. Is there something I'm missing in my local setup? In case it helps, I'm getting errors like the following:
```
  Error Message:
   Test method Stryker.Core.UnitTest.Mutants.CsharpMutantOrchestratorTests.ShouldNotMutateDependingOnWhereMultilineCommentIs threw exception:
Shouldly.ShouldAssertException: isSame
    should be
True
    but was
False

Additional Info:
    AST's are not equivalent. Line[9]
actual: if (/* Stryker disable once all*/cond) if(StrykerNamespace.MutantControl.IsActive(2)){;}else{if(StrykerNamespace.MutantControl.IsActive(3)){x--;
expect:if (/* Stryker disable once all*/cond) if(StrykerNamespace.MutantControl.IsActive(0)){;}else{if(StrykerNamespace.MutantControl.IsActive(0)){x--;}else{x++;}}
Actual(full):
using System;
using System.Text;
...
```

src/Stryker.Core/Stryker.Core/Mutants/CsharpNodeOrchestrators/CommentParser.cs
dupdob: Indeed, having multiple Stryker comments is currently undefined behavior. 
There is currently no scenario where it is needed. One can provide a list of mutators if one needs to disable or enable more than one. That is, you can do this:
`// Stryker disable arithmetic,equality`

Support for multi line comments offers new way to haves several Stryker comments for the same syntax node. But it does not provide any valid reason to do so.

Current behavior is that Stryker stops processing after the first valid Stryker comment.
Possible behaviors are:
1. Aggregate all Stryker comments, with the following logic
   1. Sequentially, that is each comment line is processed in order; command can cancel each other (if you have both enable and disable commands)
   2. With semantic ordering: there is a prioritization between enabling and disabling (prioritization TBD) 
1. Favor only one Stryker comment
   1. The first one 
   3. The last one
   4. The most specific
   That being said, Stryker could log a warning regarding the discarded comments
1. Rejects all Stryker comments and log a warning 

Bearing in mind that there is valid use cases for having multiple Stryker comments (for a single Syntax Node), the logical conclusion is to maintain the current logic but with an added warning in the log file for information.

The single real world situation (I can think of) where more than one Stryker comment is present is as a consequence of a code merge.
 
Aggregation requires implementing the aggregation logic (sequentially processing the command will not provide the expected result in the general case).
Rejecting all Stryker comments looks a bit drastic.

If valid use cases for multiple comments are identified, I am pretty sure the appropriate logic will be identified as well.

src/Stryker.Core/Stryker.Core/Mutants/CsharpNodeOrchestrators/CommentParser.cs
dupdob: thanks for the remark, it is just that I wrongly assumed that single line comment could not be trailing (which is a stupid idea in hindsight).

dupdob:
> I'm also having a hard time getting the new tests to pass when I run them locally, though it seems like they're passing in the CI/CD pipeline. Is there something I'm missing in my local setup? In case it helps, I'm getting errors like the following:
> 
> ```
>   Failed ShouldNotMutateIfDisabledByMultilineComment [14 ms]
>   Error Message:
>    Test method Stryker.Core.UnitTest.Mutants.CsharpMutantOrchestratorTests.ShouldNotMutateIfDisabledByMultilineComment threw exception:
> Shouldly.ShouldAssertException: isSame
>     should be
> True
>     but was
> False
> 
> Additional Info:
>     AST's are not equivalent. Line[9]
> actual:    if ((StrykerNamespace.MutantControl.IsActive(2)?!(condition && other):(StrykerNamespace.MutantControl.IsActive(1)?condition || other:condition && other))) /* Stryker disable once all */ {
> expect:    if ((StrykerNamespace.MutantControl.IsActive(0)?!(condition && other):(StrykerNamespace.MutantControl.IsActive(0)?condition || other:condition && other))) /* Stryker disable once all */ {
> Actual(full):
> using System;
> using System.Collections.Generic;
> using System.Text;
> ...
> ```

You should pull again the code. I usually amend commits to fix test errors and force push the PR. I do not expect people to pull the code until the factory is green

src/Stryker.Core/Stryker.Core/Mutants/CsharpNodeOrchestrators/CommentParser.cs
nwalker-et: Understood, thank you for the clarification. I think the warnings you introduced add a lot of clarity to the current implementation.

Perhaps a more realistic scenario where multiple Stryker comments might be used would be if someone needs to use a combination of `once` and non-`once` comments, as they can't be merged together (at least as far as I'm aware). For example:
```csharp
int SomeMethod(bool flagA, bool flagB, int y, int z) {
  // Stryker disable Arithmetic
  // Stryker disable once all
  var x = (flagA && flagB) ? 10 : (y + z); // Should have all mutations disabled.
  return x + 5; // Should only have Arithmetic mutations disabled.
}
```

If that's a scenario we want to support, perhaps we should only allow at most a single non-`once` comment and a single `once` comment on any given node. It might also make sense to require the `once` comment to come *after* the non-`once` comment, both for clarity and to avoid scope-related issues.

src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
nwalker-et: @dupdob I saw that you incorporated the approach I mentioned above. I assume it works as expected then? I came up with that approach after reaching some of Roslyn's documentation and source code, but I haven't had the chance to test it myself yet, so I just wanted to make sure you weren't simply taking my word for it.

nwalker-et:
> You should pull again the code. I usually amend commits to fix test errors and force push the PR. I do not expect people to pull the code until the factory is green and the PR is ready (and not draft).

Thanks, that makes sense! I was just wanting to test out some of my own suggestions, so I'll keep that in mind if I run into the issue again.

> Stryker comments may not be detected if they are within a syntax node
> `if (cond/* Stryker comment*/ition) {....}`: will have no effect
> `if (condition) /* Stryker comment*/ {....}`: will have no effect

The wording here might be worth clarifying. In most cases, if a Stryker comment is within a syntax node, it may not apply to *that* syntax node, but it will often apply to at least one of that syntax node's *descendants*. In your first example, it might only apply to part of the condition expression, and perhaps not the part one would expect (unless it came before a parenthesized expression). In the second example, it wouldn't apply to the if-statement, but it *would* apply to the if-statement's block, since that is its own syntax node.

<details>
<summary>More nuanced thoughts on when Stryker comments might not be detected at all...</summary>
<br/>

With the new logic in this PR, I believe a Stryker comment would only go *undetected* if either a) it is leading trivia for a token that does not appear first within its parent syntax node, or b) it is trailing trivia for a token that is not immediately followed by the start of a new syntax node. So, for example, the Stryker comments in the following code would not be detected at all:

```csharp
// Condition a)
SomeMethod(
  true
  /* Stryker disable all */,
  10);
// The Stryker comment is leading trivia for the `,` token, which is a middle child of the ArgumentList syntax node.

// Condition b)
SomeMethod(true, 10 /* Stryker disable all */);
// The Stryker comment is trailing trivia for the `10` token, and the token immediately after it, `)`, is not the start of a syntax node.
```

In either case, the comment would have no effect, and in my opinion, this would be the expected behavior given how Stryker currently handles scope. I haven't thought of any scenarios where one would expect different behavior when at least one of the two conditions I mentioned holds true, so I think the logic in this PR should work just fine.
</details>

src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs
dupdob: It looks like it gives the same result. I tried because it allowed me to remove the `FindPreviousNode` implementation.
I glanced at the implementation of `GetPreviousToken` and it looks overkill and a bit inefficient for this use case, but it should be fast enough.

src/Stryker.Core/Stryker.Core/Mutants/CsharpNodeOrchestrators/CommentParser.cs
dupdob: For the time being, let's keep it simple. There are several ways to achieve this granularity without using multiple Stryker comments (on the same construct).

Since you want to disable arithmetic for the whole method, why not place it at the method level
```csharp
// Stryker disable Arithmetic
int SomeMethod(bool flagA, bool flagB, int y, int z) {
  // Stryker disable once all
  var x = (flagA && flagB) ? 10 : (y + z); // Should have all mutations disabled.
  return x + 5; // Should only have Arithmetic mutations disabled.
}
```
You could also place it at the block level using multiline comments.
Alternatively, only the first line requires specific handling, so you can disable arithmetic afterward:
```csharp
int SomeMethod(bool flagA, bool flagB, int y, int z) {
  // Stryker disable once all
  var x = (flagA && flagB) ? 10 : (y + z); // Should have all mutations disabled.
 // Stryker disable Arithmetic
  return x + 5; // Should only have Arithmetic mutations disabled.
}
```

or even inject an empty statement
```csharp
int SomeMethod(bool flagA, bool flagB, int y, int z) {
  /* Stryker disable Arithmetic */;
  // Stryker disable once all
  var x = (flagA && flagB) ? 10 : (y + z); // Should have all mutations disabled.
  return x + 5; // Should only have Arithmetic mutations disabled.
}
```
which is ugly, but does the job if you cannot use an alternative approach.

Regarding command aggregation, note that people could argue for that interpretation
```csharp
  // Stryker disable all
  // Stryker disable Arithmetic all
  here, only arithmetic mutations are ignored, the others are handled normally
  ```
 
The intent of Stryker comments is to allow user to manually ignore mutations that are semantically equivalent to the original code, or can not be tested due to some project constraint.
This way, they can sill target a 100% kill rate. But the objective never was to support surgical precision, that would allow users to target the exact mutations they want to disable.
Coder should favor the `disable once all` in general most of the time, using specific mutator selection when they do need other mutations to be tested. 
And it should be avoided and used as rarely as possible. When a codebase has a lot of Stryker comments, it will be often easier to disable some mutations than it is to kill it and then the interest of mutation testing is lost.

A feature being implementable and consistent with the existing design is not sufficient to make it a good idea 😄 .

Of course, this is only my opinion, and this can be discussed. But, as a last thing, I would add we can wait a few months to see what are the real use cases for this, if any arises

dupdob:
> Stryker comments may not be detected if they are within a syntax node
 > `if (cond/* Stryker comment*/ition) {....}`: will have no effect
 > `if (condition) /* Stryker comment*/ {....}`: will have no effect

This doc is wrong here (I fixed it in a recent commit). I wrote it before doing any discovery testing: 
`if (condition) /* Stryker comment*/ {....}` actually has an effect (on the if true block or statement)

> The wording here might be worth clarifying....

Yes but no. 
```csharp
SomeMethod(true
  /* Stryker disable all */, 
  "");
  ```
  No effect here, the empty string will be mutated
  ```csharp
SomeMethod(true,
  /* Stryker disable all */
  "");
  ```
  Effect here, the empty string mutation will be ignored. Simply by moving the comma.

You and I and people familiar with AST in general and more specifically Roslyn will find this logical, but most people may just see a code style difference.
My view:
1. Documenting this precisely is tricky. As in, the only people who will get what it means are the less likely to need it
2. In practice, this will be indeed a rare occurrence
3. People rarely read the doc
4. Documenting it makes it a feature, while keeping it non documented allows behavior to be adjusted and improved in the future, without being a breaking change

PR #3041, opened 2024-09-19
fix: Reference Aliases

Pentiva:
Currently stryker fails to mutate Stryker.Core as a recent change (#3033) required applying an alias to Stryker.Regex.Parser to avoid conflicts. When running stryker against itself (e.g https://dev.azure.com/stryker-mutator/Stryker/_build/results?buildId=19584&view=logs&j=6d77c950-6c97-5c0c-ab3e-a9679c1574a4&t=6be46d81-717b-5e63-da84-91412b35485c) the alias isn't respected and causes a compilation error. This is because Buildalyzer doesn't forward that information in any reasonable way that I can see. You can see on [this line](https://github.com/phmonte/Buildalyzer/blob/5c55d2cd7b17a6630275d3bb91ffb7a1465a430a/src/Buildalyzer/Compiler/Compiler.cs#L104) that only the reference is passed along to what will be `IAnalyzerResult.References` and that the aliases for `MetadataReferences` are not used.

My fix is very bad, but works. Feel free to discard this and implement a better solution. (The correct fix would probably be to create a pr for BuildAnalyser that exposes the aliases in some way)

I have also added code comments with links to the `/reference` command line switch information I found in my search.

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultExtensions.cs
+++ b/src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultExtensions.cs
@@ -104,18 +104,31 @@ public static class IAnalyzerResultExtensions
 
     public static IEnumerable<MetadataReference> LoadReferences(this IAnalyzerResult analyzerResult)
     {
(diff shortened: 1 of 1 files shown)


rouke-broersma:
Thanks for this, I was wondering why the pipeline was failing all of a sudden but had not had time to investigate!

I think for now the quickest win would be to change the alias configuration in Stryker, and then come up with the proper fix for this. #3042 

We should support this in that format and we can go from there.

Pentiva:
I didn't realize that it being an attribute vs a child element made a difference. That is definitely the better quick fix.

PR #3045, opened 2024-09-25
feat: Add --skip-version-check flag to disable automatic version checks

### LINKED ISSUE #3044: Allow skipping the version check
Today we use the stryker tool as part of our regular build flow. During this flow we would prefer to reduce or eliminate network touches during. On startup stryker reaches out to api.nuget.org (not configurable from what I see) to check for newer versions and print a message suggesting an update.

My preference would be to add support for a flag (--skip-version-check) to allow skipping this check flow completely as we already have access to things like Dependabot for suggesting version updates so there is less need for this kind of a check in automated flows.

mobratil:
## Description
This PR introduces a new `--skip-version-check` flag to allow users to bypass the automatic version check when running Stryker. This feature is useful in environments with restricted network access or when version updates are unnecessary.

## Changes
- Added `--skip-version-check` flag to the CLI.
- Updated unit tests
- Updated documentation to include details on the new flag.

Fixes #3044


docs/configuration.md
```diff
@@ -683,3 +683,12 @@ Command line: `--break-on-initial-test-failure`
 Config file: `break-on-initial-test-failure`
 
 Instruct Stryker to break execution when at least one test failed on initial test run.
+
+### `skip-version-check` &lt;`flag`&gt;
+
+Default: `false`  
+Command line: `--dev-mode`
```
jafletch, line 690: Should be --skip-version-check

src/Stryker.CLI/Stryker.CLI/CommandLineConfig/CommandLineConfigReader.cs
```diff
@@ -194,6 +194,7 @@ private void PrepareCliOptions(IStrykerInputs inputs)
             // Category: Misc
             AddCliInput(inputs.BreakOnInitialTestFailureInput, "break-on-initial-test-failure", null, optionType: CommandOptionType.NoValue, category: InputCategory.Misc);
             AddCliInput(inputs.DevModeInput, "dev-mode", null, optionType: CommandOptionType.NoValue, category: InputCategory.Misc);
+            AddCliInput(inputs.SkipVersionCheckInput, "skip-version-check", null, optionType: CommandOptionType.NoValue, category: InputCategory.Misc);
```
jafletch, line 197: There is actually a different concept in the code for a "CLI only" input. You can see this around line 22 in this same file. I understood that to be the method for adding inputs which were relevant for the CLI and not for stryker core. I would expect this new flag only to be a CLI input.

docs/configuration.md
mobratil: Fixed, thanks

src/Stryker.CLI/Stryker.CLI/CommandLineConfig/CommandLineConfigReader.cs
mobratil: @jafletch Please take a look now. Eventually, suggest how exactly this input should be implemented.
I still use AddCliInput method, however the input is no longer present in Core part.

I hade to use strong type `SkipVersionCheckInput` because when I tried to use AddCliOnlyInput it throws an exception during the startup because that flow expects options to be string. This one is actually NoValue command option type.

rouke-broersma:
Thanks, looks good! I made a couple modifications so the input is no longer neccesary.

Issue #3046, opened 2024-09-26
Change default color of console output


draptik:
The default color of the console output is difficult to read when using a console with light background.
A similar issue was raised in the past: https://github.com/stryker-mutator/stryker-net/issues/1602

**Describe the solution you'd like**
The above ticket mentions that Stryker uses `Serilog.Sinks.Console` for generating the output.
Serilog provides an easy way to change the color theme:
https://github.com/serilog/serilog-sinks-console?tab=readme-ov-file#themes

**Describe alternatives you've considered**
- Option 1: change the default theme to `AnsiConsoleTheme.Sixteen`. It's readable on dark and light backgroud (see screenshots below).
- Option 2: make the color theme configurable.

This seems to be the relevant part in the current code, if I'm not mistaking:
https://github.com/stryker-mutator/stryker-net/blob/567f028dc28fb180fd1e877a6e9992679da90935/src/Stryker.CLI/Stryker.CLI/Logging/ApplicationLogging.cs#L17

**Additional context**
![20240926-113216-swappy](https://github.com/user-attachments/assets/71b05edd-dbef-4bb2-9455-8cc4e5d43c72)
![20240926-113329-swappy](https://github.com/user-attachments/assets/1afde17e-898c-4337-a31e-674606340b32)
![20240926-113501-swappy](https://github.com/user-attachments/assets/16257839-f2d0-4fbb-9818-798a411c2925)
![20240926-113537-swappy](https://github.com/user-attachments/assets/0ad7aa20-5a5e-4516-a1d7-691306afc8f8)

```csharp
using Serilog;
using Serilog.Sinks.SystemConsole.Themes;

namespace ConsoleApp1;

public static class Program
{
    public static void Main(string[] _)
    {
        // For details see https://github.com/serilog/serilog-sinks-console?tab=readme-ov-file#themes
        //var defaultTheme = AnsiConsoleTheme.Literate;
        //var selectedTheme = defaultTheme;
        var selectedTheme = AnsiConsoleTheme.Sixteen;
        
        Log.Logger = new LoggerConfiguration()
            .WriteTo.Console(theme: selectedTheme)
            .CreateLogger();
        
        Log.Information("Hello World!");
    }
}
```

What do you think?

rouke-broersma:
Sounds good to me

bhargava-raju:
I want to work on this ticket, can you assign this to me

rouke-broersma:
[...]

PR #3047, opened 2024-09-26
feat(ui): make console output more readable

draptik:
Using Serilog's AniConsoleTheme.Sixteen improves readability of the console log output on dark and light background consoles. See https://github.com/stryker-mutator/stryker-net/issues/3046 for details.

### DIFF
--- a/src/Stryker.CLI/Stryker.CLI/Logging/ApplicationLogging.cs
+++ b/src/Stryker.CLI/Stryker.CLI/Logging/ApplicationLogging.cs
@@ -3,6 +3,7 @@ using LibGit2Sharp;
 using Microsoft.Extensions.Logging;
 using Serilog;
 using Serilog.Events;
+using Serilog.Sinks.SystemConsole.Themes;
 using LibGitLogLevel = LibGit2Sharp.LogLevel;
 using MSLogLevel = Microsoft.Extensions.Logging.LogLevel;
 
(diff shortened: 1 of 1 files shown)


rouke-broersma:
@draptik I'm testing out the change on my own system and the default theme already seems to be working fine for me on light themes. Coupled with the note by the serilog team that ansi themes don't work well on cmd that makes me hesitant to approve the change.

Could there by anything in your specific environment configuration that's affecting this? If I run the latest released version of stryker in the VS console with light theme the console output colour is black, not white as in yours:

image


In fact if you look at your own example you can see that the default colour for the terminal seems to actually be white, because all other printed text is also white:

image

So it seems to me like serilog is actually following the configured terminal colors.

draptik:
OK, understood. Thanks for the update!
Since this issue seems to be more involved than I initially thought, feel free to discard this PR.

rouke-broersma:
> OK, understood. Thanks for the update!
> Since this issue seems to be more involved than I initially thought, feel free to discard this PR.

Could you give some more information on your terminal theme/configuration so we could determine whether the results you're seeing are intentional or not?

draptik:
Sure.
- linux: [kitty terminal](https://sw.kovidgoyal.net/kitty/) "Tango Light" / "Tango Dark"
- windows: Windows Terminal "Tango Light" / Tango Dark"
- Rider (linux/windows): built-in terminal with "Rider Dark" / "Rider Day" themes.

PR #3060, opened 2024-10-07
feat(Regex): Add more regex mutations

Pentiva:
Added more regex mutations from [weapon-regex](https://github.com/stryker-mutator/weapon-regex).

weapon-regex has a concept of levels for which mutations are enabled which might be good to add in some way to not have too many unneeded mutations.

I also refactored RegexMutantOrchestrator, because all the mutators were stored in a dictionary based on type, but the dictionary was looped through anyway, so the type check was moved to IRegexMutator and RegexMutatorBase provides a better type check as it uses the `is` expression with generics instead of reflection. (This might also make it easier to filter out specific regex expression in a config file, as each mutator could provide its own required level.). I think this aligns it much more with the core mutators.


rouke-broersma:
I will probably review this Friday, I haven't forgotten you 👌

rouke-broersma:
Thanks!

PR #3070, opened 2024-10-19
fix(reference alias): Properly support PackageReference Alias

### LINKED ISSUE #3069: Reference Assembly Aliases are not passed to compiler
**Describe the bug**
We recently added a package reference alias for a package in Stryker.Core. Mutation testing now fails on Stryker.Core because we don't correctly parse the reference alias.

**Logs**
image

**Expected behavior**
Stryker compiles with Reference Alias since we have support for this.

**Additional context**
At least currently the References returned from Buildalyzer do not contain references in the format that we try to parse to detect Aliases:

https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultExtensions.cs#L110

Instead PackageReferences does seem to include Aliases:

image

Adel-Atzouza:
Partial fix for #3069

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultExtensions.cs
+++ b/src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultExtensions.cs
@@ -107,15 +107,26 @@ public static class IAnalyzerResultExtensions
     {
         foreach (var reference in analyzerResult.References)
         {
-            if (reference.Contains('='))
+            var referenceFileNameWithoutExtension = Path.GetFileNameWithoutExtension(reference);
+            string packageWithAlias = null;
+
+            if (analyzerResult.PackageReferences is not null) // Check if the analyzer result has package references because the Unit Tests don't have them.
             {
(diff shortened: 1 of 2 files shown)


src/Stryker.Core/Stryker.Core/Initialisation/Buildalyzer/IAnalyzerResultExtensions.cs
```diff
@@ -107,15 +107,26 @@ public static IEnumerable<MetadataReference> LoadReferences(this IAnalyzerResult
     {
         foreach (var reference in analyzerResult.References)
         {
-            if (reference.Contains('='))
+            var referenceFileNameWithoutExtension = Path.GetFileNameWithoutExtension(reference);
+            string packageWithAlias = null;
+
+            if (analyzerResult.PackageReferences is not null) // Check if the analyzer result has package references because the Unit Tests don't have them.
```
rouke-broersma, line 113:
```suggestion
            if (analyzerResult.PackageReferences is not null)
```

PR #3074, opened 2024-10-21
chore: file-scoped namespaces

### LINKED ISSUE #2470: Refactor all files to use file-scoped namespaces

Myoxocephalus:
Changed all files to file-scoped namespaces. This will close #2470.
Based on the open PRs this seems to be an ok time to do this. If you prefer to merge something else first i can keep this branch up to date.


rouke-broersma:
I agree, now seems a good a time as any to push this through, however:

You seem to have reused the master branch which means this PR includes unrelated commits (that may already have been merged with our master, but it would be complicated to track this since we squash merge) and could potentially change behavior. Due to the nature of this pull request that would be impossible to properly review. Could you please create a new branch from our master branch instead of yours, and perform only the file-scoped namespace change? This would greatly assist me in reviewing this.

PR #3082, opened 2024-10-24
feat(mutators): Add Collection Expression Mutator

Pentiva:


src/Stryker.Core/Stryker.Core.UnitTest/Mutators/CollectionExpressionMutatorTests.cs
```diff
@@ -0,0 +1,523 @@
...
+                                           .Object
+            }
+        };
+
+        var target = new CsharpCompilingProcess(input);
+
+        try
+        {
+            var result =
+                target.Compile([
```
richardwerkman, line 500: Could you explain why you are compiling the mutations here?

src/Stryker.Core/Stryker.Core.UnitTest/Mutants/CsharpMutantOrchestratorTests.cs
```diff
@@ -1948,4 +1951,315 @@ public void ShouldIncrementMutantCountUniquely()
...
+    }
+
+    [TestMethod]
+    public void ShouldMutateImplicitCollectionExpressionParameter()
+    {
+        var source = """
+                     public void M() {
+                         Iter([1]);
+                     }
+                     public IEnumerable<T> Iter<T>(IList<T> list) {
```
richardwerkman, line 2208: I think we should strive to keep these tests as simple and clear as possible. In this case I think we don't need this method here. To keep this test simple we could remove this and only test the implicit collection part

src/Stryker.Core/Stryker.Core.UnitTest/Mutants/CsharpMutantOrchestratorTests.cs
```diff
@@ -1948,4 +1951,315 @@ public void ShouldIncrementMutantCountUniquely()
...
+    }
+
+    [TestMethod]
+    public void ShouldMutateUsedCollectionExpression()
+    {
+        var source = """
+                     public void M() {
+                         // Stryker disable String : Not mutation under test
+                         Span<string> weekDays = ["Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"];
+                         foreach (var day in weekDays)
```
richardwerkman, line 2029: I think we should strive to keep these tests as simple and clear as possible. In this case I think we can remove the foreach to only test the collection expression

src/Stryker.Core/Stryker.Core.UnitTest/Mutants/CsharpMutantOrchestratorTests.cs
```diff
@@ -1948,4 +1951,315 @@ public void ShouldIncrementMutantCountUniquely()
...
+                         string lithium = "Li";
+                         string beryllium = "Be";
+                         string boron = "B";
+                         string carbon = "C";
+                         string nitrogen = "N";
+                         string oxygen = "O";
+                         string fluorine = "F";
+                         string neon = "Ne";
+                         string[] elements = [hydrogen, helium, lithium, beryllium, boron, carbon, nitrogen, oxygen, fluorine, neon];
+                         foreach (var element in elements)
```
richardwerkman, line 2124: I think we should strive to keep these tests as simple and clear as possible. In this case we can remove the foreach to only test the collection expression

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/CollectionExpressionMutatorTests.cs
```diff
@@ -0,0 +1,523 @@
...
+    public void ShouldAddValueToEmptyCollectionExpression(string expression)
+    {
+        var expressionSyntax = SyntaxFactory.ParseExpression(expression) as CollectionExpressionSyntax;
+        var target = new CollectionExpressionMutator();
+        var result = target.ApplyMutations(expressionSyntax, null);
+        
+        var mutation = result.ShouldHaveSingleItem();
+        mutation.DisplayName.ShouldBe("Collection expression mutation");
+        var replacement = mutation.ReplacementNode.ShouldBeOfType<CollectionExpressionSyntax>();
+        replacement.Elements.ShouldNotBeEmpty();
```
richardwerkman, line 50: Should we check here what the element is?

richardwerkman (commented):
Thanks for your contribution! I have a few comments and a question, but this already seems good to me :)

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/CollectionExpressionMutatorTests.cs
Pentiva: To ensure that adding a mutation doesn't introduce ambiguity on compile. I've tried to test as many cases as I can that have the compiler infer a collection type from the context, and I don't wont any of them (mainly the mutation to a empty collection) to have it unable to infer a type. I originally added these to the integration project, but I moved them because I was struggling to consistently run the integration tests. I can move it back, or remove it out-right, as these mutations should be implemented as `(mutated ? [original] : [])` and therefore always have the same type information as before.

src/Stryker.Core/Stryker.Core.UnitTest/Mutants/CsharpMutantOrchestratorTests.cs
Pentiva: I mainly copied all these examples directly from the Microsoft docs on collection expressions, (https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/collection-expressions) just to have a large test area, but I agree that most of the examples are to show usage, not creation and can be trimmed down. Same with the other tests you have commented on.

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/CollectionExpressionMutatorTests.cs
Pentiva: I'll update it to do so. I think I originally didn't because I also had another mutation that added two default elements. But thinking now, they would all be check-able anyway, so it was probably just carelessness.

src/Stryker.Core/Stryker.Core.UnitTest/Mutators/CollectionExpressionMutatorTests.cs
richardwerkman: Sounds about right. It would make more sense to me to test these things in the integration tests. But I also don't mind keeping it here, since it makes sense to test the return types for these kind of mutations.

richardwerkman (approved):
Thanks for your contribution! I'll merge this and it will be part of our next release :)

PR #3086, opened 2024-10-28
added since on line level changes

shivambagadia23:


rouke-broersma (changes requested):
Please revert the files to file scope namespace

src/Stryker.Core/Stryker.Core/DiffProviders/GitDiffProvider.cs
```diff
@@ -1,81 +1,116 @@
...
-        var diffResult = new DiffResult()
-        {
-            ChangedSourceFiles = new Collection<string>(),
-            ChangedTestFiles = new Collection<string>()
-        };
-
-        // A git repository has been detected, calculate the diff to filter
-        var repository = _gitInfoProvider.Repository;
-        var commit = _gitInfoProvider.DetermineCommit();
+        public TestSet Tests { get; }
```
inishantmishra, line 18: Why did we move these inside class?

shivambagadia23: @inishantmishra We did not change anything there it was like that only

src/Stryker.Core/Stryker.Core/MutantFilters/SinceMutantFilter.cs
```diff
@@ -74,12 +93,30 @@ public IEnumerable<IMutant> FilterMutants(IEnumerable<IMutant> mutants, IReadOnl
...
+
+                    _logger.LogDebug("Original line span is found to be {actualLineSpan}", actualLineSpan);
+                    var start = actualLineSpan.Span.Start.Line;
+                    var end = actualLineSpan.Span.Start.Line;
+                    if (start <= line && line <= end)
+                    {
+                        if (mutant.ResultStatus != MutantStatus.NoCoverage)
+                        {
+                            mutant.ResultStatus = MutantStatus.Pending;
+                            mutant.ResultStatusReason = "Mutant changed compared to target commit";
```
rouke-broersma, line 112: The status reason is not correct, the mutant has not been changed but a covering test has been changed

src/Stryker.Core/Stryker.Core/MutantFilters/SinceMutantFilter.cs
```diff
@@ -55,17 +54,37 @@ public IEnumerable<IMutant> FilterMutants(IEnumerable<IMutant> mutants, IReadOnl
...
+                        }
```
rouke-broersma, line 82: Why do you not use SetMutantStatusForFileChanged?

src/Stryker.Core/Stryker.Core/MutantFilters/SinceMutantFilter.cs
rouke-broersma, line 112: Why do you not use ResetMutantStatusForChangedTests?

Issue #3102, opened 2024-11-08
Add top-level-statements to integration test project


richardwerkman:
The integration test should contain [top-level-statements](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/program-structure/top-level-statements
) to ensure that these are mutated as expected.

Make sure to include some stryker comments as well since these caused issues before #3073.

PR #3116, opened 2024-11-21
feat: Protect compiler directives during mutation

### LINKED ISSUE #3081: Stryker messes with compiler directives
**Describe the bug**
Mutated code may end up containing ill-formed compiler directives. I.e, Stryker does not ensure directives appear at the beginning of source lines. They may end up being captured within flow control statements.

**Logs**
When mutating this method
```csharp
public void SomeMethod() {
    var x = 0;
#if !DEBUG
    x++;
#endif
}
```
Stryker generates this version
```csharp
public void SomeMethod() {if(StrykerNamespace.MutantControl.IsActive(0)){}else{
    var x = 0;
if(StrykerNamespace.MutantControl.IsActive(1)){;}else{if(StrykerNamespace.MutantControl.IsActive(2)){#if !DEBUG
    x--;
}else{#if !DEBUG
    x++;
}}#endif
}}
```

**Expected behavior**
Compiler directives should not be transformed by the mutation step. 
Previous example should result in something like this:
```csharp
public void SomeMethod() {if(StrykerNamespace.MutantControl.IsActive(0)){}else{
    var x = 0;
#if !DEBUG
if(StrykerNamespace.MutantControl.IsActive(1)){;}else{if(StrykerNamespace.MutantControl.IsActive(2)){    x--;
}else{
    x++;
}}
#endif    
}}
```

 - OS: NA
 - Type of project NA
 - Framework Version NA
 - Stryker Version 4.3 , and earlier versions

dupdob:
Ensure compiler directives (#if, #pragma....) are not mutated but kept around mutated syntax nodes.

This required small changes in every mutator, as they are not responsible for removing any directive.
I took this as an opportunity to improve the design.

Also, added generation of semantic model during mutation tests to ensure mutator related unit tests are relevant.
I may have fixed some minor issues on them.

Also added some top level statement in integration test as it was a quick win

FIxes #3081,  #3102


dupdob:
I am happy to see this PR reduces the total number of lines, and it increases the number of mutations. This is probably related to the redesign of the pattern mutator that I modified so it no longer does some internal scan and reenabled orchestration of patterns in general

integrationtest/TargetProjects/NetCore/EmptyTestProject/Main.cs
```diff
@@ -0,0 +1,2 @@
+
+Console.WriteLine("this is a global statement");
```
rouke-broersma, line 2: I think this should be in https://github.com/stryker-mutator/stryker-net/tree/master/integrationtest/TargetProjects/NetCore/TargetProject/Constructs as a new file called `Csharp9_TopLevelStatement.cs`. We are trying to keep an accurate representation of csharp features per language level so that we can easily figure out if we're missing certain constructs in the integration tests.

rouke-broersma: Mhm now that I say that, is that even possibly or does your entry point have to be top-level as well for this to work?

richardwerkman: I think that should work. `An application must have only one entry point. A project can have only one file with top-level statements.` It doesn't state that it has to be a file in the root of the project or with a specific name.

dupdob: well
Now that I have tried to make the change, I know why I did not choose this route. Top level statements are allowed only for executable targets.
Shall I convert the project to output an executable instead of a library assembly?

rouke-broersma: I think that's fine yes

dupdob: done

src/Stryker.Core/Stryker.Core/Mutators/NullCoalescingExpressionMutator.cs
```diff
@@ -74,6 +69,8 @@ private static bool IsNullable(SyntaxNode node, SemanticModel semanticModel)
         }
 
         var typeInfo = semanticModel.GetTypeInfo(node);
-        return typeInfo.Nullability.FlowState == NullableFlowState.MaybeNull;
+        // assume nullability if type resolution failed for some reason
+        return (typeInfo.ConvertedType is { } && typeInfo.ConvertedType.TypeKind is TypeKind.Error or TypeKind.Unknown)
```
Liam-Rougoor, line 73:
```suggestion
        return (typeInfo.ConvertedType is not null && typeInfo.ConvertedType.TypeKind is TypeKind.Error or TypeKind.Unknown)
```

Liam-Rougoor (commented):
Thanks for your PR, @dupdob ! It's looking pretty good.

One more question, just to verify my understanding: now, all mutators should remove almost all trivia, including directives. Then, the InstrumentationEngines re-add the removed trivia, correct?

Is this process designed like this to make sure that the trivia is only added once (to prevent duplicate directive trivia, for example)?

src/Stryker.Core/Stryker.Core/Instrumentation/ConditionalInstrumentationEngine.cs
```diff
@@ -22,15 +23,16 @@ public ParenthesizedExpressionSyntax PlaceWithConditionalExpression(ExpressionSy
                 SyntaxFactory.ConditionalExpression(
                     condition: condition,
                     whenTrue: mutated,
-                    whenFalse: original))
-            // Mark this node as a MutationConditional node. Store the MutantId in the annotation to retrace the mutant later
-            .WithAdditionalAnnotations(Marker);
+                    whenFalse: original)).
+        WithTriviaFrom(original).
```
Liam-Rougoor, line 27: Does this always work? In some mutators, we mutate sub-nodes, so we might remove trivia there, but not re-add them here. Is that an issue?

src/Stryker.Core/Stryker.Core/Instrumentation/IfInstrumentationEngine.cs
```diff
@@ -35,7 +37,7 @@ protected override SyntaxNode Revert(IfStatementSyntax ifNode)
     {
         if (ifNode.Else?.Statement is BlockSyntax block)
         {
-            return block.Statements.Count == 1 ? block.Statements[0] : block;
+            return (block.Statements.Count == 1 ? block.Statements[0] : block).WithTriviaFrom(ifNode);
```
Liam-Rougoor, line 40: Does this always work? In some mutators, we mutate sub-nodes, so we might remove trivia there, but not re-add them here. Is that an issue?

dupdob:
@Liam-Rougoor: short answer is yes. 
Long answer is: I am not certain, it depends on what Roslyn handles as trivia and how it anchors is. 
Discovering that directives were modeled as trivia was disconcerting, to say the least.

src/Stryker.Core/Stryker.Core/Mutants/CsharpMutantOrchestrator.cs
```diff
@@ -143,15 +145,16 @@ internal IEnumerable<Mutant> GenerateMutationsForNode(SyntaxNode current, Semant
         {
             foreach (var mutation in mutator.Mutate(current, semanticModel, Options))
             {
+                mutation.OriginalNode = current;
```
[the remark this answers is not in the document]
dupdob: This is now a strong requirement: there should be not 'out of band' mutation. I.e. mutators must generate mutations for the provided node and no other.

src/Stryker.Core/Stryker.Core/Instrumentation/ConditionalInstrumentationEngine.cs
dupdob: see previous comment: there shall be no sub node mutation. Otherwise it will not work; I failed to find a way to support both sub node mutations AND general trivia support.

src/Stryker.Core/Stryker.Core/Instrumentation/IfInstrumentationEngine.cs
dupdob: see previous comment: there shall be no sub node mutation. Otherwise it will not work; I failed to find a way to support both sub node mutations AND general trivia support.

rouke-broersma:
Sounds like a good improvement, we will need to be vigilant to keep this in place.

dupdob:
sub node mutation fails currently, as in the mutation will fail to be injected and will raise an exception, alerting the developer he/she is violating something. 
I feel this is a better design, but, as there is no trivia cleaning for now (or any other change to the original node), this is actually required. So the decision could be reversed; either way, I propose will keep it as is until we have a clear benefit to choose one or the other.

I tried for days to clean trivia beforehand, but I failed to achieve a correct orchestration while doing it, so I reverted my strategy (as presented in the slack, IIRC), but kept what I feel like a cleaner design, due to the separation of concerns between orchestration and mutation

rouke-broersma:
I don't see any reason to try to reverse this decision unless we notice that it causes other problems, as I agree that the clearer separation sounds like a good direction 👌.

PR #3131, opened 2024-12-02
Add class level mutation control (for method only)

dupdob:
This is an extension of the mutation logic design to support mutations for whole methods.


dupdob:
closing for being too old. Branch is kept for the time being

PR #3132, opened 2024-12-02
feat: Restore support for aliases via Buildalyzer

dupdob:
This PR restores support for reference aliases thanks to Buildalyzer V7.10.
In details:
- Use Buildalyzer provided Aliases (support multiple aliases for a single assembly)
- Drop supports for package aliases (no need)
- Existing unit tests have been updated
- Integration tests projects also have aliases, to capture any regression from Buildlayzer


rouke-broersma:
@dupdob this seems complete to me, is it blocked by something?

dupdob:
Yep
I forgot to switch to ready for review.

PR #3140, opened 2024-12-12
chore: rename package variable in release script because it's a reserved keyword

michha:
package is a reserved keyword, so SonarQube lists an issue about it: https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&severities=BLOCKER&id=stryker-net


rouke-broersma:
Thanks!

PR #3163, opened 2025-01-17
chore: Add testrunner abstractions

richardwerkman:
To prepare for adding the new MsTestRunner we need to add a wrapper for the current testrunner. That makes it possible to add a second one and support switching between the runners.


src/Stryker.TestRunner/Tests/TestGuidsList.cs
```diff
@@ -0,0 +1,89 @@
+using Stryker.Abstractions.Testing;
+using System.Collections.Generic;
+using System;
+using System.Linq;
+
+namespace Stryker.TestRunner.Tests;
+
+public class TestGuidsList : ITestIdentifiers
```
rouke-broersma, line 8: Isn't this class duplicate?

richardwerkman: Not exactly. But I'll see if I can make the class generic to reuse it

Issue #3181, opened 2025-02-03
ArgumentNullException during the mutation test run


mrlacey:
**Describe the bug**

Trying Stryker for the first time, but getting an internal exception when generating mutations.

**Logs**
For us to best assist you in diagnosing the problem we would ask you to run stryker with the option `--log-to-file` or `-L` and to attach the resulting logs to the issue.

```ascii

   _____ _              _               _   _ ______ _______  
  / ____| |            | |             | \ | |  ____|__   __| 
 | (___ | |_ _ __ _   _| | _____ _ __  |  \| | |__     | |    
  \___ \| __| '__| | | | |/ / _ \ '__| | . ` |  __|    | |    
  ____) | |_| |  | |_| |   <  __/ |    | |\  | |____   | |    
 |_____/ \__|_|   \__, |_|\_\___|_| (_)|_| \_|______|  |_|    
                   __/ |                                      
                  |___/                                       


[20:51:27 INF] Logging enabled at level Trace
Version: 4.5.0
[20:51:27 DBG] Stryker starting, version: 4.5.0

[20:51:27 VRB]   CACHE https://api.nuget.org/v3/registration5-gz-semver2/dotnet-stryker/index.json
[20:51:27 VRB]   CACHE https://api.nuget.org/v3/registration5-gz-semver2/dotnet-stryker/index.json
[...]

PR #3184, opened 2025-02-05
fix: Improve project analysis when normal analysis fails

dupdob:
Fix issues identified during analyis of issue #3181 :

- skip buildanalyzer result with no defined target framework (`null`). It triggered an obscure exception before
- fallbaback to project reference scan when assembly reference scan fails to identify a mutable project.

Both fixes improve tolerance in support for alternative SDKs (eg MAUI)


slang25:
@dupdob only just spotted this, I also opened #3185 which may be relevant

src/Stryker.Utilities/Buildalyzer/IAnalyzerResultExtensions.cs
```diff
@@ -147,6 +147,8 @@ public Assembly LoadFromPath(string fullPath)
 
     public static NuGetFramework GetNuGetFramework(this IAnalyzerResult analyzerResult)
     {
+        if (string.IsNullOrEmpty(analyzerResult.TargetFramework))
+            return null;
```
rouke-broersma, line 151:
```suggestion
        if (string.IsNullOrEmpty(analyzerResult.TargetFramework))
        {
            return null;
        }
```

Issue #3038, opened 2024-09-18
Upgrading from 4.0.6 to 4.2.0 analysis no longer works for project


martincostello:
**Describe the bug**

Attempting to upgrade from 4.0.6 to 4.2.0, analysis is failing for one of our test projects, causing an error.

Specifying `--dev-mode` doesn't appear to add any additional details.

### 4.0.6

```console
Running mutation tests for '../src/Polly/Polly.csproj'. Test Project: 'Polly.Specs/Polly.Specs.csproj'

   _____ _              _               _   _ ______ _______  
  / ____| |            | |             | \ | |  ____|__   __| 
 | (___ | |_ _ __ _   _| | _____ _ __  |  \| | |__     | |    
  \___ \| __| '__| | | | |/ / _ \ '__| | . ` |  __|    | |    
  ____) | |_| |  | |_| |   <  __/ |    | |\  | |____   | |    
 |_____/ \__|_|   \__, |_|\_\___|_| (_)|_| \_|______|  |_|    
                   __/ |                                      
                  |___/                                       


Version: 4.0.6

A new version of Stryker.NET (4.2.0) is available. Please consider upgrading 
using `dotnet tool update -g dotnet-stryker`

[09:02:00 INF] Analysis starting.
[09:02:06 WRN] Could not find a project analysis for the chosen target framework net8.0. 
The available target frameworks are: net462,net472,net6.0,netstandard2.0. 
The first available framework will be selected, which is net462.
[09:02:06 INF] Found project D:\a\Polly\Polly\src\Polly\Polly.csproj to mutate.
[09:02:06 INF] Analysis complete.
[09:02:06 INF] Building test project D:\a\Polly\Polly\test\Polly.Specs\Polly.Specs.csproj (1/1)
[09:03:18 INF] Number of tests found: 1986 for project D:\a\Polly\Polly\src\Polly\Polly.csproj. Initial test run started.
```

### 4.2.0

```console
Running mutation tests for '../src/Polly/Polly.csproj'. Test Project: 'Polly.Specs/Polly.Specs.csproj'

   _____ _              _               _   _ ______ _______  
  / ____| |            | |             | \ | |  ____|__   __| 
 | (___ | |_ _ __ _   _| | _____ _ __  |  \| | |__     | |    
  \___ \| __| '__| | | | |/ / _ \ '__| | . ` |  __|    | |    
  ____) | |_| |  | |_| |   <  __/ |    | |\  | |____   | |    
[...]

PR #3185, opened 2025-02-06
Pass TFM into Buildalyzer build

slang25:
See here: https://github.com/phmonte/Buildalyzer/issues/297#issuecomment-2638340201
and #3181 for context.

Passing the TFM into Buildalyzer prevents all TFMs being built for a multitarget project.

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/BuildAnalyzerTestsBase.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/BuildAnalyzerTestsBase.cs
@@ -254,6 +254,7 @@ public class BuildAnalyzerTestsBase : TestBase
         projectAnalyzerMock.Setup(x => x.Build(It.IsAny<string[]>(), It.IsAny<EnvironmentOptions>())).Returns(sourceProjectAnalyzerResultsMock);
         projectAnalyzerMock.Setup(x => x.Build(It.IsAny<EnvironmentOptions>())).Returns(sourceProjectAnalyzerResultsMock);
         projectAnalyzerMock.Setup(x => x.Build(It.IsAny<string>())).Returns(sourceProjectAnalyzerResultsMock);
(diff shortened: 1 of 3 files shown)


dupdob:
Not sure of the link with issue #3181.
Specifying the TFM to Buildalyzer/MsBuild does result in **forcing** the TFM for every project, even if no configuration exist for said TFM.

We tried this in the past and it resulted in unexpected results when working on non trivial multitargeting project, such as library project targeting some `netstandard` version and test projects targeting some actual framework version.

At best, we could make this behavior an option, so user could reduce the initialization time when this is supported by the project, but It cannot be the default behavior.

martincostello:
> We tried this in the past and it resulted in unexpected results when working on non trivial multitargeting project, such as library project targeting some `netstandard` version and test projects targeting some actual framework version.

Yeah, I think this was #3038 as it caused issues with Polly.

dupdob:
yes. @martincostello 's comment gave me the opportunity to look again at my resolution table and fix an error there.

dupdob:
that was not my point. I am afraid this will not solve the issue mentioned in #3038: we do not know how framework selection work with project dependency resolution. Trying to force a framework for every single project will lead to undefined behvaviour.

What problem are you trying to solve here? you should open an issue so we can discuss how to achieve the desired outcome

slang25:
Sure I'll create an issue to explain

dupdob:
I saw the Buildalyzer [issue 297](https://github.com/phmonte/Buildalyzer/issues/297). I think my PR (#3184) workaround the Buildalyzer issue with framework selection: on a sample MAUI project, `Build()` returns two results, one with a null TFM and a legal one (with net9.0).
So I simply ignore the null TFM result and everything moves on as expected. So nothing more to fix here.

The ideal logic regarding TFM could be that Stryker forces the TFM only for test projects analysis.
But, Stryker must perform a project analysis to identify which projects are test projects. Kind of a catch22 situation here....

PR #3200, opened 2025-02-17
feat(mutating): Mutate unsigned right-shift operator

martincostello:
Add support for mutating to/from the [unsigned right-shift operator (`>>>`)](https://learn.microsoft.com/dotnet/csharp/language-reference/operators/bitwise-and-shift-operators#unsigned-right-shift-operator-).

| **Original** | **Mutations** |
|:-------------|:--------------|
| `>>`         | `<<`, `>>>`   |
| `<<`         | `>>`, `>>>`   |
| `>>>`        | `<<`, `>>`    |
| `>>=`        | `<<=`, `>>>=` |
| `<<=`        | `>>=`, `>>>=` |
| `>>>=`       | `<<=`, `>>=`  |

### DIFF
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -3,12 +3,6 @@ run-name: e2e ${{ github.head_ref }} on m1 macos
 
 on: [pull_request]
 
-env:
-  NUGET_FEED: https://f.feedz.io/stryker/stryker-net/nuget/index.json
-  FEEDZ_NUGET_API_KEY: T-9O45ZZZ5lwQ6ptrGpDuoX8tdZLliNkScn9k
-  RestoreLockedMode: true
-  VERSION: "0.0.0-github-${{ github.run_number }}"
-
 jobs:
   integration-test:
     runs-on: macos-14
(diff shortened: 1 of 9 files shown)


rouke-broersma:
Thanks! Could you also add an example of this operation to our integration tests?

https://github.com/stryker-mutator/stryker-net/blob/master/integrationtest/TargetProjects/NetCore/TargetProject/Constructs/CSharp11.cs

martincostello:
Thanks - I didn't find that as I assumed everything relevant was in the main Stryker `.sln` file.

Is there some documentation on how to setup these tests so I can get them passing locally?

If I try to run them as-is, I get errors about missing files:

```
InitCommandDefaults => Json file missing
CSharp_NetCore_SingleTestProject => System.IO.DirectoryNotFoundException : Could not find a part of the path 'C:\Coding\stryker-mutator\stryker-net\integrationtest\TargetProjects\NetCore\NetCoreTestProject.XUnit\StrykerOutput'.
```

I assume it involves replicating this?

https://github.com/stryker-mutator/stryker-net/blob/b873b6268c7bdd9ca7529ac50e6329d05327f389/.github/workflows/integration-test.yaml#L23-L33

martincostello:
It's OK, I've reverse engineered it.

I'll push up a commit with a script to make it easier for contributors in the future once I've updated the tests.

martincostello:
Integration tests added.

martincostello:
Looks like I need to update the results for the tests that run in AzDO too.

martincostello:
Fixed.

PR #3205, opened 2025-02-23
feat(mutating): Support static using references for Math mutations

### LINKED ISSUE #2933: Math mutations are not placed when static using is used
**Describe the bug**

``` cs
Math.Floor(5.5); // mutated

using static System.Math;
Floor(5.5); // not mutated
```

**Expected behavior**
I'd expect the second Floor method to be mutated as well.

**Additional context**
We currently check the syntax tree for the class name. We should use the semantic model for this.

jacinyan:
### Summary
Fixes #2933 by using the semantic model to detect calls to System.Math when using `static using`. This ensures `Floor(5.5)` is correctly mutated, just like `Math.Floor(5.5)`.

### Changes
- Added semantic model checks in both member access and direct call branches.
- Updated MathMutator to handle IdentifierNameSyntax calls to Math methods.

### Testing
- Verified with a sample C# project using static imports (`FloorValue`, `MixedFloorValue`, etc.).
- Mutation score increased and static calls now generate mutants.

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutators/MathMutatorTest.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutators/MathMutatorTest.cs
@@ -159,7 +159,7 @@ namespace TestApplication
     }
 
     /// <summary>
-    /// Test Method to check, if different expressions aren't mutated (in case of non-Math classes)
+    ///     Test Method to check, if different expressions aren't mutated (in case of non-Math classes)
     /// </summary>
     [TestMethod]
     [DataRow("MyClass")]
@@ -240,6 +240,36 @@ namespace TestApplication
         }
     }
 
+    /// <summary>
+    ///     Explicit test: verifies that using static import, Floor(5.0) is mutated to Ceiling(5.0).
+    /// </summary>
+    [TestMethod]
+    public void ShouldMutateStaticFloorToCeiling()
+    {
+        var target = new MathMutator();
+        var expression = GenerateStaticCallExpression("Floor");
+        var result = target.ApplyMutations(expression, null).ToList();
+
+        result.Count.ShouldBe(1);
+        var mutatedMethodName = ((IdentifierNameSyntax)((InvocationExpressionSyntax)result[0].ReplacementNode).Expression).Identifier.ValueText;
+        Enum.Parse<MathExpression>(mutatedMethodName).ShouldBe(MathExpression.Ceiling);
+    }
+
+    /// <summary>
+    ///     Explicit test: verifies that using static import, Exp(5.0) is mutated to Log(5.0).
+    /// </summary>
+    [TestMethod]
+    public void ShouldMutateStaticExpToLog()
+    {
+        var target = new MathMutator();
+        var expression = GenerateStaticCallExpression("Exp");
+        var result = target.ApplyMutations(expression, null).ToList();
+
+        result.Count.ShouldBe(1);
+        var mutatedMethodName = ((IdentifierNameSyntax)((InvocationExpressionSyntax)result[0].ReplacementNode).Expression).Identifier.ValueText;
+        Enum.Parse<MathExpression>(mutatedMethodName).ShouldBe(MathExpression.Log);
+    }
+
     private static IEnumerable<object[]> TrigonometricHyperbolicTestData
     {
         get
@@ -276,4 +306,6 @@ namespace TestApplication
             }
         }
     }
+
+
 }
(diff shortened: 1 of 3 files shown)


dupdob:
thanks for the PR. You need to add some tests to explicitly test for static using.

May I also suggest you should check first if the method name is of interest before resolving its symbol ?
As it is written, this mutator will ask to resolve the full symbol for every method invocation which may add to the total execution time.

Alternatively, having a benchmark (on a significant/large project) that demonstrates the cost being negligible is ok too .

jacinyan:
> thanks for the PR. You need to add some tests to explicitly test for static using.
> 
> May I also suggest you should check first if the method name is of interest before resolving its symbol ? As it is written, this mutator will ask to resolve the full symbol for every method invocation which may add to the total execution time.
> 
> Alternatively, having a benchmark (on a significant/large project) that demonstrates the cost being negligible is ok too .

Hi @dupdob ,

Thanks for your feedback. I've added explicit tests for static using calls and updated MathMutator to check the method name before resolving the symbol. Unfortunately, I haven't been able to run benchmarks on a large project yet, but based on theoretical analysis, the filtering should mitigate performance concerns.

Let me know if you have any further suggestions.

dupdob:
thanks @jacinyan. This is exactly what I had in mind; filtering on the name and then checking for actual type looks like the optimized approach.
I added some remarks, in line, but I will reproduce them here as it looks you did not see them:
> Thanks for the optimization. There is still one small issue: this mutator contains two versions of the list of mutateable mathematical functions: one as strings (`KnownMathMethodNames`) and one as an enumeration (`MathExpression)`. Which means that if someone changes one of the list, she will have to do it on both list, otherwise strange issues may appear.
I think you can remove the strings version using `Enum.TryParse(methodNameText, out _)` instead of `KnownMathMethodNames.Contains(methodNameText))`.
This way, we just need one list which means easier and safer maintenance.
> That being done, please fix failing unit tests and this PR should be good to merge.

jacinyan:
> thanks @jacinyan. This is exactly what I had in mind; filtering on the name and then checking for actual type looks like the optimized approach. I added some remarks, in line, but I will reproduce them here as it looks you did not see them:
> 
> > Thanks for the optimization. There is still one small issue: this mutator contains two versions of the list of mutateable mathematical functions: one as strings (`KnownMathMethodNames`) and one as an enumeration (`MathExpression)`. Which means that if someone changes one of the list, she will have to do it on both list, otherwise strange issues may appear.
> > I think you can remove the strings version using `Enum.TryParse(methodNameText, out _)` instead of `KnownMathMethodNames.Contains(methodNameText))`.
> > This way, we just need one list which means easier and safer maintenance.
> > That being done, please fix failing unit tests and this PR should be good to merge.

Thanks @dupdob . I've made some changes accordingly.  All tests passed. Please review.

jacinyan:
Hi @rouke-broersma,I was wondering if you would also be available to review this PR. Greatly appreciated.

rouke-broersma:
Hi @jacinyan 

I will take a look if I have time today however @dupdob is also a project member and they are already reviewing your PR.

It looks like the integration tests are failing, is that expected?

dupdob:
Sorry, I have been very busy past weeks and I assumed your were fixing the code so that tests pass.
as you can see here, the CI clearly have failing tests:

`[Tests](https://dev.azure.com/stryker-mutator/506a1f46-900e-434e-805f-ff8d36fc81af/_build/results?buildId=20899&view=ms.vss-test-web.build-test-results-tab)

    Failed: 138 (2.82%)
    Passed: 4,737 (96.75%)
    Other: 21 (0.43%)
    Total: 4,896`
 
 Can you look into it, do you need help with this?

jacinyan:
> can
Oh, sincere apologies.@rouke-broersma and especially @dupdob for overlooking things again and I've been busy with my term lately too. I'll get around to fixing the last bits in the next few weeks. 🙏

dupdob:
@rouke-broersma : yes, and thanks for the reminder. I planned to look into this PR, but forgot due to may paying job 😄 .
I just highlighted but copilot comment for I am not sure what to with it.

that being said, this PR is definitely merge ready for me.

rouke-broersma:
@dupdob I am testing out this copilot feature in the hopes of finding a useful use of all these new LLM tools somewhere, you don't need to do anything with it. Just noticed it was available today. I believe it mostly creates a summary of the changes, and tries to suggest improvements. If the summary and improvements prove to be helpful I might contain to use this, if not I will stop using it. On first glance I would say that the comment is correct, but that the risk is pretty low so doesn't need to be changed.

dupdob:
keep using it, it looks interesting, at least..

jacinyan:
@dupdob @rouke-broersma Just checked in, and saw it merged already. Really appreciate your work. BTW, what model and version is the current copilot on?

rouke-broersma:
image


🤷‍♂️

jacinyan:
> image
> 
> 🤷‍♂️

😂👌

PR #3252, opened 2025-06-04
fix: adjust RegexMutator to handle new parsing orchestration. fix rel…

### LINKED ISSUE #3241: Regex is always mutated into uncompilable code
**Describe the bug**
I have a simple regex in my code, all mutations are invalid as the constructor is removed in the result.

```csharp
public bool TestInput(string input)
{
    var regex = new Regex(@"^[0-9]{3,}$");

    return regex.IsMatch(input);
}
```

**Mutations**

There are multiple, all causing the same compilation error, but 3 examples:

Regex anchor removal mutation:

```diff
-   var regex = new Regex(@"^[0-9]{3,}$");
+   var regex = "^[0-9]{3,}"
```

Regex greedy quantifier quantity mutation:

```diff
-   var regex = new Regex(@"^[0-9]{3,}$");
+   var regex = "^[0-9]{4,}"
```

 Regex character class range modification:

```diff
-   var regex = new Regex(@"^[0-9]{3,}$");
+   var regex = "^[0-8]{3,}$"
```

Tested both with `var` and explicit type.

**Expected behavior**
No compilation errors.

 - OS: Windows
 - Type of project: .NET
 - Framework Version: 9.0
 - Stryker Version: 4.5.1

dupdob:
Ensure RegexMutator generates correct mutations. See below for explanation
Fixes #3241.

#### Details
PR #3116 implied an important change for mutators: now they must provide a mutated version of the full node that they received for evaluation. This change should have been applied to RegexMutator as well, but it was not done and this was not detected by existing tests.

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutators/RegexMutatorTest.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutators/RegexMutatorTest.cs
@@ -29,7 +29,15 @@ public class RegexMutatorTest : TestBase
         var mutation = result.ShouldHaveSingleItem();
 
         mutation.DisplayName.ShouldBe("Regex anchor removal mutation");
-        mutation.ReplacementNode.ToString().ShouldBe("new Regex(\"abc\")");
+        var replacementNode = mutation.ReplacementNode as InvocationExpressionSyntax;
+        replacementNode.ShouldNotBeNull();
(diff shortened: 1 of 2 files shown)


dupdob:
Quality gate fails due to a wrong report about changed lines not being covered by test. No other pb

rouke-broersma:
Why is code coverage collection with sonarsource so difficult 🫠

PR #3253, opened 2025-06-04
Support filtering of implicit constructors

willseward:
Currently constructors are filtered using the following syntax in the "method-filters" config: "*{TypeGlob}.ctor". However, this only matches constructors which are called using the full type name in the call, i.e. `var x = new string("asdf")`.

This PR enables filtering of constructors which use the implicit object creation expression syntax: `string x = new("asdf")`.


dupdob:
Thanks for this PR providing an improvement. 
I was about to comment it to be partially solving the issue when I realized you were perfectly aware of that.

dupdob (approved):

PR #3258, opened 2025-06-12
fix: Safe parsing of implicit new in IgnoreMutantFilter

### LINKED ISSUE #3257: Stryker throws null reference exception since version 4.6.0
**Describe the bug**
Since update 4.6.0 Stryker gives me a null reference exception. This seems to happen because of the newly added support of filtering implicit constructors. See also 

As far as i can gather it does not handle the implicit object creation correctly. For example when the implicit object creation is not inside a variable declaration.

For example `DoSomething(new("test"));` will be null for `creation.FirstAncestorOrSelf<VariableDeclarationSyntax>()` since there is no variable declaration. This than gives a null reference exception.

See also https://github.com/stryker-mutator/stryker-net/commit/52a99616832f021bd7bae5c27e846ddba3739601#diff-81655880c375520982183e68e29cb430c74cc9f6d7efe0669d5af8ce7be81debR49

**Logs**
Image 

**Expected behavior**
Something like a null check on `creation.FirstAncestorOrSelf<VariableDeclarationSyntax>()`

 - OS: [e.g. Windows]
 - Type of project: core
 - Framework Version: .NET 8
 - Stryker Version: 4.6.0

**Additional context**

```
public class Example
{
    public Example(string value) { }

    public static Example Create() => new("test"); // explicit, should be handled correctly

    public static Example CreateImplicit() => new("test"); // implicit, but still in a return statement, should be handled correctly

    public void DoSomething(Example ex) { }

    public void CallWithImplicit()
    {
        DoSomething(new("test")); // <-- implicit object creation, not in a variable declaration, this is probably failing
    }
}
```

dupdob:
IgnoreMutantFilter assumes any implicit new construct is part of variable declaration which leads to NRE when this is not the case.
This PR fixes this problem and regression test for this.

Fixes #3257

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/IgnoredMethodMutantFilterTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/MutantFilters/IgnoredMethodMutantFilterTests.cs
@@ -265,6 +265,35 @@ public class IgnoredMethodMutantFilter_NestedMethodCalls
     }
 
     [TestMethod]
+    public void MutantFilter_WorksWithImplicitNew()
+    {
+        // Arrange
+        var source = @"
+public class IgnoredMethodMutantFilter_NestedMethodCalls
+{
+    private void TestMethod()
+    {
+        Dispose(new(""anchor""));
+    }
+}";
+        var options = new StrykerOptions
+        {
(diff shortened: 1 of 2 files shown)


rouke-broersma:
Sorry it took a while to review this, thanks for the fix!

Issue #3260, opened 2025-06-19
Forward msbuild path option to Buildalyzer


dupdob:
**Is your feature request related to a problem? Please describe.**
Buildalyzer uses its own logic to identify which MsBuild version to use to analyze a project.
While Stryker offers an option to specify which MsBuild version to use (`msBuild-path`), this has no effect on Buildalyzer, which severely limits its value.

**Describe the solution you'd like**
The MsBuild executable defined in configuration should be forwardede to Buildalyzer.

**Describe alternatives you've considered**
None identified

**Additional context**
This was found when working on #3144

dupdob:
While working on it, I found out that MsBuild related tests are flaky as they depend on the local machine installs.
I need to add mocking there to have a controlled environment?

Issue #3144, opened 2024-12-19
TargetFramework cannot be set in Directory.Build.props


Liversage:
**Describe the bug**

When `<TargetFramework>` is set in `Directory.Build.props` Stryker doesn't work.

**Logs**

I have a a solution folder with a `Directory.Build.props` that specifies `<TargetFramework>net9.0</TargetFramework>` and there are no `<TargetFramework>` in the `.csproj` files. The project to mutate and the unit test project are in subdirectories and the unit test project has a reference to the project to mutate. When executing `dotnet stryker --verbosity debug` in the unit test project I get the following console output:

<pre>
[13:07:16 INF] Logging enabled at level Debug
Version: 4.4.1
[13:07:16 DBG] Stryker starting, version: 4.4.1

[13:07:16 DBG] Stryker started with options: [REMOVED FOR BREVITY]
[13:07:16 INF] Analysis starting.
[13:07:16 DBG] Using C:\Users\[REDACTED]\[REDACTED].UnitTests\[REDACTED].UnitTests.csproj as test project
[13:07:16 INF] Analyzing 1 test project(s).
[13:07:16 DBG] Analyzing 0 projects.
[13:07:16 DBG] Analyzing [REDACTED].UnitTests.csproj
[13:07:22 DBG] Analysis of project [REDACTED].UnitTests.csproj succeeded.
[13:07:23 DBG] Analyzing ..\[REDACTED]\[REDACTED].csproj
[13:07:29 DBG] Analysis of project ..\[REDACTED]\[REDACTED].csproj succeeded.
[13:07:29 INF] Analysis complete.
[13:07:29 INF] Time Elapsed 00:00:13.2623456
Stryker.NET failed to mutate your project. For more information see the logs below:

No project references found. Please add a project reference to your test project and retry.
</pre>

**Expected behavior**

According to the logs Stryker correctly detects the project to mutate and should go ahead and perform the tests.

Adding `<TargetFramework>net9.0</TargetFramework>` to both `.csproj` files fixes the problem.

**Desktop**
 - OS: Windows
 - Type of project: "core"
 - Framework Version: .NET 9.0
 - Stryker Version: 4.4.1

**Additional context**

I haven't tried to locate the actual problem in the source code. Even if it's not worth fixing it might be helpful to explain this in the documentation. This is my first time using Stryker and it took me quite some time to understand why my project didn't work with Stryker despite following the clear and useful instructions in "Getting started".

[...]

PR #3262, opened 2025-06-23
feat(build): Improve targetframework and msbuild-path options

dupdob:
This PR aims at improving Stryker support for complex setup/project situations. As such it:
1. Add a project analysis attempt forcing the target framework when normal attempts failed. This works only when a target framework is defined
2. Ensure Buildalyzer uses the msbuild executable set via the `--msbuild-path` option. This is useful when buildalyzer fails to automatically select the most suitable msbuild version

Issues:
Fixes: #3260
Fixes: #3144

Note: there is unit test for (2) (Buildalyzer using the --msbuild-path`) as its difficult to add. But code is straightforward and I performed manual end-to-end tests on this feature

### DIFF
--- a/src/Stryker.CLI/Stryker.CLI.UnitTest/packages.lock.json
+++ b/src/Stryker.CLI/Stryker.CLI.UnitTest/packages.lock.json
@@ -900,13 +900,13 @@
           "Serilog.Sinks.Console": "[6.0.0, )",
           "ShellProgressBar": "[5.2.0, )",
           "Spectre.Console": "[0.50.0, )",
-          "Stryker.Abstractions": "[4.6.0, )",
-          "Stryker.Configuration": "[4.6.0, )",
-          "Stryker.DataCollector": "[4.6.0, )",
+          "Stryker.Abstractions": "[1.0.0, )",
+          "Stryker.Configuration": "[1.0.0, )",
+          "Stryker.DataCollector": "[1.0.0, )",
           "Stryker.Regex.Parser": "[1.0.0, )",
-          "Stryker.RegexMutators": "[4.6.0, )",
-          "Stryker.TestRunner.VsTest": "[4.6.0, )",
-          "Stryker.Utilities": "[4.6.0, )",
+          "Stryker.RegexMutators": "[1.0.0, )",
+          "Stryker.TestRunner.VsTest": "[1.0.0, )",
+          "Stryker.Utilities": "[1.0.0, )",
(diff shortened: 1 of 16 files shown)


dupdob:
Looks like I am having a hard time with package lock files. For some reason, I never get the proper checksums on my machine, despite having reinstalled everything more than once (including Windows). Maybe it's because I using an emulator. I usually work around the issue by simply never modifying the dependencies, but here I needed to add support for mockfilesystem....

rouke-broersma:
@dupdob I'll see if I can correct the package lock files for you. They're not very consistent in nuget unfortunately, there are many edge cases that result in incorrect 'local' hashes depending on local environment. I've reduced most of them with msbuild config but clearly not all. 

In this case I think the issue was actually caused by adding the package `System.IO.Abstractions` instead of the renamed `TestableIO.System.IO.Abstractions.Wrappers` and since we don't use the old package there was also no package version defined in `Directory.Packages.Props`.

rouke-broersma:
A unit test is failing:

```
Error Message:
   Test method Stryker.Core.UnitTest.Initialisation.InitialBuildProcessTests.InitialBuildProcess_WithPathAsBuildCommand_ShouldThrowStrykerInputExceptionOnFailWithQuotes threw exception: 
Shouldly.ShouldAssertException: Should.Throw<InputException>(target.InitialBuild(true, null, _cProjectsExampleCsproj, null, @"C:\Program Files\Microsoft Visual Studio\2022\Professional\MSBuild\Current\Bin\MSBuild.exe"))
            .Details
    should be
"Initial build of targeted project failed. Please make sure the targeted project is buildable. You can reproduce this error yourself using: ""C:\Program Files\Microsoft Visual Studio\2022\Professional\MSBuild\Current\Bin\MSBuild.exe" Example.csproj""
    but was
"Initial build of targeted project failed. Please make sure the targeted project is buildable. You can reproduce this error yourself using: ""C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\MSBuild.exe" Example.csproj""

```

dupdob:
thanks, and sorry about that
looking into it

dupdob:
~So, test work on my machine (of course). Which means there is part of the msbuild search logic that still directly access the filesystem and bypass the mock.~
Nahh, it does not work; it is just that I was stupid enough to hide failing test(s) in the test explorer. My only excuse is that I am used to rely on NCrunch real time results and it no longer works on Stryker for unidentified reasons.

Sorry about that

Issue #3234, opened 2025-05-05
Since Feature Ignoring All Mutants


StevenAngel-ai:
**Describe the bug**
When attempting to utilize the since feature to see if writing some tests killed any mutants, it ignored all mutants as "Mutant not changed compared to target commit"
From the docs, it would appear that the since feature should detect the code coverage change and re-run the affected mutants. "* For changes on test project files all mutants covered by tests in that file will be seen as changed."
Playing around further, I started making and committing changes in the file I was testing and it still ignored all mutants for the same reason. I can see in the output that Stryker sees that the files were changed. Unless I am missing something, it seems that the Since feature ignores all mutants.

**Logs**

[logs.zip](https://github.com/user-attachments/files/20039390/logs.zip)

**Expected behavior**
When running Stryker with --since:develop it should generate a report based only on mutants/test coverage that changed.

**Desktop (please complete the following information):**
 - OS: [Windows]
 - Type of project [NET]
 - Framework Version [8.0]
 - Stryker Version [4.5.1]

kurnakovv:
I have same problem + I try to use `--with-baseline` flag, and it does not work, no errors, just this output
```
[04:29:40 INF] 1     mutants got status CompileError. Reason: Mutant caused compile errors
[04:29:40 INF] 23    mutants got status Ignored.      Reason: Mutant not changed compared to target commit
[04:29:40 INF] 24    total mutants are skipped for the above mentioned reasons
[04:29:40 INF] 0     total mutants will be tested
[04:29:40 WRN] It looks like all mutants with tests were ignored. Try a re-run with less ignoring!
[04:29:40 WRN] It looks like all mutants resulted in compile errors. Mutants sure are strange!
```

commands:
```
Run dotnet stryker -f .config/stryker-config.json --since:main --dashboard-api-key *** --version stryker-test2-3 --reporter dashboard --reporter html
```
and
```
Run dotnet stryker -f .config/stryker-config.json --with-baseline:main --dashboard-api-key *** --version stryker-test2-3 --reporter dashboard --reporter html
```

@dupdob @rouke-broersma is there any news about this problem? :)

rouke-broersma:
There is no indication that there is any problem. The logs seem pretty clear to me. There were no changed mutants. If you think that there were changes that were not detected by stryker please provide sufficient evidence that this is the case.

rouke-broersma:
@StevenAngel-ai I inspected your logs and I find the following information:

[...]

PR #3283, opened 2025-08-15
fix(filter): materialize mutants to prevent deferred where clause issue in SinceMutantFilter

chinchala:
This PR fixes SinceMutantFilter problem described here:
https://github.com/stryker-mutator/stryker-net/issues/3234#issuecomment-3166953462


kurnakovv:
@dupdob Can you review these changes? They are very important for us :)

PR #3292, opened 2025-08-28
Fix nowarn special character inclusion in diagnostic options dictionary

### LINKED ISSUE #3291: NoWarn diagnostic ids not having special chars trimmed
**Describe the bug**
I have a lot of NoWarn attributes in Directory.props file which were not respected during stryker re-compilation (`CSharpCompilingProcess.TryCompilation` and `CSharpRollbackProcess` flows). This led to multiple unrecoverable errors existing during build, causing `updatedSyntaxTree == originalTree` in `CSharpRollbackProcess:71` and subsequent fatal error during execution.

After investigation, my understanding of the issue is that NoWarn suppressions from csproj/Directory.props files do not trim properly beginning/ending spaces and newlines, causing mismatch on evaluation and unexpected errors appearing in stryker post-mutation build.

**Logs**
E.g. for a set of nowarn attributes authored as such:

Image

the current behavior interprets the following directory members (indicative, not the exact same values):

Image

**Expected behavior**
Each nowarn attribute should have trimmed special chars and whitespaces to be properly picked up as compilation option.

dgounaris:
Fixes #3291

### DIFF
--- a/src/Stryker.Utilities/Buildalyzer/IAnalyzerResultExtensions.cs
+++ b/src/Stryker.Utilities/Buildalyzer/IAnalyzerResultExtensions.cs
@@ -5,6 +5,7 @@ using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Text.RegularExpressions;
 using Buildalyzer;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
@@ -259,6 +260,7 @@ Generated source code may be missing.", analyzer);
 
         return diagnostics
             .Split(";")
+            .Select(x => Regex.Replace(x, "[^a-zA-Z0-9_.]+", "", RegexOptions.Compiled))
             .Distinct()
             .Where(x => !string.IsNullOrWhiteSpace(x));
     }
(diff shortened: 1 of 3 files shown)


src/Stryker.Utilities/Buildalyzer/IAnalyzerResultExtensions.cs
```diff
@@ -259,6 +260,7 @@ private static IEnumerable<string> ParseDiagnostics(string diagnostics)
 
         return diagnostics
             .Split(";")
+            .Select(x => Regex.Replace(x, "[^a-zA-Z0-9_.]+", "", RegexOptions.Compiled))
```
rouke-broersma: Would stripping of newline and whitespace characters not be sufficient? I personally try to avoid regex whenever possible. This currently seems to strip any non-alphanumeric character and I think we are doing too much there. non-alphanumeric characters except newlines and `;` are invalid characters and we should not be more forgiving in that regard that msbuild itself imo. Let me know if I'm missing something here.

dgounaris: @rouke-broersma  Fair point, modified the clause accordingly.

PR #3301, opened 2025-09-15
Skip --open-report hint in CI

martincostello:
Do not output the hint for `--open-report` when `CI=true`.

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Reporters/DashboardReporter.cs
+++ b/src/Stryker.Core/Stryker.Core/Reporters/DashboardReporter.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Extensions.Logging;
@@ -51,7 +52,7 @@ public class DashboardReporter : IReporter
             {
                 _browser.Open(reportUri);
             }
-            else
+            else if(!string.Equals(Environment.GetEnvironmentVariable("CI"), bool.TrueString, StringComparison.OrdinalIgnoreCase))
             {
                 var aqua = new Style(Color.Aqua);
                 _console.WriteLine(
--- a/src/Stryker.Core/Stryker.Core/Reporters/Html/HtmlReporter.cs
+++ b/src/Stryker.Core/Stryker.Core/Reporters/Html/HtmlReporter.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.IO.Abstractions;
@@ -7,7 +8,6 @@ using Stryker.Abstractions;
 using Stryker.Abstractions.Options;
 using Stryker.Abstractions.ProjectComponents;
 using Stryker.Abstractions.Reporting;
-using Stryker.Core.ProjectComponents.TestProjects;
 using Stryker.Core.Reporters.Html.RealTime;
 using Stryker.Core.Reporters.Json;
 using Stryker.Core.Reporters.WebBrowserOpener;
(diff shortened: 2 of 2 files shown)


src/Stryker.Core/Stryker.Core/Reporters/DashboardReporter.cs
```diff
@@ -51,7 +52,7 @@ private void OpenDashboardReport(string reportUri)
             {
                 _browser.Open(reportUri);
             }
-            else
+            else if(!string.Equals(Environment.GetEnvironmentVariable("CI"), bool.TrueString, StringComparison.OrdinalIgnoreCase))
```
martincostello:
```suggestion
            else if (!string.Equals(Environment.GetEnvironmentVariable("CI"), bool.TrueString, StringComparison.OrdinalIgnoreCase))
```

src/Stryker.Core/Stryker.Core/Reporters/DashboardReporter.cs
```diff
@@ -51,7 +52,7 @@ private void OpenDashboardReport(string reportUri)
             {
                 _browser.Open(reportUri);
             }
-            else
+            else if (!string.Equals(Environment.GetEnvironmentVariable("CI"), bool.TrueString, StringComparison.OrdinalIgnoreCase))
```
rouke-broersma: This is one of possible dozens of CI hints. Do you know which CI systems do/don't set this information? [`https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#continuousintegrationbuild`](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#continuousintegrationbuild) is a well-known msbuild property to reflect this information but this might only be used if reproducible builds is used, I'm unsure what we should be using here. Do we need to start keeping a list of environment variables that indicate this is a CI build to be safe? If so we might copy the list that reproducible builds uses.

martincostello: I know that the following set `CI=(T|t)rue` from past experience:

- AppVeyor
- GitHub Actions
- Travis CI

It might not cover every possible CI system in existence, but it's a start. I didn't intend this to be bullet-proof, but instead be a lowest common denominator. I figured this would be lighter weight (and less burden on the end-user) than creating an entire new configuration setting and requiring the user to tell Stryker whether it was CI or not.

I'm not sure `ContinuousIntegrationBuild=true` would be relevant here, as this code is executing outside the context of MSBuild so wouldn't be available or set as an environment variable even if set by the user in their project?.

FWIW, here's the list used by [dotnet/reproducible-builds](https://github.com/dotnet/reproducible-builds/blob/318bdde09ab18e23eda6b5654a236c2cf3ad084c/src/DotNet.ReproducibleBuilds/DotNet.ReproducibleBuilds.props#L15-L38) to try and auto-detect - there's 11 different tests there (though I would argue that at least 3 of them are redundant as while they have their own custom environment variable they also support `CI`), and I imagine testing all of them is overkill here.

rouke-broersma: > I'm not sure ContinuousIntegrationBuild=true would be relevant here, as this code is executing outside the context of MSBuild so wouldn't be available or set as an environment variable even if set by the user in their project?.

We resolve msbuild properties so ContinuousIntegrationBuild=true would be available in msbuild properties if set in any way (for example by using reproducible builds). You don't need to drag that into this PR, we can do without for now. I would like to at least also support `TF_BUILD` since Azure Devops is a dotnet ecosystem native build system and from a quick search it does not look like they set CI=true. You're right that we don't need to be exhaustive for a first support, the rest can be added if we get requests for it.

martincostello: I can add `TF_BUILD` then.

Regarding, `ContinuousIntegrationBuild`, is the code I've changed actually executing _inside_ MSBuild where the value is available though?

rouke-broersma: No the reporter itself is not, but we we analyze msbuild properties from the user provided projects, so those will resolve any msbuild properties like ContinuousIntegrationBuild and we can read the property from there and make it available in IReadOnlyProjectComponent.

But you can disregard that, CI=true and TF_BUILD=true is sufficient.

PR #3354, opened 2025-11-19
Added CSharp14 version to expected message.

kudima03:
Added CSharp14 version to expected message in LanguageVersionInputTests.ShouldHaveHelpText test.

This pr fixes #3353 test error.

### DIFF
--- a/integrationtest/TargetProjects/NetCore/Generator/SourceGenerator.csproj
+++ b/integrationtest/TargetProjects/NetCore/Generator/SourceGenerator.csproj
@@ -7,7 +7,7 @@
   
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces"
-        Version="5.0.0"
+        Version="4.14.0"
         PrivateAssets="all" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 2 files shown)


kudima03:
@dupdob Integration mutation testing is failing most likely due to a version mismatch in **Microsoft.CodeAnalysis.Common**: the current branch uses **5.0.0**, while Stryker **4.8.1** (used for mutation testing of this project) depends on **4.14.0**.

If I'm right, probably we should ignore these errors.

Could you take a look when you have a minute?

kudima03:
Or i can downgrade version of Microsoft.CodeAnalysis.CSharp.Workspaces [here](https://github.com/kudima03/stryker-net/blob/d37b6ce383089b93482bbad0a810beb9be72f0a1/integrationtest/TargetProjects/NetCore/Generator/SourceGenerator.csproj#L8-L13) from 5.0.0 to previous 4.14 and bump it after Stryker release

kudima03:
If you decide not to downgrade, I'll revert last commit

rouke-broersma:
The integration test uses a prerelease version of Stryker built from the branch, so the workspace version conflict must be something else

PR #3575, opened 2026-05-10
fix(mutating): out var in lambdas triggering NRE

### LINKED ISSUE #3572: Delegate with Out Result Crashes Stryker During Mutation
**Describe the bug**

Given the following code:

```csharp
public static class Foo
{
    private delegate bool ExceptionResultDelegate(out bool result);
    
    public static void Thing()
    {
        var results = new List<ExceptionResultDelegate>();
        
        results.Add((out result) =>
        {
            result = false;
            return false;
        });
    }
}
```

During mutation, Stryker will crash:
```
Unhandled exception. System.AggregateException: One or more errors occurred. (Object reference not set to an instance of an object.)
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.CodeAnalysis.SyntaxNodeExtensions.WithLeadingTrivia[TSyntax](TSyntax node, IEnumerable`1 trivia)
   at Microsoft.CodeAnalysis.SyntaxNodeExtensions.WithoutLeadingTrivia[TSyntax](TSyntax node)
   at Microsoft.CodeAnalysis.SyntaxNodeExtensions.WithoutTrivia[TSyntax](TSyntax syntax)
   at Stryker.Core.Helpers.RoslynHelper.BuildDefaultExpression(TypeSyntax type) in /_/src/Stryker.Core/Stryker.Core/Helpers/RoslynHelper.cs:line 151
   at Stryker.Core.Instrumentation.DefaultInitializationEngine.<>c.<InjectOutParametersInitialization>b__1_1(ParameterSyntax p) in /_/src/Stryker.Core/Stryker.Core/Instrumentation/DefaultInitializationEngine.cs:line 57
   at System.Linq.Enumerable.SelectListIterator`2.MoveNext()
   at System.Linq.Enumerable.UnionIterator`1.MoveNext()
   at Microsoft.CodeAnalysis.SyntaxList`1.CreateNode(IEnumerable`1 nodes)
   at Microsoft.CodeAnalysis.SyntaxList`1..ctor(IEnumerable`1 nodes)
   at Microsoft.CodeAnalysis.CSharp.SyntaxFactory.List[TNode](IEnumerable`1 nodes)
   at Microsoft.CodeAnalysis.CSharp.SyntaxFactory.Block(IEnumerable`1 statements)
   at Stryker.Core.Instrumentation.DefaultInitializationEngine.InjectOutParametersInitialization(BlockSyntax body, IEnumerable`1 parameters) in /_/src/Stryker.Core/Stryker.Core/Instrumentation/DefaultInitializationEngine.cs:line 57
   at Stryker.Core.Mutants.MutantPlacer.InjectOutParametersInitialization(BlockSyntax block, IEnumerable`1 parameters) in /_/src/Stryker.Core/Stryker.Core/Mutants/MutantPlacer.cs:line 103
   at Stryker.Core.Mutants.CsharpNodeOrchestrators.BaseFunctionOrchestrator`1.InjectMutations(T sourceNode, T targetNode, SemanticModel semanticModel, MutationContext context) in /_/src/Stryker.Core/Stryker.Core/Mutants/CsharpNodeOrchestrators/BaseFunctionOrchestrator.cs:line 127
... abbreviated
```

This appears to be due to the fact that the `InjectOutParametersInitialization` not having a type for that parameter.  (not sure exactly *why* it doesn't figure out the type given that the delegate defines the signature for the type)

**Logs**

[log-20260508.txt](https://github.com/user-attachments/files/27534847/log-20260508.txt)

**Expected behavior**
A failure to mutate shouldn't crash the process, it should just trigger the standard rollback.  It was probably assumed that the type could never be null, but since #nullable isn't used it wasn't known that this was the case.

 - OS: Windows (W11) & Linux (Ubuntu)
 - Type of project: core
 - Framework Version 10.0
 - Stryker Version 4.14.1 (also tested on the HEAD commit from master in case it had been fixed between releases)

**Additional context**

dupdob:
Improve `out` variables initialization to default to handle situation where type is not explicitly defined in function signatures, such as in lambdas.
It then uses the `default` keyword instead of `default(type)` format. Note this will fail if project uses C# version 7 or less.

Fixes #3572

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/CsharpMutantOrchestratorTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/CsharpMutantOrchestratorTests.cs
@@ -1029,11 +1029,46 @@ static Mutator_Flag_MutatedStatics()
     [TestMethod]
     public void ShouldInitializeOutVars()
     {
-        var source = @"public void SomeMethod(out int x, out string text) { x = 1; text = ""hello"";}";
-        var expected = @"public void SomeMethod(out int x, out string text) {{x= default(int);text= default(string);}if(StrykerNamespace.MutantControl.IsActive(0)){}else{ x = 1; text = (StrykerNamespace.MutantControl.IsActive(1)?"""":""hello"");
-        }}";
+        const string Source = """public void SomeMethod(out int x, out string text) { x = 1; text = "hello";}""";
+        const string Expected = """
+                                public void SomeMethod(out int x, out string text) {{x= default(int);text= default(string);}if(StrykerNamespace.MutantControl.IsActive(0)){}else{ x = 1; text = (StrykerNamespace.MutantControl.IsActive(1)?"":"hello");
+                                        }}
+                                """;
 
-        ShouldMutateSourceInClassToExpected(source, expected);
+        ShouldMutateSourceInClassToExpected(Source, Expected);
+    }
+
+    [TestMethod]
+    public void ShouldInitializeOutVarsForLambdas()
+    {
+        const string Source = """
+                              public static void Thing()
+                              {
+                                  var results = new List<ExceptionResultDelegate>();
+
+                                  results.Add((out result) =>
+                                  {
+                                      result = false;
+                                      return false;
+                                  });
+                              }
+                              """;
+        const string Expected = """
+                                public static void Thing()
+                                {if(StrykerNamespace.MutantControl.IsActive(0)){}
+                                else{
+                                    var results = new List<ExceptionResultDelegate>();
+
+                                    if(StrykerNamespace.MutantControl.IsActive(1)){;}else{results.Add((out result) =>
+                                {{result= default;}    if(StrykerNamespace.MutantControl.IsActive(2))    {}else{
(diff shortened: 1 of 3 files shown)


iamwyza:
@dupdob Any idea when this might be able to be finished up and merged?

rouke-broersma:
@dupdob only supporting C# 7.1 and up when using mutators that require placing the return default construct seems like a pretty big breaking change. Do you think it's justifiable to introduce this breaking change in this way? Or is the chance that a user using < C# 7.1 would hit that issue very low you think?

dupdob:
@rouke-broersma I understand your question and I wondered myself. 
But thinking about it, the answer is pretty simple: risk is very low. 
First of all, C# 7.1 is more than 10 years old. But most importantly, the current status is that Stryker crashes when it encounters this syntax construct, whatever C# version is defined. 
So, at worst, this means this fixes the issue for C# 7.2 + and older version get a compilation error instead. 
But I am not even sure anonymous lambda were available then.

dupdob:
Update: using parameter modifiers without specifying the type is a C#14 feature. So the issue does not exist on earlier C# version. I should have checked this first.
So it is definitely not a problem, `(out x) => ....` being legal implies `x = default;` being legal as well.

PR #3578, opened 2026-05-12
Almost fixes #3572.  Issue with the new test that I haven't sorted out yet.

### LINKED ISSUE #3572: Delegate with Out Result Crashes Stryker During Mutation

iamwyza:
Digging through https://github.com/stryker-mutator/stryker-net/issues/3572 it appears that there may not be a way to actually get the property type in the scenario of the lambda function in that scenario.  Even looking through the `SemanticModel` that is up in the `BaseFunctionOrchestrator` doesn't help, it still resolves the symbol as a `ErrorTypeSymbol (?)`.  So in this scenario, I think the only viable option is to bail out if any of the parameters' types are null. 

This is that change.  However the unit test I added for this doesn't work yet.  Getting weird results when trying to compare the before/after.  They are identical as far as I can tell, but the comparison says they are different.  Will continue to dig into this tomorrow.

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/MutantPlacerTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutants/MutantPlacerTests.cs
@@ -177,7 +177,6 @@ public class MutantPlacerTests : TestBase
         CheckMutantPlacerProperlyPlaceAndRemoveHelpers<PropertyDeclarationSyntax>(source, expected, placer.ConvertToBlockBody);
     }
 
-
     [TestMethod]
     public void ShouldInjectInitializersAndRestore()
     {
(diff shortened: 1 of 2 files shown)


dupdob:
I opened #3575 that fixes this a couple of days ago. Forgot to update the issue to reflect it. Note that said PR does fully support typeless parameters for lambdas

PR #3595, opened 2026-05-19
fix(mutating): avoid uncompilable mutations in several edge cases

### LINKED ISSUE #3594: A handful of Stryker compilation failures on a modern ASP.NET Core (.NET 10) project — with a repro and proposed fixes
Hi Stryker team! 👋

First off — thanks for all the work you put into Stryker.NET. I ran into a cluster of compilation failures while trying to introduce Stryker into a fairly small ASP.NET Core project, and after digging in I think they're all genuine bugs (plus one root cause that explains why a couple of the suggested fixes only half-worked when I tried them).

I've put together a minimal repro here: **https://github.com/slang25/stryker-repro** — all four issues reproduce with `dotnet stryker` from `StrykerRepro.Tests/`. The README in that repo walks through each bug in detail, including the cascade between #1 and #4.

### The bugs

**1. `LinqMutator` false positive on non-LINQ `Append`/`Prepend`**
`LinqMutator` pattern-matches on the method *name* only. `Microsoft.AspNetCore.Http.IResponseCookies.Append(name, value, options)` gets rewritten to `Prepend(...)`, which doesn't exist on that interface → **CS1501** in the instrumented compilation.

**2. CS9234 cascade from #1 on .NET 10**
On .NET 10, `RequestDelegateGenerator` emits checksum-based `[InterceptsLocation(version: 1, data: "…")]` attributes (SHA-256 of the source file plus a byte offset). After Stryker's first compilation fails on #1 and `CSharpRollbackProcess` rolls back the `Prepend` mutation, the file content changes but the generated `GeneratedRouteBuilderExtensions.g.cs` (with stale checksums) is reused → **CS9234** against the generated file, which the rollback process can't trace back to a source mutation → the whole run aborts. Fixing #1 makes this go away in the common case, but the underlying fragility around source generators + rollback is worth knowing about. On `net8.0` the older line/column interceptor format isn't checksum-sensitive, so it doesn't bite there.

**3. `ConditionalInstrumentationEngine` re-declares pattern variables**
`RelationalPatternMutator` mutations of `> 0` inside `is { Length: > 0 } region` get placed at expression level via a ternary, producing something like:

```csharp
if (active == 34 ? cfg is { Length: >= 0 } region
  : active == 33 ? cfg is { Length: < 0 } region
  :                cfg is { Length: > 0 } region) { ... }
```

Each branch re-declares `region` in the same scope → **CS0136**. `ContainsDeclarations()` already exists for the `DeclarationPattern`/`DeclarationExpression` cases, but it doesn't detect `SingleVariableDesignation` inside recursive/var patterns.

**4. `ObjectCreationMutator` + C# 11 `required` members**
The mutator always emits an empty-initializer variant (`new Notification {}`), which fails CS8852 when the type has `required` properties/fields.

### Why some of the \"suggested fix\" paths didn't fully work for me at first

While prototyping fixes I noticed that `LinqMutator`'s receiver-type check via the semantic model doesn't actually fire for the `IResponseCookies` case — and the reason turned out to be unrelated to `LinqMutator`. In `CsharpMutationProcess.Mutate`, the semantic models are built from `projectInfo.GetAllFiles()` (user source files only). On modern SDK projects this skips auto-generated trees added via `AddCompilationSyntaxTree` — most importantly `*.GlobalUsings.g.cs`. So any mutator that consults the semantic model silently sees `IErrorTypeSymbol` for anything brought in by `<ImplicitUsings>enable`. Worth a look as a standalone issue — it likely affects other mutators too.

### Proposed fixes

I've prototyped fixes for all four locally and `dotnet stryker` against the repro now completes cleanly with a ~74% mutation score. All existing Stryker.Core unit tests still pass. Happy to open one or more PRs if you'd like — wasn't sure whether you'd prefer a single PR or one per bug, so I figured I'd raise this first.

Rough shape of the fixes, for reference:

- **1 `LinqMutator`** — use the semantic model to verify the receiver implements `IEnumerable<T>` or that the method is declared on `System.Linq.Enumerable`/`Queryable`/`ParallelEnumerable` before substituting. Falls back to legacy name-only matching when no semantic model is available (preserves unit-test behavior).
- **3 `ConditionalInstrumentationEngine`** — extend `ContainsDeclarations()` to detect `SingleVariableDesignation`, and have `MutationStore.Inject(ExpressionSyntax, …)` forward mutations to an enclosing block when the source expression introduces a declaration.
- **4 `ObjectCreationMutator`** — skip the empty-initializer mutation when the resolved type (or any base) has a member with `IsRequired == true`.
- **Semantic-model fidelity** — include the project's `CompilationSyntaxTrees` (alongside the original file trees) when constructing semantic models in `CsharpMutationProcess.Mutate`.

### Environment

- Stryker: latest (built from `main`)
- .NET SDK: 10.0.300
- Target framework: `net10.0`
- OS: macOS

Thanks again, and happy to chat through any of this!

slang25:
## Summary
- Avoid CS0136 by skipping ternary mutation placement when the expression introduces a declaration (e.g. recursive patterns like `x is { } v`); mutations are forwarded to the enclosing block instead.
- Include auto-generated compilation trees (e.g. `GlobalUsings.g.cs`) when building semantic models so mutators that consult the semantic model see resolved types.
- Restrict `LinqMutator` to invocations whose receiver implements `IEnumerable<T>` (or whose containing type is `Enumerable`/`Queryable`/`ParallelEnumerable`), preventing incorrect rewrites of unrelated methods such as `IResponseCookies.Append`.
- Skip the `ObjectCreationMutator` initializer-removal mutation when the target type has any `required` members (CS9035).

Fixes #3594

## Test plan
- [ ] Run existing unit test suite
- [ ] Verify mutation runs against a project using ASP.NET Core (`IResponseCookies.Append`) no longer produce CS-failure mutants
- [ ] Verify mutation runs against a project using `required` members no longer produce CS9035 mutants
- [ ] Verify mutation runs against a project using recursive patterns no longer produce CS0136 mutants

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Helpers/RoslynHelperTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Helpers/RoslynHelperTests.cs
@@ -1,0 +1,36 @@
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Shouldly;
+using Stryker.Core.Helpers;
+
+namespace Stryker.Core.UnitTest.Helpers;
+
+[TestClass]
+public class RoslynHelperTests : TestBase
+{
+    [TestMethod]
+    [DataRow("s is { Length: > 0 } region")]
+    [DataRow("s is var captured")]
+    [DataRow("s is int n")]
+    public void ContainsDeclarationsShouldDetectPatternDesignations(string expression)
+    {
+        SyntaxFactory.ParseExpression(expression).ContainsDeclarations().ShouldBeTrue();
+    }
+
+    [TestMethod]
+    public void ContainsDeclarationsShouldDetectOutVarDeclaration()
+    {
+        SyntaxFactory.ParseExpression("int.TryParse(s, out var value)")
+            .ContainsDeclarations().ShouldBeTrue();
+    }
+
+    [TestMethod]
+    [DataRow("s.Length > 0")]
+    [DataRow("s is [_, _]")]
+    [DataRow("(a, b)")]
+    public void ContainsDeclarationsShouldReturnFalseWithoutDeclarations(string expression)
+    {
+        SyntaxFactory.ParseExpression(expression).ContainsDeclarations().ShouldBeFalse();
+    }
+}
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Mutators/LinqMutatorTest.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Mutators/LinqMutatorTest.cs
@@ -1,4 +1,5 @@
 using System.Linq;
+using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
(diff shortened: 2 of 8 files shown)


slang25:
Closing per maintainer feedback to split into focused PRs. Replaced by:

- #3600 — Include generated trees when building semantic models (#3599)
- #3601 — LinqMutator: only mutate calls on `IEnumerable<T>`-shaped receivers (#3596)
- #3602 — Avoid CS0136 from recursive/var pattern designations (#3597)
- #3603 — ObjectCreationMutator: skip empty initializer for `required` members (#3598)

Bug #2 from the original write-up (CS9234 interceptor cascade) is largely resolved by #3601 in the common case; I've folded it into the #3601 narrative rather than tracking it separately.

PR #3603, opened 2026-05-19
fix(object-creation-mutator): skip empty initializer for required members

### LINKED ISSUE #3598: ObjectCreationMutator empty-initializer mutation fails CS9035 on types with required members
Split out from #3594.

`ObjectCreationMutator` always emits an empty-initializer variant (`new Notification {}`), which fails CS9035 (\"Required member 'X' must be set in the object initializer or attribute constructor\") when the type has C# 11 `required` properties/fields.

**Proposed fix:** skip the empty-initializer mutation when the resolved type (or any base) has a member with `IsRequired == true`.

slang25:
## Summary
- `ObjectCreationMutator` always emits an empty-initializer variant (e.g. `new Notification {}`), which fails to compile with CS9035 when the target type has C# 11 `required` properties/fields.
- Use the semantic model to walk the resolved type and its base types; if any member has `IsRequired == true`, skip the empty-initializer mutation.

Fixes #3598. Split out from #3594.

## Test plan
- [ ] Existing `ObjectCreationMutatorTests` cases still pass
- [ ] New tests cover required property, required field, base-type required member, and a no-required-members baseline that still mutates

### DIFF
--- a/integrationtest/Validation/ValidationProject/ValidateStrykerResults.cs
+++ b/integrationtest/Validation/ValidationProject/ValidateStrykerResults.cs
@@ -81,7 +81,7 @@ public class ValidateStrykerResults
 
         var report = await strykerRunOutput.DeserializeJsonReportAsync();
 
-        CheckReportMutants(report, total: 660, ignored: 271, survived: 4, killed: 9, timeout: 2, nocoverage: 340);
+        CheckReportMutants(report, total: 656, ignored: 271, survived: 4, killed: 9, timeout: 2, nocoverage: 340);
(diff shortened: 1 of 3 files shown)


dupdob:
As said in the related issue, I am afraid I don't think this filtering logic adds enough value to warrant a merge.
But it also gave me an idea:
Overall the intent of Stryker is to maximize the number of significant mutations. As you said elsewhere, Stryker fails to mutate creating of objects/structs/records when they have required members.

Instead of non mutating them, why not using your detection logic to initialize said required members to default value?

I simply wonder if those would be sensible mutations.

What do you think?

PR #3606, opened 2026-05-19
fix(compilation): skip source-generator output files to prevent Blazo…

### LINKED ISSUE #3605: Support running mutation testing on projects that depend on Blazor (handle Razor source-generator output gracefully)
## Problem

When running Stryker.NET against a test project whose **source project is a Blazor application** (WebAssembly or Server), the tool crashes with an unhandled `CompilationException` after spending several minutes trying to compile mutated code:

```
Unhandled exception. System.AggregateException: One or more errors occurred. (Internal error due to compile error.)
 ---> Stryker.Abstractions.Exceptions.CompilationException: Internal error due to compile error.
   at Stryker.Core.Compiling.CSharpRollbackProcess.Start(...)
   at Stryker.Core.Compiling.CsharpCompilingProcess.TryCompilation(...)
   ...
```

### Root cause

Blazor projects use the **Razor source compiler** (a Roslyn `ISourceGenerator`) to generate `*.razor.g.cs` files from `*.razor` component files at build time. These generated files:

1. Are listed in the project's effective `SourceFiles` by Buildalyzer (because the compiler includes them in the `Compile` item group for downstream tools).
2. Carry an `// <auto-generated/>` header, so Stryker correctly identifies them as "generated" and adds them to `compilationSyntaxTrees` (no mutation, but included for compilation).
3. Are **also regenerated** when Stryker calls `RunSourceGenerators` during its own compilation phase.

This results in **duplicate partial-class declarations** for every Blazor component — one from the on-disk `*.razor.g.cs` file and one freshly produced by the Razor source generator. The C# compiler errors out on duplicate members. Since no Stryker mutations exist in those generated trees, the rollback process cannot identify anything to remove and exhausts all 50 retry attempts before throwing.

### Concrete reproduction

Run Stryker.NET with a test project that exercises a Blazor source project containing `.razor` components:

```
dotnet run --project Stryker.CLI/Stryker.CLI.csproj -- \
  --test-project "path/to/MyBlazorApp.Tests/MyBlazorApp.Tests.csproj"
```

Expected: mutation testing runs normally.  
Actual: `CompilationException: Internal error due to compile error.` after ~5 minutes.

Example repository to run dotnet stryker: https://github.com/acamacho-sprbrk/dotnet-stryker-blazor-project/tree/main

---

## Proposed solution

Three coordinated fixes (already prototyped in the linked PR):

### 1. Classify `*.g.cs` files as **source-generator output** (not just "generated")

Extend `GeneratedCodeFilterExtension` with an `IsSourceGeneratorOutput()` method that detects files ending in `.g.cs` — the standard naming convention for all Roslyn source generator output (`*.razor.g.cs`, `*.g.cs`, etc.).

### 2. Skip source-generator output files entirely during project component building

In `CsharpProjectComponentsBuilder.FindProjectFilesUsingBuildalyzer`, when a file is both `IsGenerated()` **and** `IsSourceGeneratorOutput()`, skip it completely (do not add it to `compilationSyntaxTrees`). The source generator will regenerate the correct trees during `RunSourceGenerators`. Static generated files (`.designer.cs`, etc.) continue to be added to `compilationSyntaxTrees` as before.

> This removes the duplicate-declaration root cause entirely.

### 3. Make the rollback process resilient to errors in un-mutated generator trees

As defence-in-depth, when `CSharpRollbackProcess` finds compilation errors in a tree that contains **no Stryker mutations** (e.g., a freshly generator-added tree), instead of immediately throwing it now removes the problematic tree from the compilation and continues. This allows subsequent retry attempts to succeed once the causally-responsible user-code mutation has been rolled back in a different tree. The `LastChance` throw is preserved as the final safety net.

### User-facing configuration (immediate workaround)

Until a fix is released, users can add a `stryker-config.json` in their test project directory to exclude Razor files from mutation:

```json
{
  "stryker-config": {
    "mutate": [
      "**/*",
      "!**/*.razor",
      "!**/*.razor.cs"
    ]
  }
}
```

`!**/*.razor` — excludes Razor markup files from mutation candidates.  
`!**/*.razor.cs` — optionally excludes Blazor codebehind files (partial classes tightly coupled to Razor-generated code); remove this line if you want those mutations tested.

---

## Alternatives considered

- **Suppress additional Razor error codes in `IgnoredErrors`** — only addresses symptoms of generator failures, not the duplicate-declaration root cause.
- **Re-run source generators after every rollback step** — correct in theory but architecturally expensive and not justified given the simpler fix above.
- **Never run the Razor source generator during Stryker compilation** — would break Blazor component rendering and make test execution meaningless.

---

## Additional context

- Affects both Blazor WebAssembly and Blazor Server projects.
- The existing `IgnoredErrors = ["RZ3600", "CS8784"]` in `CsharpCompilingProcess` shows the team has already encountered Razor-related compilation issues; this is a more fundamental instance of the same class of problem.
- `.designer.cs` files (WinForms/WPF) are **not** affected — they are not regenerated by source generators and correctly remain in `compilationSyntaxTrees`.

acamacho-sprbrk:
…r compilation errors

Blazor projects carry *.razor.g.cs files in their effective SourceFiles list (produced by the Razor source compiler). Stryker previously added those files to compilationSyntaxTrees as non-mutated compilation input, then also re-ran the Razor source generator which creates the same partial-class declarations again. This produced duplicate-member compilation errors that the rollback process could not fix (no Stryker mutations exist in generator-added trees), causing an unhandled CompilationException after exhausting all 50 retry attempts.
- GeneratedCodeFilterExtension: extend IsGeneratedFileName regex to cover *.g.cs; add public IsSourceGeneratorOutput() extension on SyntaxTree (and a string overload) to distinguish dynamic source-generator output from static generated files (.designer.cs)
- CsharpProjectComponentsBuilder: when a file is both IsGenerated() and IsSourceGeneratorOutput(), skip it entirely instead of adding it to compilationSyntaxTrees; RunSourceGenerators will regenerate the correct tree from the current (mutated) compilation state
- CSharpRollbackProcess: when rollback finds errors in a tree that contains no Stryker mutations, remove that tree from the compilation and continue rather than throwing immediately; provides defence-in- depth for other generators that may exhibit the same pattern while keeping LastChance mode as the true terminal failure path Fixes #3605


dupdob:
This is not how to fix this & I have found a proper fix, will open a PR in the near future

PR #3616, opened 2026-06-05
feat(baseline)!: baseline disk provider now follows output path option

### LINKED ISSUE #3615: Allow configuring the disk baseline output directory (--baseline-output)
When using the disk baseline provider, the baseline report path is hard-coded to <projectPath>/StrykerOutput/<version>/stryker-report.json and ignores the configured --output / output value, even though reports and logs already honor --output. Since projectPath is just the current working directory, the only practical way to relocate the baseline today is to launch Stryker from a different directory — which also changes project discovery, config resolution, and where reports and logs end up. The real pain is that this forces the baseline to live inside the test project's directory, where, without a non-obvious workaround, Stryker flags it as a change to the test project and ignores the baseline, thus defeating the purpose of the feature in the first place.

I'd like to add an optional --baseline-output argument (config: baseline.output) that lets the disk provider store and load the baseline from a configurable directory.  My goal is to maintain backward compatibility and minimize the blast radius so I did not touch Azure or S3 (but can if you wish).

The non-obvious workaround is to add the following to the stryker config.json .  This feels like more of a fortunate accident that this works rather than an intentional design (since it's stashed in the 'since' section which is suppoesd to be mutually exclusive from baseline), so I'm trying to provide a more intentional solution.

"since": { "enabled": false, "ignore-changes-in": [ "**/StrykerOutput/**" ]

**Additional context**
I've got a PR coming that implements this feature.  Unit tests added, verified working via manual testing also.

mpownby:
The disk baseline provider always writes the baseline to a hard-coded "StrykerOutput" folder under the project path. Storing the baseline inside the test project's directory is problematic: without a non-obvious workaround, Stryker flags it as a change to the test project and ignores the baseline.

The disk baseline provider now stores and loads baselines from the stable output root. If --output is supplied, the disk baseline follows that path. Without --output, the baseline remains under StrykerOutput, while individual run reports continue to use timestamped subfolders. Only the disk provider is affected; the Azure and S3 providers are untouched.

fix #3615 

BREAKING CHANGE: The disk baseline provider now follows the configured output path.

Previously, disk baselines were always stored under <project>/StrykerOutput, even when --output was configured. Disk baselines are now stored and loaded from the stable output root. With the default output this remains <project>/StrykerOutput. When --output is supplied, baselines are stored under that output directory.

If you use disk baselines in CI and cache/persist baseline files, update the cached path to match your configured --output value. Existing baselines are not migrated automatically. Stryker also writes .gitignore to the stable output root so baseline files and run outputs are ignored by default.

### DIFF
--- a/src/Stryker.Configuration/Options/Inputs/BaselineOutputInput.cs
+++ b/src/Stryker.Configuration/Options/Inputs/BaselineOutputInput.cs
@@ -1,0 +1,25 @@
+using Stryker.Abstractions.Baseline;
+using Stryker.Abstractions.Exceptions;
+
+namespace Stryker.Configuration.Options.Inputs;
+
+public class BaselineOutputInput : Input<string>
+{
+    protected override string Description => "The directory the disk baseline provider stores and loads the baseline report from. Required when the disk baseline provider is selected.";
+
+    public override string Default => string.Empty;
+
+    public string Validate(BaselineProvider baselineProvider, bool withBaseline)
+    {
+        if (withBaseline && baselineProvider == BaselineProvider.Disk)
+        {
+            if (string.IsNullOrWhiteSpace(SuppliedInput))
+            {
+                throw new InputException("BaselineOutput can't be null or whitespace");
+            }
+
+            return SuppliedInput;
+        }
+        return Default;
+    }
+}
--- a/src/Stryker.Configuration/Options/Inputs/BaselineOutputInput.cs
+++ b/src/Stryker.Configuration/Options/Inputs/BaselineOutputInput.cs
@@ -1,3 +1,4 @@
+using System.IO;
 using Stryker.Abstractions.Baseline;
 using Stryker.Abstractions.Exceptions;
 
@@ -16,6 +17,11 @@ public class BaselineOutputInput : Input<string>
             if (string.IsNullOrWhiteSpace(SuppliedInput))
             {
                 throw new InputException("BaselineOutput can't be null or whitespace");
+            }
+
+            if (!Path.IsPathFullyQualified(SuppliedInput))
+            {
+                throw new InputException($"BaselineOutput must be a fully qualified path, but was: {SuppliedInput}");
             }
 
             return SuppliedInput;
--- a/src/Stryker.Core/Stryker.Core/Baseline/Providers/DiskBaselineProvider.cs
+++ b/src/Stryker.Core/Stryker.Core/Baseline/Providers/DiskBaselineProvider.cs
@@ -16,7 +16,6 @@ public class DiskBaselineProvider : IBaselineProvider
     private readonly IStrykerOptions _options;
     private readonly IFileSystem _fileSystem;
     private readonly ILogger<DiskBaselineProvider> _logger;
-    private const string _outputPath = "StrykerOutput";
 
     public DiskBaselineProvider(IStrykerOptions options, IFileSystem fileSystem = null)
     {
@@ -29,7 +28,7 @@ public class DiskBaselineProvider : IBaselineProvider
     public async Task<IJsonReport> Load(string version)
     {
         var reportPath = FilePathUtils.NormalizePathSeparators(
-            Path.Combine(_options.ProjectPath, _outputPath, version, "stryker-report.json"));
+            Path.Combine(_options.ProjectPath, _options.BaselineOutputPath, version, "stryker-report.json"));
 
         if (_fileSystem.File.Exists(reportPath))
         {
@@ -45,7 +44,7 @@ public class DiskBaselineProvider : IBaselineProvider
     public async Task Save(IJsonReport report, string version)
     {
         var reportDirectory = FilePathUtils.NormalizePathSeparators(
-            Path.Combine(_options.ProjectPath, _outputPath, version));
+            Path.Combine(_options.ProjectPath, _options.BaselineOutputPath, version));
 
         _fileSystem.Directory.CreateDirectory(reportDirectory);
 
--- a/src/Stryker.Core/Stryker.Core/Baseline/Providers/DiskBaselineProvider.cs
+++ b/src/Stryker.Core/Stryker.Core/Baseline/Providers/DiskBaselineProvider.cs
@@ -27,8 +27,9 @@ public class DiskBaselineProvider : IBaselineProvider
 
     public async Task<IJsonReport> Load(string version)
     {
+        // at this point, BaseLineOutputPath is a fully qualified path
         var reportPath = FilePathUtils.NormalizePathSeparators(
-            Path.Combine(_options.ProjectPath, _options.BaselineOutputPath, version, "stryker-report.json"));
+            Path.Combine(_options.BaselineOutputPath, version, "stryker-report.json"));
 
         if (_fileSystem.File.Exists(reportPath))
         {
@@ -43,8 +44,9 @@ public class DiskBaselineProvider : IBaselineProvider
 
     public async Task Save(IJsonReport report, string version)
     {
+        // at this point, BaseLineOutputPath is a fully qualified path
         var reportDirectory = FilePathUtils.NormalizePathSeparators(
-            Path.Combine(_options.ProjectPath, _options.BaselineOutputPath, version));
+            Path.Combine(_options.BaselineOutputPath, version));
 
         _fileSystem.Directory.CreateDirectory(reportDirectory);
 
--- a/src/Stryker.Abstractions/Options/IStrykerOptions.cs
+++ b/src/Stryker.Abstractions/Options/IStrykerOptions.cs
@@ -15,6 +15,7 @@ public interface IStrykerOptions
     string S3Endpoint { get; init; }
     string S3Region { get; init; }
     BaselineProvider BaselineProvider { get; init; }
+    string BaselineOutputPath { get; init; }
     bool BreakOnInitialTestFailure { get; set; }
     int Concurrency { get; init; }
     string Configuration { get; init; }
--- a/src/Stryker.CLI/Stryker.CLI.UnitTest/Logging/InputBuilderTests.cs
+++ b/src/Stryker.CLI/Stryker.CLI.UnitTest/Logging/InputBuilderTests.cs
@@ -27,7 +27,8 @@ public class InputBuilderTests
         var gitIgnoreFile =
             fileSystemMock.AllFiles.Single(x => x.EndsWith(Path.Combine(".gitignore")));
         gitIgnoreFile.ShouldNotBeNull();
-        DateTime.TryParse(Directory.GetParent(gitIgnoreFile)!.Name.Split(".")[0], out _).ShouldBeTrue();
+        // the gitignore lives at the stable output root, not the per-run timestamped folder
+        Directory.GetParent(gitIgnoreFile)!.Name.ShouldBe("StrykerOutput");
         var fileContents = fileSystemMock.GetFile(gitIgnoreFile).Contents;
         Encoding.Default.GetString(fileContents).ShouldBe("*");
     }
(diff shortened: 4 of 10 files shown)


rouke-broersma:
@mpownby I don't think a new option is warranted here, instead stryker should follow the outputpath parameter for disk baseline provider, it is then auto-ignored because it will be covered by a gitignore.

mpownby:
## Reproduction

The disk baseline provider writes its report to `StrykerOutput` in the current working directory and ignores the `--output` parameter. These steps reproduce that, plus the downstream incremental-run breakage, using the **released** tool against a public sample repo.

### 1. Clone the repo and restore the Stryker tool

```bash
git clone https://github.com/RulecityLLC/DotNetMutationTestingSandbox.git
cd DotNetMutationTestingSandbox
dotnet tool restore
```

Reproduces on any recent 4.x and on current `main` — the path is hard-coded in `DiskBaselineProvider`.

### 2. Move into the test project

The documented baseline workflow runs from the test project directory, so this is the realistic case:

```bash
cd RestApi.Unit.Test
```

### 3. Seed a baseline, explicitly redirecting Stryker's output elsewhere

```bash
dotnet stryker --with-baseline:89b7eb28a64d1fa3f418faf99760e6d2a4bf63e5 --version main --output ../redirected-stryker-output
```

`89b7eb2…` is the repo's initial commit, so everything counts as "new" and a full baseline is written.

**Inspect the result — two things to notice:**

- `../redirected-stryker-output/` received the reports/logs **and** an auto-generated `.gitignore` containing `*`.
- But the baseline was written to **`RestApi.Unit.Test/StrykerOutput/baseline/main/stryker-report.json`** — i.e. `--output` was ignored for the baseline, which landed in the current directory instead.
- From the repo root, `git status` lists `RestApi.Unit.Test/StrykerOutput/` as **untracked**. Compare the two folders: the redirected output folder *has* an auto-`.gitignore`; the baseline folder does **not**, and `StrykerOutput` is not in the repo's `.gitignore` either. **The baseline is the one Stryker artifact that isn't auto-ignored.**

### 4. Run again incrementally — the baseline should be reused, but isn't

```bash
dotnet stryker --with-baseline:main --version main --output ../redirected-stryker-output
```

The log shows `Found baseline report for current branch main`, but then it re-tests **every** mutant with a message like *"Non-CSharp files in test project were changed"* — because the untracked baseline JSON inside the test project shows up in the git diff as a changed test file. The baseline is found yet provides no benefit, and there's no documented, supported way to relocate or suppress it.

## The problem and proposed solution

The disk baseline provider currently ignores `--output` and always writes `StrykerOutput` into the current working directory, and it's the **only** Stryker artifact not covered by the auto-generated `.gitignore` (which is only written into the timestamped report folder, never the baseline folder). So the baseline surfaces as an un-ignored file in the current working directory — which a) may surprise users as this behavior seems to be undocumented and b) breaks incremental runs, since it lands in the git diff and forces a full re-test.

`--baseline-output` lets users redirect it (the behavior they'd reach for `--output` to get) **without changing anything for existing users**. Two reasons a new opt-in option may make sense rather than making the baseline follow `--output`:

1. **The default output path is per-run timestamped.** When `--output` isn't supplied, the output path is `StrykerOutput/<timestamp>`, a new folder every run — so a baseline written there would never be found on the next run. "Just follow `--output`" can't work as-is.
2. **It may silently break existing users.** Anyone who already passes `--output` (for reports) would have their baseline relocate from `StrykerOutput` to the output directory, so the next run wouldn't find the existing baseline. A new opt-in option keeps current behavior untouched and lets users move the baseline only when they ask to.

mpownby:
> @mpownby I don't think a new option is warranted here, instead stryker should follow the outputpath parameter for disk baseline provider, it is then auto-ignored because it will be covered by a gitignore.

My first pass at this PR was to have disk baseline provider following --output if the user explicitly set it (or ignore it otherwise) but I was worried about breaking any existing users.  If you'd rather me go back to this solution, I can do that.

rouke-broersma:
Baseline and custom output directory are currently functionally at odds as you justifiably reported. Since we've never had this report before, it's unlikely it is actually being used together much in practice. I would class this as an oversight/bug when custom output directory was introduced. 

The two options are simply incompatible right now, and this must be addressed. It would also be acceptable to me if disk baseline provider places it's own .gitignore in the baseline directory instead of following the output directory, however I expect that users expect that output directory affects all files generated by stryker anyway.

Beside that we could also additionally support a new config parameter to explicitly set a baseline output path, that is also a valid ask.

mpownby:
> Baseline and custom output directory are currently functionally at odds as you justifiably reported. Since we've never had this report before, it's unlikely it is actually being used together much in practice. I would class this as an oversight/bug when custom output directory was introduced.
> 
> The two options are simply incompatible right now, and this must be addressed. It would also be acceptable to me if disk baseline provider places it's own .gitignore in the baseline directory instead of following the output directory, however I expect that users expect that output directory affects all files generated by stryker anyway.
> 
> Beside that we could also additionally support a new config parameter to explicitly set a baseline output path, that is also a valid ask.

If disk baseline follows the custom output directory, then another flag needs to be added to indicate whether the user set a custom output directory, because by the time the disk baseline runs, the output directory will be set to either an auto-generated one (with timestamp) or a user-supplied one, and it will fail if it writes to an auto-generated timestamp path.

I think if few are relying on this feature, then the 'right' solution is to make it follow the output path _if_ the user explicitly sets it.

If you agree, I can rework this PR to go down that path.

rouke-broersma:
I'm considering whether or not we should move full output path generation including timestamp back to stryker.core and should only pass the working directory or custom output directory from stryker.cli, then output directory can be used directly in baseline. I am not yet sure if that is worth it.

richardwerkman:
I'm also not in favor of adding a new option for this. Especially when setting the path in the settings file it could cause unexpected behavior when the relative path can not be resolved on another system. It just seems like a bug to me that this file is not outputted in the output directory. 

I suggest we place the file in the root of the StrykerOutput dir, not in a run specific sub-dir. Because the baseline feature needs to be able to find the file later. It would be impractical to search all sub-dirs for the latest baseline file.

rouke-broersma:
@richardwerkman yes but we don't have the 'root' available in core because it's computed in Stryker cli that's why I'm considering moving the computing back up to core

rouke-broersma:
@mpownby Thanks for your patience and allowing us to think about this

We talked it over and I'd still like to avoid a new parameter for this. I also don't think moving the output directory computation back to Core is the way to go. The CLI is the correct location. I consider it a bug that the disk baseline ignores --output and isn't covered by the gitignore. I propose keeping with my initial suggestion: make the disk baseline follow the output path, but make sure that location is gitignored.

As you said the default output path is per-run timestamped, so the baseline can't live there or it'd never be found next run. My suggestion is to introduce the notion of a stable "output root" with the StrykerOutput directory before the timestamp and place the gitignore here. Also compute the BaselineOutput based on the StrykerOutput directory in Stryker.CLI and pass it as a new input `BaselineOutputInput` to Stryker.Core. 

Basically almost identical to your current change except without the new param. This would technically be a breaking change as the baseline output is now covered by a gitignore and would follow --output but functionally probably what people want.

WDYT?

mpownby:
> @mpownby Thanks for your patience and allowing us to think about this
> 
> We talked it over and I'd still like to avoid a new parameter for this. I also don't think moving the output directory computation back to Core is the way to go. The CLI is the correct location. I consider it a bug that the disk baseline ignores --output and isn't covered by the gitignore. I propose keeping with my initial suggestion: make the disk baseline follow the output path, but make sure that location is gitignored.
> 
> As you said the default output path is per-run timestamped, so the baseline can't live there or it'd never be found next run. My suggestion is to introduce the notion of a stable "output root" with the StrykerOutput directory before the timestamp and place the gitignore here. Also compute the BaselineOutput based on the StrykerOutput directory in Stryker.CLI and pass it as a new input `BaselineOutputInput` to Stryker.Core.
> 
> Basically almost identical to your current change except without the new param. This would technically be a breaking change as the baseline output is now covered by a gitignore and would follow --output but functionally probably what people want.
> 
> WDYT?

(just got back from vacation, apologies for the delay) I'll give implementation a go.  Are you saying that you want the baseline to appear one level up from the timestamp directory (so the timestamp directory would be a child within the output directory instead of being the output directory) ?

UPDATE: I misspoke in my above comment and I think I understand what you mean now.  If --output is specified, the baseline folder will be put in the output directory, as a sibling of 'reports' and .gitignore .  And if it's not specified, a StrykerOutput directory will be created which will contain the .gitignore, the baesline folder and a timestamp folder which will contain the reports folder.

rouke-broersma:
> If --output is specified, the baseline folder will be put in the output directory, as a sibling of 'reports' and .gitignore . And if it's not specified, a StrykerOutput directory will be created which will contain the .gitignore, the baesline folder and a timestamp folder which will contain the reports folder.

Yes correct, so make Baseline follow the output param and move the gitignore up one level so it covers everything in output. Stryker.Core should no longer try to compute the BaselineDirectory, it should take this verbatim from the input as computed by Stryker.CLI. I think you were already most of the way there with the param implementation. 

I think it's fine to always place the gitignore even if output is specified by the user. If they user wants to include some files in the output directory into git they can modify the gitignore afterwards since we only place it if the file doesn't exist.

src/Stryker.Core/Stryker.Core/Baseline/Providers/DiskBaselineProvider.cs
```diff
@@ -16,7 +16,16 @@ public class DiskBaselineProvider : IBaselineProvider
...
     private readonly ILogger<DiskBaselineProvider> _logger;
-    private const string _outputPath = "StrykerOutput";
+    private const string _defaultOutputPath = "StrykerOutput";
+
+    /// <summary>
+    /// The directory (relative to the project path, or absolute) the baseline is stored under.
+    /// Honors the configured baseline output path, falling back to the stable StrykerOutput
+    /// folder so baselines persist across runs when no output path was supplied.
+    /// </summary>
+    private string OutputPath => string.IsNullOrWhiteSpace(_options.BaselineOutputPath)
```
rouke-broersma: We should always take the baseline outputpath, we no longer should compute this here

src/Stryker.Configuration/Options/Inputs/BaselineOutputInput.cs
```diff
@@ -0,0 +1,17 @@
+namespace Stryker.Configuration.Options.Inputs;
+
+public class BaselineOutputInput : Input<string>
+{
+    protected override string Description => "The directory the disk baseline provider stores and loads the baseline report from. This is derived from the output path (the stable output root) rather than supplied directly, so the baseline follows --output, is covered by the gitignore, and persists across runs since it lives outside the per-run timestamped folder.";
+
+    public override string Default => "StrykerOutput";
```
rouke-broersma: Should not have a default, input is mandatory

mpownby: I followed pattern of S3EndpointInput. is that acceptable?

src/Stryker.Core/Stryker.Core/Baseline/Providers/DiskBaselineProvider.cs
mpownby: ok I think I've fixed this according to your wishes

src/Stryker.Core/Stryker.Core/Baseline/Providers/DiskBaselineProvider.cs
```diff
@@ -29,7 +28,7 @@ public DiskBaselineProvider(IStrykerOptions options, IFileSystem fileSystem = nu
     public async Task<IJsonReport> Load(string version)
     {
         var reportPath = FilePathUtils.NormalizePathSeparators(
-            Path.Combine(_options.ProjectPath, _outputPath, version, "stryker-report.json"));
+            Path.Combine(_options.ProjectPath, _options.BaselineOutputPath, version, "stryker-report.json"));
```
rouke-broersma: Not sure if we should still path combine with the project path, I think BaselineOutputPath should always be a full path now. Maybe we should add a validation in the input validation to make sure it's a full path. Makes our lives easier in the rest of the code.

src/Stryker.Configuration/Options/Inputs/BaselineOutputInput.cs
rouke-broersma: yes looks good

src/Stryker.Core/Stryker.Core/Baseline/Providers/DiskBaselineProvider.cs
mpownby: ok I think I got this sorted.  I ran another manual test and it still seems to be working properly.

PR #3650, opened 2026-06-19
fix(mtp): Add StackOverflowException handling in MTP testrunner

richardwerkman:
fixes #3648 
relates to #3094
- The MTP runner now can handle mutations that cause a StackOverflowException. These mutants will be marked as `RuntimeError`.
- Fixes a performance issue where inter-process communication caused severer slowdown on large projects. Using `MemoryMappedFile` instead of a normal file fixes this issue while keeping results the same.

todo:

- [x] Integration tests
- [x] Update all reporters with new state
- [x] Fix performance issue

PR #3652, opened 2026-06-20
fix: Replace file read with a MemoryMapped file for performance win

richardwerkman:
Reduces testing time from 2 minutes to 20 seconds on project regex-parser.

Based on #3650 because I needed that fix to test this

image

todo:

- [ ] Investigate why there are more error results in this branch


richardwerkman:
I moved this into #3650 because it was merging together anyway..

PR #3669, opened 2026-06-26
fix(dashboard): do not delete the dashboard report

0xced:
For the past two years I have wondered why I would get "Report could not be found..." error on the dashboard when performing this command:

```sh
dotnet tool run dotnet-stryker --reporter dashboard --open-report:dashboard --version main --dashboard-api-key ******
````

Today I looked at the code and it turns out that the HTTP client performs a DELETE request on the report when all the mutants are tested, obviously deleting it!

Note: this regression was introduced in #2563.

### DIFF
--- a/src/Stryker.Core/Stryker.Core/Clients/DashboardClient.cs
+++ b/src/Stryker.Core/Stryker.Core/Clients/DashboardClient.cs
@@ -98,9 +98,6 @@ public class DashboardClient : IDashboardClient
                 batchResponse.EnsureSuccessStatusCode();
                 _batch.Clear();
             }
-
-            var deleteResponse = await _httpClient.DeleteAsync(url);
-            deleteResponse.EnsureSuccessStatusCode();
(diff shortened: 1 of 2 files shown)


rouke-broersma:
If I remember correctly the deletion is by design. It is a failsafe in case anything in the mutant matching, transport or whatever else (a risk was identified but I don't remember the specific reason that was brought forward) happens with real time reporting. However immediately after that the non-realtime report (so the full report as calculated by Stryker client) is supposed to be published iirc so it should not have any user impact so obviously something is going wrong there still.

0xced:
> However immediately after that the non-realtime report (so the full report as calculated by Stryker client) is supposed to be published.

The thing is, the current implementation publishes the report immediately _before_ and not immediately _after_ deletion. 😉

Addressed in 63ee722c182b144030af766c943311e905500e81 and restored the DELETE HTTP call.

PR #3674, opened 2026-06-29
feat: Add bail to MTP test runner

### LINKED ISSUE #3655: MTP stop testrun (bail) on first test failure
With vstest, we can stop a testrun when the first testcase fails. This saves runtime since we can mark the mutant killed without knowing the outcome of the other tests.

We should implement the same for MTP. It should be disabled when `--disable-bail` is passed, just like for vstest. That setting now does nothing with MTP.

**Additional context**
It might be as easy as setting a flag when starting the run. If not, it might be a bit complicated to implement. We'd have to listen to each testresult during the run and cancel manually. We don't have the setup for that yet.

richardwerkman:
closes #3655 and #3675

- Set `MinVerSkip` to true for improved compatibility.
- Send cancel RPC command when the first test failure result is received.

On first inspection this feature doesn't seem to do much on smaller projects. The tests are done before the cancel testrun command is received through RPC.

todo:

- [ ] Test on a larger project if the testrun actually cancels when the command is received, polly could be a good candidate


src/Stryker.Abstractions/Options/IStrykerOptions.cs
```diff
@@ -47,11 +47,7 @@ public interface IStrykerOptions
     string SinceTarget { get; init; }
     string SolutionPath { get; init; }
     string SourceProjectName { get; init; }
-    string TargetFramework { get; init; }
-    string TestCaseFilter { get; init; }
```
richardwerkman, line 51: Check if these options aren't used from the interface

src/Stryker.TestRunner.MicrosoftTestPlatform/MicrosoftTestingPlatformRunner.cs
```diff
@@ -158,7 +158,8 @@ private void WriteMutantIdToFile(int mutantId)
     {
         var envVars = new Dictionary<string, string?>
         {
-            ["STRYKER_MUTANT_FILE"] = _mutantFilePath
+            ["STRYKER_MUTANT_FILE"] = _mutantFilePath,
+            ["MinVerSkip"] = "true"
```
richardwerkman, line 162: This can be removed again

PR #3685, opened 2026-07-03
ci: Replace azure pipelines nightly integration tests with github workflows

rouke-broersma:


richardwerkman:
If I'm correct we now only use the azure pipeline for releasing. Which is fine, but now we don't have any verification in the release process that the package actually works. So I would prefer we move the release process to gh as well. Or do you think we can somehow verify the gh integration test succeeds when releasing in ado?

rouke-broersma:
Pull requests enforce running the integration tests on github, and since any code changes must go through a pull request all code changes are subject to integration tests.

PR #3694, opened 2026-07-07
fix(mutator): mutate implicit array creation expressions

### LINKED ISSUE #1024: Feature request: Mutate implicit array initializer

georgeglarson:
Following up on @dupdob's 2024 note that #1024 could restart once the SemanticModel plumbing landed (which it has). Opening as draft to get a read on the approach before pushing to merge.

`ArrayCreationMutator.ApplyMutations` switches on the node type and has no arm for `ImplicitArrayCreationExpressionSyntax`, so `new[]{1, 2}` is skipped entirely. `ShouldNotRemoveValuesFromImplicitArrayCreation` asserts the skip.

Adding an arm that resolves the element type via `semanticModel.GetTypeInfo(node).Type` (`IArrayTypeSymbol.ElementType`) and emits an explicit empty array gives the same mutant the explicit-array and stackalloc arms produce: `new[]{1, 2}` → `new int[]{ }`. Using `.Type` rather than `.ConvertedType`, since the conversion target (e.g. `IEnumerable<int>`) would emit mutants that don't compile.

One snag worth flagging. Building the replacement with the SyntaxFactory crashed Roslyn consistently: `IndexOutOfRangeException` in `Binder.ConvertAndBindArrayInitialization`, during target-type inference, in any position (argument, collection expression, local). Building it with `SyntaxFactory.ParseExpression` instead avoids it, and the mutant then applies everywhere. I couldn't shrink the crash to a standalone source repro (re-parsing the same source sidesteps it), so I haven't filed it upstream; happy to if it's useful.

Anonymous-type element arrays are skipped (`new[] { new { x = 1 } }`, nested in arrays or tuples too), since the element type can't be written explicitly.

Where it stands: new unit and orchestrator tests cover int, double, jagged, tuple, anonymous, and null-SemanticModel cases; two existing tests had fixtures shift and are updated; `Stryker.Core.UnitTest` is 1530/0. End-to-end on a two-file sample (`public static int[] FixedNumbers() => new[]{1, 2, 3};` plus a test asserting the value), Stryker produced 3 mutants, two of them `Array initializer mutation` on the implicit arrays, all killed.

Closes #1024.


richardwerkman:
Thanks for your contribution! Seems to be implemented correctly. It would be good to add this case to the integration test project. We have example code there which causes all kinds of mutations. Or maybe we already have an implicit array in [the constructs](https://github.com/stryker-mutator/stryker-net/tree/master/integrationtest/TargetProjects/NetCore/TargetProject/Constructs) (we'd have to check).

georgeglarson:
Good call. Implicit arrays were already in the Constructs (CSharp11 list patterns, CSharp12's `sequence`), so the new mutator was already firing on them. But none of those had a test asserting the values, so every mutant landed as NoCoverage.

Added a dedicated construct in CSharp3 (the C# 3.0 home for implicitly typed arrays) with a test that asserts the inferred elements, so the empty-array mutant is killed end-to-end. Refreshed the baselines for the scenarios that mutate TargetProject (Single, Multiple, Solution, and the MTP variants). NetFramework and WebApi targets have no implicit arrays, so they're untouched.

All ten netcore integration scenarios pass locally.

georgeglarson:
Added in the PR via commit 8460fb28 (`test(integration): cover implicit-array creation + refresh baselines`). The construct is at `TargetProject/Constructs/CSharp3.cs:12`:

```
public static int[] ImplicitArrayCreation() => new[] { 1, 2, 3 };
```

and `NetCoreTestProject.XUnit/Constructs.cs:22` asserts on it:

```
Assert.Equal(new[] { 1, 2, 3 }, CSharp3.ImplicitArrayCreation());
```

The empty-array mutant fails that assert, so it gets killed. A local run on the integration project confirms.

Happy to add a different case if you had one in mind, or anything else needed.

georgeglarson:
Reverting the 13 Jul master merge to get CI green again (back to the `bf8808da` tree). The mutator work was green at the prior merge and the red came in with the master merge, not with these changes. Filing a separate issue for the dep question.

georgeglarson:
Correction to my comment above: "the red came in with the master merge" isn't supported.

I retriggered on the reverted tree (no master-merge deps) and it still fails, on different platforms across runs (windows vs ubuntu on byte-identical source). So the red isn't tied to the master merge — I can't reproduce a stable result and don't know what's driving the variation.

See #3706. Apologies for the earlier framing.

Issue #3689, opened 2026-07-03
MTP per-test coverage analysis


richardwerkman:
We need `per-test` and `per-test-in-isolation` coverage analysis in MTP just like we have in VSTest. Since for MTP we can reuse the testhost between sessions, we can reuse the process between running each test. This might cause some issues with coverage analysis. That's why we should enable `per-test-in-isolation` by default for MTP.

hiSandog:
Per-test coverage with MTP should define how test identity is mapped back to mutants before implementation. Parameterized tests, dynamic display names, and retries can otherwise make coverage attribution unstable. A good first test fixture would include one normal test, one parameterized test, and one test that shares setup code, then assert the same mutants are selected across repeated MTP runs.

dupdob:
@hiSandog : thanks for the reminder, but please note that there is no standard way or unified logic regarding how this tests are identified; it ultimately depends on the test framework. Hense these tests would be valid for tested framework.

You can check [testing-framework.md](https://github.com/stryker-mutator/stryker-net/blob/master/docs/technical-reference/testing-framework.md) . It contains several findings regarding test framework (in the VsTest context, not sure it translates as is in MTP)

richardwerkman:
Fixed by #3752

Issue #3629, opened 2026-06-11
MTP runner: coverage analysis assigns every covered mutant the full test suite (no per-test coverage), so each mutant runs against all tests


fpatseas:
**Describe the bug**

With the Microsoft Testing Platform (MTP) runner, mutation testing runs every *covered* mutant against the **entire test suite** instead of just the tests that cover it. On a project with 3,296 tests, every covered mutant is logged as `will be tested against (3296) tests`. Per-mutant time is therefore dominated by running hundreds-to-thousands of irrelevant tests (and frequently hitting the per-run timeout, which forces a test-server restart), rather than the ~handful of tests that actually cover the mutant.

Root cause: the MTP coverage capture produces only **aggregate** coverage (which mutants were hit during the whole run) with **no per-test attribution**, and then explicitly tells the analyser that *every test covers every covered mutant*.

Step by step (verified identical in 4.14.2 and `master`):

1. **Coverage mode accumulates a single global covered-mutant set.** The injected `MutantControl.RegisterCoverage(id)` adds the id to a static `_coveredMutants` list shared across *all* tests; `FlushCoverageToFile()` writes that list once on process exit. There is no per-test boundary/reset. `SingleMicrosoftTestPlatformRunner.ReadCoverageData()` reads that one aggregate file.

2. **`MicrosoftTestPlatformRunnerPool.CaptureCoverage()` then claims every test covers every covered mutant**:

   ```csharp
   // For cumulative coverage, we return a single coverage result that applies to all tests
   // Each test is assumed to cover all the mutations that were covered during the full test run
   return _testDescriptions.Values.Select(testDescription =>
       CoverageRunResult.Create(testDescription.Id, CoverageConfidence.Normal,
           allCoveredMutants,   // every test's result lists ALL covered mutants
           allStaticMutants, []));
   ```

3. **So `CoverageAnalyser` gives every covered mutant the whole suite.** In `CoverageForThisMutant`, `mutationToResultMap[mutantId]` then contains every test, so `testGuids` = all tests and `mutant.CoveringTests` = the full suite -> the `will be tested against (3296) tests` log line.

[...]

PR #3696, opened 2026-07-08
fix(mtp): write one coverage file per test assembly host to prevent nondeterministic coverage loss

npnelson:
# fix(mtp): write one coverage file per test assembly host to prevent nondeterministic coverage loss

### Summary

MTP solution runs start one test host per test assembly, but all hosts currently receive the same `STRYKER_COVERAGE_FILE`. The injected `MutantControl` flushes coverage with `File.WriteAllText` on process exit, so each host replaces the previous contents instead of merging with them. In a multi-assembly run, the final successful flush determines which assembly's coverage survives.

This PR assigns each assembly host its own coverage file and unions those files when reading coverage. `MutantControl` is unchanged.

### Symptom

The `integration-test (windows-latest, MTPSolution, netcore)` job failed intermittently with:

```text
actualSurvived   should be 1  but was 2
actualTimeout    should be 0  but was 2
actualNoCoverage should be 360 but was 357
```

Example: Integration Tests run `28915858912`, job `85782638342`, on #3695. That PR changed MTP runner isolation behavior, but not the coverage-file plumbing.

### Root Cause

The three moved mutants are the relational mutants of `while (1 < 0)` in `StrykerFeatures/Timeout.cs`:

- `<` -> `<=` still never enters the loop, so it survives.
- `<` -> `>` and `<` -> `>=` loop forever, so they time out.

Of the four MTP sample projects, only `UnitTests.MSTest` covers `Timeout.cs`; the XUnit, NUnit, and TUnit samples do not touch it. Since every sample covers the same `KilledMutants.IsExpiredBool` mutants, the MSTest host's covered set is a superset of the others.

Observed fingerprints from `Coverage capture complete: N mutations covered`:

- Clobbered outcome: `3 covered`, `NoCoverage: 360`, matching the checked-in baseline.
- Failing Windows job: `14 covered`, `NoCoverage: 357`, with `Timeout.cs` tested.
- With this fix: `14 covered` on successful coverage captures, independent of flush order.

So the old `MTPSolution` baseline was green because coverage from one or more hosts was being lost. Single-project MTP categories have one host and nothing to overwrite, which is why `MSTestMTP` already expects `timeout: 2` from these same mutants.

### Fix

- `SingleMicrosoftTestPlatformRunner` assigns a coverage file per assembly:
  `stryker-coverage-{processId}-{runnerId}-{nonce8}-{assemblyName<=32}-{pathHash8}.txt`
- The path is stable per assembly for the runner's lifetime.
- `ReadCoverageData()` unions covered/static ids across all assigned per-assembly files.
- Missing files are skipped, so one crashed host does not erase coverage from the others.
- `SetCoverageMode` and `Dispose` clean up all assigned coverage files.

### Reviewer Notes

1. Baseline update vs ignore rule

   The second commit changes `CSharp_NetCore_MTPSolution` to:

   ```text
   survived: 2
   killed: 1
   timeout: 2
   nocoverage: 357
   ```

   These are the values produced when the MSTest coverage survives, and they are consistent with `MSTestMTP` already expecting `timeout: 2` for the same `Timeout.cs` mutants. The trade-off is two timeout mutants in the MTPSolution job. If maintainers prefer keeping that integration job faster, an ignore rule for `Timeout.cs` in the solution config would also work, with `nocoverage` adjusted accordingly.

2. Process id + nonce in the file name

   Coverage files are only deleted after their paths have been assigned. With predictable names, a run could read a stale file left by a crashed earlier run, and concurrent Stryker processes could share names. The process id handles ordinary cross-process cases; the nonce covers pid reuse and same-id runner instances in one process.

   Hard-crash leftovers may remain in temp instead of being overwritten by a later run, but these files are tiny and the existing mutant-id file already behaves similarly on crash.

3. Mutant-id file intentionally unchanged

   `stryker-mutant-{runnerId}.txt` has an analogous concurrent-process collision risk, but it is a live control channel with different failure modes and existing tests. I left it unchanged to keep this PR focused on coverage loss.

4. No merge-on-write in `MutantControl`

   An alternative fix would be read-union-write under a cross-process lock in the injected helper. That would add locking/retry logic to C# 2-compatible code that runs inside every user's test process. Once each coverage file has one owning host, merge-on-write is unnecessary.

### Verification

- `Stryker.TestRunner.MicrosoftTestPlatform.UnitTest` passes, including:
  - `ReadCoverageData_ShouldUnionCoverageAcrossAssemblies`
  - path stability/uniqueness tests
  - name truncation test
  - same-id runner nonce separation test
- Local before/after on `MTPSolution` with packed CLI and `--solution MicrosoftTestPlatform.slnx --test-runner mtp`:
  - pre-fix: `3 covered`, `NoCoverage: 360`
  - post-fix: `14 covered`, `NoCoverage: 357`
- `MSTestMTP` passes against its unchanged baseline, confirming single-assembly behavior is unaffected.

### Related

- #3689 / #3629: the cumulative "every test covers every covered mutant" model is unchanged. This PR only makes the cumulative set complete and deterministic.
- #3516: per-test coverage via host restarts reuses the same `ProcessExit` flush. Per-assembly coverage files compose with it, though both PRs touch `SingleMicrosoftTestPlatformRunner.cs`, so whichever lands second will need a small rebase.
- Explains the Windows CI flake observed on #3695.

### DIFF
--- a/integrationtest/Validation/ValidationProject/ValidateStrykerResults.cs
+++ b/integrationtest/Validation/ValidationProject/ValidateStrykerResults.cs
@@ -197,7 +197,11 @@ public class ValidateStrykerResults
 
         var report = await strykerRunOutput.DeserializeJsonReportAsync();
 
-        CheckReportMutants(report, total: 670, ignored: 274, survived: 1, killed: 1, timeout: 0, nocoverage: 360);
+        // Coverage is the union over all four test projects, so the Timeout.cs mutants covered only
+        // by the MSTest project count as covered (1 survived + 2 timeout), like in the MSTestMTP run.
+        // Before coverage files were split per test host, the final flush overwrote the shared
+        // file, usually losing exactly those three mutants to NoCoverage.
+        CheckReportMutants(report, total: 670, ignored: 274, survived: 2, killed: 1, timeout: 2, nocoverage: 357);
         CheckReportTestCounts(report, total: 0); // MTP doesn't report tests yet
     }
 
--- a/src/Stryker.TestRunner.MicrosoftTestPlatform.UnitTest/SingleMicrosoftTestPlatformRunnerCoverageTests.cs
+++ b/src/Stryker.TestRunner.MicrosoftTestPlatform.UnitTest/SingleMicrosoftTestPlatformRunnerCoverageTests.cs
@@ -22,29 +22,32 @@ public class SingleMicrosoftTestPlatformRunnerCoverageTests
         _discoveryLock = new object();
     }
 
+    private SingleMicrosoftTestPlatformRunner CreateRunner(int runnerId) =>
+        new(runnerId,
+            _testsByAssembly,
+            _testDescriptions,
+            _testSet,
+            _discoveryLock,
+            NullLogger.Instance);
+
     [TestMethod]
     public async Task SetCoverageMode_ShouldEnableCoverageMode()
     {
         var runnerId = 600;
-        var coverageFilePath = Path.Combine(Path.GetTempPath(), $"stryker-coverage-{runnerId}.txt");
-        
+        string? coverageFilePath = null;
+
         try
         {
-            // Create an existing coverage file that should be deleted
-            await File.WriteAllTextAsync(coverageFilePath, "1,2,3");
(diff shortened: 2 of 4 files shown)


richardwerkman:
Looks good to me, this can be merged if the integration tests are stable with a few reruns

PR #3724, opened 2026-07-18
fix(mtp): isolate mutant-id control files per runner instance

npnelson:
Runner IDs are local to each MTP pool, but mutant-id control files were named using only that ID. Concurrent runs could therefore share a control file, change which mutant another run’s test hosts execute, and interfere with cleanup.

Use the runner’s existing process- and instance-qualified identity for the mutant-id filename, matching the coverage-file naming scheme.

Regression tests reproduce the collision with the old filename and verify that another runner’s construction, writes, and disposal leave the original runner’s live memory-mapped control file intact and responsive to subsequent updates.

Follow-up to #3696.


richardwerkman:
@npnelson Thanks for this submission! Looks good, but we were a bit slow to review so now there are some merge conflicts. I'm sorry about that. Would you be able to resolve them? I'm good with merging after that.

npnelson:
@richardwerkman Resolved the merge conflicts! Thanks for all of your hard work.

Issue #3692, opened 2026-07-06
Microsoft Testing Platform (MTP) hardcoded 3min timeout kills discovery


gurufordy:
**Describe the bug**

When using the **Microsoft Testing Platform (MTP) runner** (`--test-runner mtp`) on a test project whose full suite takes **longer than 3 minutes to run**, the *initial* test run is always cancelled and Stryker aborts with:

```
No test result reported. Make sure your test project contains test and is compatible with VsTest.
```

The root cause is a **hardcoded 3-minute timeout** on the `testing/runTests` JSON-RPC call in the MTP client. In `Stryker.TestRunner.MicrosoftTestPlatform/TestingPlatformClient.cs`, `RunTestsAsync` wraps the RPC invocation in:

```csharp
using CancellationTokenSource cancellationTokenSource = new(TimeSpan.FromMinutes(3));
...
await JsonRpcClient.InvokeWithParameterObjectAsync(
    "testing/runTests",
    new RunTestsRequest(RunId: requestId, TestCases: testNodes),
    cancellationToken: cancellationTokenSource.Token);
```

For the **initial test run** there is no baseline timing yet, so `SingleMicrosoftTestPlatformRunner`/`AssemblyTestServer` invoke `RunTestsAsync` with `timeout: null` (i.e. Stryker intends *no* per-run timeout). But the client's own hardcoded 3-minute `CancellationTokenSource` still fires, cancelling the run at exactly ~180s regardless. Stryker retries once (attempt 2/2), hits the same 3-minute cancel, marks the test host as crashed, and then reports "No test result reported".

This makes the MTP runner unusable for any project whose initial full-suite run exceeds 3 minutes. The 3-minute value is not configurable — `additional-timeout` only affects per-mutant timeouts (`timeout = initialTestTime + additionalTimeout`), not this initial-run RPC cap.

Smaller suites (well under 3 minutes) run fine; the failure is purely a function of total initial-run duration, not of the tests themselves (the same suite passes under a normal `dotnet test` / MTP run).

**Logs**

Run with `-L` / `--log-to-file` (xUnit v3 project on MTP, ~2,400 tests, full suite > 3 min). Key excerpt:

```
[DBG] Initial build successful
[WRN] The Microsoft Test Platform testrunner is currently in preview. Results should be verified since this feature is still being tested.
[DBG] MtpRunner-1: Discovered 2408 tests in ...\MyApp.Tests.dll
[...]

PR #3731, opened 2026-07-19
feat(timeouts): Calculate timeout values based on actual mutant runtimes

### LINKED ISSUE #3653: MTP calculate timeout per test
In the mtp testrunner we use the full testrun as the base for the timeout calculation.

We should capture the duration per testcase and calculate a timeout value for each mutant.

richardwerkman:
closes #3653

Changes:

- Refine VsTest + MTP timeout estimation to use the union of assessing tests for the mutant group being executed.
- Add TimeoutRatio/timeout-ratio configuration support and thread it into TimeoutValueCalculator.
- Update docs with new behaviour
- Fix bug where test discovery was cancelled after 3 minutes #3692


docs/configuration.md
```diff
@@ -301,19 +301,30 @@ If HTML and/or JSON reporting is being used you can use this option to change th
...
 
 *\* Timeout is in milliseconds.*
 
+### `timeout-ratio` &lt;`number`&gt;
+
+Default: `1.5`  
+Command line: `N/A`  
+Config file: `"timeout-ratio": 2.0`
+
+The ratio the estimated test time is multiplied by when calculating the [timeout](#additional-timeout-number) for a mutant. A higher ratio gives mutants more time to run before they are cancelled, a lower ratio catches endless loops faster. Must be higher than `0`.
```
richardwerkman:
```suggestion
The ratio the estimated test time is multiplied by when calculating the [timeout](#additional-timeout-number) for a mutant. A higher ratio gives mutants more time to run before they are cancelled, a lower ratio catches endless loops faster. Must be higher than `1`.
```

docs/configuration.md
```diff
@@ -301,19 +301,30 @@ If HTML and/or JSON reporting is being used you can use this option to change th
 
 ### `additional-timeout` &lt;`number`&gt;
 
-Default: `5000`
```
richardwerkman, line 304: I'm not sure if we should lower this. Timeout values will already be lower because of the actual mutants and this is also significantly lower. It could increase the chance for false timeouts

richardwerkman:
I have tested multiple times on a large project with per mutant timeout, and sadly the number of timeout results increased a lot. 24 timeouts on master, 54 in this branch. Note that this was on a project with a testrun duration of 30 seconds. So the timeout value on master is 50 seconds. On this branch with the old defaults it is considerably lower at around 5 seconds. Only after I increased the additional-timeout value to 10s I started to see similar results. So for now I'm increasing the defaults to 10 seconds and a multiple of 2. This will increase the timeout values a bit on small projects, but on larger projects like stryker it will still be lower, while keeping results the same.

@rouke-broersma agreed?

rouke-broersma:
Sure

PR #3736, opened 2026-07-22
fix(core): lock reportedMutants dedup to fix HashSet race under concurrency

### LINKED ISSUE #3727: Crash: HashSet corruption in MutationTestProcess.OnMutantTested under --concurrency > 1 (data race on reportedMutants)
## Summary

Running mutation testing at the default concurrency intermittently aborts the whole run with an
unhandled exception thrown from `HashSet<IMutant>.Add` inside `MutationTestProcess.OnMutantTested`.
The `reportedMutants` set is a plain, non-thread-safe `HashSet<IMutant>` that is written from two
code paths which overlap when concurrency > 1. Running with `--concurrency 1` avoids it reliably.

## Environment

- Stryker.NET **4.16.0** (dotnet global tool)
- .NET SDK 10.0.x, Windows
- Test project: standard xUnit/VSTest-based unit test project (~2,900 tests), ~100 mutants in scope
- Default concurrency (multi-core host)

## Steps to reproduce

1. Run `dotnet stryker` on a project with a non-trivial number of mutants and tests at default concurrency.
2. Intermittently, after the baseline run, during mutant testing the process aborts with the stack trace below.
3. Re-running with `--concurrency 1` completes normally every time.

(The race is timing-dependent and thread-pressure-sensitive, so it reproduces more readily on many-core
hosts and larger mutant sets; `--concurrency 1` is a reliable negative control.)

## Stack trace (abridged)

```
System.Collections.Generic.HashSet`1.Add(T item)
  at Stryker.Core.MutationTest.MutationTestProcess.OnMutantTested(IMutant mutant, ISet`1 reportedMutants)  MutationTestProcess.cs:153
  at Stryker.Core.MutationTest.MutationTestProcess.OnMutantsTested(IEnumerable`1 mutants, ISet`1 reportedMutants)  MutationTestProcess.cs:140
  at Stryker.Core.MutationTest.MutationTestProcess.<TestMutantsAsync>b__17_0(List`1 mutants, CancellationToken)  MutationTestProcess.cs:99
  at System.Threading.Tasks.Parallel.<>c__53`1.<<ForEachAsync>b__53_0>d.MoveNext()
  ...
  at Stryker.Core.MutationTest.MutationTestProcess.TestMutantsAsync(...)  MutationTestProcess.cs:90
  at Stryker.Core.StrykerRunner.RunMutationTestAsync(...)  StrykerRunner.cs:112
```

## Root cause

`reportedMutants` is created per mutant-group inside the `Parallel.ForEachAsync`
(`MutationTestProcess.TestMutantsAsync`), and is a non-thread-safe `HashSet<IMutant>`:

```csharp
// MutationTestProcess.TestMutantsAsync
await Parallel.ForEachAsync(mutantGroups, new ParallelOptions { MaxDegreeOfParallelism = _options.Concurrency },
    async (mutants, cancellationToken) =>
{
    var reportedMutants = new HashSet<IMutant>();            // (A) plain HashSet, no synchronization
    await _mutationTestExecutor.TestAsync(..., (t, f, r, o) =>
        TestUpdateHandler(t, f, r, o, reportedMutants));      // (B) writes reportedMutants via the test-progress callback
    OnMutantsTested(mutants, reportedMutants);                // (C) writes reportedMutants again, AFTER the await
});
```

Both `TestUpdateHandler` and `OnMutantsTested` call `OnMutantTested`, which does an unsynchronized
`Contains` + `Add`:

```csharp
private void OnMutantTested(IMutant mutant, ISet<IMutant> reportedMutants)
{
    if (mutant.ResultStatus == MutantStatus.Pending || reportedMutants.Contains(mutant))
        return;
    _reporter?.OnMutantTested(mutant);
    reportedMutants.Add(mutant);        // <-- throws here, corrupted HashSet
}
```

The callback in (B) is delivered by the VSTest runner from **its own IPC/background threads**. In
`VsTestRunner.RunTestSession`:

```csharp
runEventHandler.ResultsUpdated += HandlerUpdate;
foreach (var source in ...) { RunVsTest(..., runEventHandler); }   // session runs inside Task.Run(...)
runEventHandler.ResultsUpdated -= HandlerUpdate;                    // unsubscribe
return (runEventHandler.GetResults(), ...);                        // returns WITHOUT joining in-flight handler invocations
void HandlerUpdate(object s, EventArgs e) => updateHandler?.Invoke(runEventHandler.GetResults());
```

`RunTestSession` returns as soon as the session task completes, but does not guarantee that all
in-flight `ResultsUpdated` → `HandlerUpdate` → `updateHandler.Invoke` → `OnMutantTested` invocations
have finished. So a late `ResultsUpdated` handler (running on a VSTest background thread, writing
`reportedMutants`) can still be executing when control returns from `await TestAsync` and step **(C)**
`OnMutantsTested` writes the same `HashSet` from the `Parallel.ForEachAsync` worker thread — a data
race on a non-thread-safe collection, which corrupts it and makes `Add`/`Contains` throw.

`--concurrency 1` removes the thread pressure (runner pool size is `Math.Max(1, Concurrency)` and the
outer `ForEachAsync` degree is 1), so the window effectively closes — consistent with a race.

This is the same class as the previously-fixed #2899 ("Concurrency issue with hashset on MutantPlacer"),
in a different collection.

## Proposed fix

Synchronize access to `reportedMutants` in `MutationTestProcess.OnMutantTested`. Since the set is
per-group, the lock is essentially uncontended:

```csharp
private void OnMutantTested(IMutant mutant, ISet<IMutant> reportedMutants)
{
    if (mutant.ResultStatus == MutantStatus.Pending) return;
    lock (reportedMutants)
    {
        if (!reportedMutants.Add(mutant)) return;   // atomic dedup: Add returns false if already present
    }
    _reporter?.OnMutantTested(mutant);              // report once, outside the lock
}
```

(Optionally, as defense in depth, `VsTestRunner.RunTestSession` could await all pending
`ResultsUpdated` handler invocations before returning — but the lock is the tight, runner-agnostic fix.)

I'm happy to open a PR with the fix plus a concurrency stress regression test on `OnMutantTested`.

## Workaround

Run with `--concurrency 1`.

anagnorisis2peripeteia:
Mutation runs at the default concurrency occasionally aborted the whole run with an unhandled `HashSet<IMutant>.Add` exception thrown from `MutationTestProcess.OnMutantTested`; running with `--concurrency 1` reliably avoided it. Fixes #3727.

**The cause:** the per-mutant-group `reportedMutants` set is written from two threads — the VsTest progress callback (raised by the runner on its own notification thread) and `OnMutantsTested`, which runs on the `Parallel.ForEachAsync` worker after `await TestAsync`. On the timeout/abort path the runner can return to `OnMutantsTested` while a final progress callback is still in flight (`VsTestRunner.WaitForEnd` stops waiting before the session completes), so the two overlap and a concurrent `Contains`/`Add` corrupts the non-thread-safe `HashSet`. So concurrency isn't the root cause — it just makes the overlap likely enough to hit, and `--concurrency 1` is a mitigation rather than a fix.

**The change:** guard the dedup `Contains`/`Add` in `OnMutantTested` with a lock on the (per-group, uncontended) set, and report each mutant exactly once via `Add`'s bool result. The reporter notification stays outside the lock, so reporting — which can block, e.g. real-time Dashboard publishing — never gates an unrelated worker.

**Testing:** adds a regression test that deterministically overlaps the two writers — it parks the progress callback inside the reporter (after the `Contains` check, before it would `Add`) while `OnMutantsTested` runs for the same mutant — and asserts the mutant is reported exactly once. It reports twice without the lock and once with it. The existing unit and integration suites pass.

**Evidence:** running that test against this exact change exercises the real `OnMutantTested` path end to end: with the lock it observes one report per mutant, and with the lock reverted to `main` it observes two (the second writer passing the `Contains` check while the first is parked before its `Add`) — deterministically, in ~90 ms. The unit test project and the VsTest integration run both pass on this commit.

---
Review history: https://gist.github.com/a1bb87d22c189c68cc7135b7a7a2229d

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/MutationTest/MutationTestProcessTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/MutationTest/MutationTestProcessTests.cs
@@ -1,8 +1,10 @@
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.IO.Abstractions.TestingHelpers;
 using System.Linq;
 using System.Reflection;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
@@ -10,6 +12,8 @@ using Moq;
 using Shouldly;
 using Stryker.Abstractions;
 using Stryker.Abstractions.Exceptions;
+using Stryker.Abstractions.Testing;
+using Stryker.TestRunner.Tests;
 using Stryker.Abstractions.Options;
 using Stryker.Abstractions.Reporting;
 using Stryker.Configuration.Options;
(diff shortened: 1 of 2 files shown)


anagnorisis2peripeteia:
Closing this for now.

Straight account of where it landed: we did hit this in a real run — Stryker 4.16 aborted at default concurrency on a ~2,900-test project and completed once we dropped to `--concurrency 1` — and the fix guards a `reportedMutants` `HashSet` that genuinely is written from two paths without synchronization. But the gaps matter:

- I couldn't reproduce the crash on demand — ran the original scenario ~95 times, it never fired.
- That target never produces a test **timeout**, so the VsTest abort path I proposed as the trigger (a late `ResultsUpdated` callback racing `OnMutantsTested`) is never exercised there. So I can't demonstrate the mechanism, only that the set is unsynchronized.

This is at best a defensive fix for a real-but-unreproduced race, and your skepticism about the analysis was warranted. Rather than ask you to weigh a change I can't back with a reproduction, I'll close it — and reopen with an actual repro if a reproducible case ever surfaces. Thanks for the careful review.

PR #3750, opened 2026-07-30
fix(MTP): send the runTests filter under the property the server binds

heyadamyoung:
## Problem

The `testing/runTests` request serializes its test filter as `testCases`, but Microsoft.Testing.Platform's server binds the run filter from the `tests` property (and `uid` / `display-name` on each entry). The server treats the unbound property as the absence of a filter and runs the **complete assembly** for every run request.

For mutation runs this is severe:

- every session pays the full-suite execution cost, regardless of coverage analysis — filtered runs are silently unfiltered;
- tests that are not part of the batch execute against whichever mutant is currently active, which corrupts verdicts in both directions for any activation scheme that assumes only the requested tests run.

## Reproduction / verification

Verified against a live Microsoft.Testing.Platform server (shipped with Microsoft.NET.Test.Sdk 18.7, xunit.v3 test assembly, .NET 10):

- **Before**: a run request filtered to a single test executed all ~1,700 tests in the assembly (observed via the update stream: full suite reported in-progress/passed).
- **After**: the same request runs exactly the requested test.
- Binding `displayName` instead of `display-name` is rejected by the shipped server's binder (`KeyNotFoundException: Key 'display-name' was not found`), so the entry shape matches the shipped server rather than the camelCase used elsewhere in the protocol.

## Change

- `RunTestsRequest` serializes the filter as `tests`, with a dedicated `RunRequestTestNode(uid, display-name)` entry shape (the response-side `TestNode` keeps its existing shape).
- `TestingPlatformClient.RunTestsAsync` maps the requested nodes into that shape.
- New serialization regression test pinning the wire property names; the existing fake-server round-trip test now exercises the corrected `tests` binding end-to-end.

All 201 tests in `Stryker.TestRunner.MicrosoftTestPlatform.UnitTest` pass.

### DIFF
--- a/src/Stryker.TestRunner.MicrosoftTestPlatform.UnitTest/RunTestsRequestTests.cs
+++ b/src/Stryker.TestRunner.MicrosoftTestPlatform.UnitTest/RunTestsRequestTests.cs
@@ -16,13 +16,17 @@ public class RunTestsRequestTests
         // serialized under any other property name is silently ignored and the
         // server runs the complete assembly, so mutation runs execute tests
         // that have no mutant assignment.
+
+        // Arrange
         var runId = Guid.Parse("11111111-2222-3333-4444-555555555555");
         var request = new RunTestsRequest(
             runId,
             [new RunRequestTestNode("case-uid", "Example.SampleTests.Case")]);
 
+        // Act
         var serialized = JsonSerializer.Serialize(request, RpcJsonSerializerOptions.Default);
 
+        // Assert
         using var document = JsonDocument.Parse(serialized);
         var root = document.RootElement;
         root.GetProperty("runId").GetString().ShouldBe(runId.ToString());
--- a/src/Stryker.TestRunner.MicrosoftTestPlatform.UnitTest/RunTestsRequestTests.cs
+++ b/src/Stryker.TestRunner.MicrosoftTestPlatform.UnitTest/RunTestsRequestTests.cs
@@ -1,0 +1,35 @@
+using System.Text.Json;
+using Shouldly;
+using Stryker.TestRunner.MicrosoftTestPlatform.Models;
+using Stryker.TestRunner.MicrosoftTestPlatform.RPC;
+
+namespace Stryker.TestRunner.MicrosoftTestPlatform.UnitTest;
+
+[TestClass]
+public class RunTestsRequestTests
+{
+    [TestMethod]
+    public void RunTestsRequest_ShouldSerializeFilterUnderTheServersPropertyNames()
+    {
+        // Microsoft.Testing.Platform's server binds the run filter from the
+        // "tests" property and each entry's "uid" and "display-name". A filter
+        // serialized under any other property name is silently ignored and the
+        // server runs the complete assembly, so mutation runs execute tests
+        // that have no mutant assignment.
+        var runId = Guid.Parse("11111111-2222-3333-4444-555555555555");
+        var request = new RunTestsRequest(
+            runId,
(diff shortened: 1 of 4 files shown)


heyadamyoung:
Withdrawing this for now.

rouke-broersma:
@Polis59 could you clarify why you're withdrawing this PR? We now can't determine if this is something to be determined at a later stage or if it was an incorrect problem analysis or solution.

richardwerkman:
I've noticed this issue as well. I've already fixed this in branch `feature/per-test-coverage-mtp`.

dupdob:
@richardwerkman : quick question: my understanding that this is not a bug per se today, as the MTP Runner does not have yet detailed test coverage, that is until your PR is merged. you confirm?

For context: I suspect we are starting to see bug report which are based on AI code review/bug hunt more than real world discovery and I had a glance at the code and failed to see a code path were the parameter was used..

rouke-broersma:
@dupdob it is technically a bug in that we have code in the testrunner harness that could be used to pass a test filter, however this code is unused until the coverage feature is implemented which is probably why the mistake in protocol was not discovered before and was caught immediately when working on that feature.

I suspect that you are correct that these are ai findings but I'm not entirely sure these are only bug hunting expeditions. I for the most part still think these are legitimate users hitting an issue, but then attempting to solve the problem using agentic tools.

richardwerkman:
I concur. It should not be an issue for anyone now, except when using the test-case-filter option, that won't work for now.

dupdob:
@rouke-broersma : yep, probably.  My overall point is that ultimately, this is not helping anyone. There is an increase of _finger pointing_ issues (there is a but there) vs _I have an issue_.  Or in orphan PR's.

I must confess I am quite anti AI. To be more specific, I am dead against: "I gave an objective to my agent, the rest is not my problem." 

That is why I attempt to tag AI sourced issues and PRs: to see what is the trend. But my tagging may be biased: I am more likely to tag a bad issue or PR as AI than a correct one

dupdob:
Circling back to this: this is not a bug. A bug refers to some incorrect behavior or result. At worst, it is a dormant bug. 
This has consequences: it is not reproducible. 
We know Stryker is a complex endeavor entangled in dependencies we have no control over: from Buildalyzer to Roslyn compiler, from test frameworks to code generator. 

I spend some time digging this issue, which have been reported twice. If I had not factored in that it was probably a fluke, I would have:
- assumed this was a edge case, otherwise the problem would happen at lest randomly with integration tests
- potentially a regression, leading me to review recent changes 
- trying to figure out if it was linked to a change in MTP, xUnit or some other important component 

I had a look at this PR; the fix is obviously simple, so why having so many unit tests. Which covers far more than this simple error. 

My point is: poorly or unsupervised AI leads to extra work for other people. 
In our case, free extra work.

rouke-broersma:
Just to clarify I consider the bug being that we _have_ the code, but aren't using it and aren't testing it which ultimately led to it being completely wrong. The correct fix would have been either to implement test filter end-to-end or remove the dead code. The fact that test filter is not available for MTP is not a bug.

I agree that we are receiving more bug reports and PRs created with agentic support and that it is creating more work for us, with unclear benefit and clear drawbacks. For now I am unsure what to do about it. I think it can be useful as signals that there are issues our users (agentic or otherwise) are struggling with using Stryker. I created a discussion where we can track and discuss what we do about this. #3765

richardwerkman:
This has been fixed by #3752 so I'm closing this now.

Issue #3746, opened 2026-07-28
Nondeterministic crash in solution mode: ArgumentException 'Destination is too short' in InputFileResolver.AnalyzeAllNeededProjects (4.16.0)


martinmdrive:
**Describe the bug**

Running Stryker.NET **4.16.0** in solution mode on a large solution nondeterministically crashes during the initial project analysis — before any mutation runs — with an unhandled `ArgumentException` out of `InputFileResolver.AnalyzeAllNeededProjects`:

```
[19:00:19 ERR] An error occurred during the mutation test run
System.ArgumentException: Destination is too short. (Parameter 'destination')
Unhandled exception. System.ArgumentException: Destination is too short. (Parameter 'destination')
   at Stryker.Core.Initialisation.InputFileResolver.AnalyzeAllNeededProjects(List`1 projects, String normalizedProjectUnderTestNameFilter, IStrykerOptions options, ScanMode mode) in /_/src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs:line 376
   at Stryker.Core.Initialisation.InputFileResolver.ScanInSolutionMode(IStrykerOptions options, SolutionFile solution, String normalizedProjectUnderTestNameFilter) in /_/src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs:line 216
   at Stryker.Core.Initialisation.InputFileResolver.ResolveSourceProjectInfos(IStrykerOptions options) in /_/src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs:line 97
   at Stryker.Core.Initialisation.InitialisationProcess.GetMutableProjectsInfo(IStrykerOptions options) in /_/src/Stryker.Core/Stryker.Core/Initialisation/InitialisationProcess.cs:line 58
   at Stryker.Core.Initialisation.ProjectOrchestrator.MutateProjectsAsync(IStrykerOptions options, IReporter reporters, ITestRunner runner) in /_/src/Stryker.Core/Stryker.Core/Initialisation/ProjectOrchestrator.cs:line 57
   at Stryker.Core.StrykerRunner.RunMutationTestAsync(IStrykerInputs inputs) in /_/src/Stryker.Core/Stryker.Core/StrykerRunner.cs:line 61
   at Stryker.CLI.StrykerCli.RunStrykerAsync(IStrykerInputs inputs) in /_/src/Stryker.CLI/Stryker.CLI/StrykerCLI.cs:line 157
[...]

PR #3751, opened 2026-07-31
feat: Support solution configuration

### LINKED ISSUE #3424: Project Analysis fails on a project using MTP
> When running stryker for our productive repos (we couldn't run it for a while, since we have switched to MTP), we are getting the following error. (I have tested it for 2 different repos, which I cannot share due to data reasons).
> 
> This is the command line:
> ```shell
> dotnet-stryker --output .stryker/run_1 --project Add.Cli.csproj --test-project test/Add.Cli.Test/Add.Cli.Test.csproj --test-project test/StWriter.Test/StWriter.Test.csproj --test-project test/StWriter.Test.Unit/StWriter.Test.Unit.csproj --test-project test/Cli.Test.Unit/Cli.Test.Unit.csproj --test-project test/Cli.Test/Cli.Test.csproj --test-project test/FileIo.Test/FileIo.Test.csproj --test-project test/FileIo.Test.Unit/FileIo.Test.Unit.csproj --test-project test/Add.Cli.Test.Unit/Add.Cli.Test.Unit.csproj  --testrunner mtp --mutation-level Advanced
> ```
> 
> This is the output:
> ```
> Version: 4.12.0
> [05:50:37 INF] Analysis starting.
> [05:50:37 INF] Stryker will mutate solution SafetyCli.
> [05:50:37 INF] Using configuration/platform 'Debug|Any CPU'.
> [05:50:37 INF] Identifying projects to mutate in /SafetyCli.sln. This can take a while.
> [05:50:45 INF] Could not find an assembly reference to a mutable assembly for project /test/Cli.Test.Unit/Cli.Test.Unit.csproj. Will look into project references.
> [05:50:45 INF] Could not find an assembly reference to a mutable assembly for project /test/StWriter.Test/StWriter.Test.csproj. Will look into project references.
> [05:50:45 INF] Could not find an assembly reference to a mutable assembly for project /test/Cli.Test/Cli.Test.csproj. Will look into project references.
> [05:50:45 INF] Could not find an assembly reference to a mutable assembly for project /test/FileIo.Test/FileIo.Test.csproj. Will look into project references.
> [05:50:45 INF] Could not find an assembly reference to a mutable assembly for project /test/StWriter.Test.Unit/StWriter.Test.Unit.csproj. Will look into project references.
> [05:50:45 INF] Could not find an assembly reference to a mutable assembly for project /test/Add.Cli.Test/Add.Cli.Test.csproj. Will look into project references.
> [05:50:45 INF] Could not find an assembly reference to a mutable assembly for project /test/Cli.Test.Apax/Cli.Test.Apax.csproj. Will look into project references.
> [05:50:45 INF] Could not find an assembly reference to a mutable assembly for project /test/FileIo.Test.Unit/FileIo.Test.Unit.csproj. Will look into project references.
> [05:50:45 INF] Project /src/Add.Cli/Add.Cli.csproj analysis failed hence can't be mutated.
> [05:50:45 INF]   referenced by test project /test/Cli.Test.Unit/Cli.Test.Unit.csproj, analysis failed.
> [05:50:45 INF]   referenced by test project /test/Add.Cli.Test/Add.Cli.Test.csproj, analysis failed.
> [05:50:45 WRN]   can't be mutated because all referencing test projects' analysis failed.
> [05:50:45 INF] Test project /test/StWriter.Test/StWriter.Test.csproj does not appear to test any mutable project, analysis failed.
> [05:50:45 INF] Test project /test/Cli.Test/Cli.Test.csproj does not appear to test any mutable project, analysis failed.
> [05:50:45 INF] Test project /test/FileIo.Test/FileIo.Test.csproj does not appear to test any mutable project, analysis failed.
> [05:50:45 INF] Test project /test/StWriter.Test.Unit/StWriter.Test.Unit.csproj does not appear to test any mutable project, analysis failed.
> [05:50:45 INF] Test project /test/Cli.Test.Apax/Cli.Test.Apax.csproj does not appear to test any mutable project, analysis failed.
> [05:50:45 INF] Test project /test/FileIo.Test.Unit/FileIo.Test.Unit.csproj does not appear to test any mutable project, analysis failed.
> [05:50:45 WRN] Use --diag option to have the analysis logs in the log file.
> [05:50:45 INF] Analysis complete.
> [05:50:45 INF] Time Elapsed 00:00:08.2930823
> Stryker.NET failed to mutate your project. For more information see the logs 
> below:
> Failed to analyze project builds. Stryker cannot continue.
> ```
> 
> The diag log can be found as attachment and it took really long ~3:30m. _I have redacted all sensitive data._
> 
> [output.log](https://github.com/user-attachments/files/25044649/output.log) 

 _Originally posted by @Lorilatschki in [#3094](https://github.com/stryker-mutator/stryker-net/issues/3094#issuecomment-3840828741)_

dupdob:
# Summary
Massive refactor of initialization/project discovery/project analysis phase.
So that

- classes have clearer responsibilities
- ensure consistency between solution & project mode
- ensure consistency in configuration/platform/framework handling
- reduce number of edge cases
- simplify problems' diagnosis

## Benefits

- Improve support for solution with better cross projects dependency analysis, especially for multi targeting projects.
- Better feedback when project analysis phase failed.
- support the use of solution configurations, allowing user to add Stryker specific configurations when relevant
- simplify maintenance of this code part for the future

## Fixes
- fix #3424
- improve reporting for issues like #3746 by ensuring the root cause exception's stack is logged as well as the error message

## Remark
I closed and reopened this PR to prevent it from being buried and forgotten


src/.run/FullFrameworkApp.Test.run.xml
richardwerkman, line 1: Ah, you're using Rider? I removed the runsettings files because I thought nobody was using them

src/Stryker.CLI/Stryker.CLI/Logging/LoggingInitializer.cs
```diff
@@ -28,13 +28,17 @@ public void SetupLogOptions(IStrykerInputs inputs, IFileSystem fileSystem = null
         inputs.OutputPathInput.SuppliedInput = outputPath;
 
         var diagnoseMode = inputs.DiagModeInput.Validate();
-        var logLevel = diagnoseMode ? LogEventLevel.Verbose : inputs.VerbosityInput.Validate();
+        var logLevel = inputs.VerbosityInput.Validate();
+        if (diagnoseMode && logLevel > LogEventLevel.Debug)
```
rouke-broersma: What is the intention behind capping the loglevel at debug with diagnose mode instead of respecting the user setting? Do you consider a higher verbosity to be detrimental to diagnosing or something?

src/Stryker.Utilities/Buildalyzer/IAnalyzerResultExtensions.cs
```diff
@@ -186,20 +187,32 @@ public static bool IsValid(this IAnalyzerResult br) => br.Succeeded || (br.Sourc
...
+        || (targetFrameworks.Length>0
+            && Array.TrueForAll(targetFrameworks, fmw => br.Results.Any( r=> r.IsValidFor(fmw))));
 
     public static bool IsTestProject(this IEnumerable<IAnalyzerResult> analyzerResults) => analyzerResults.Any(x => x.IsTestProject());
 
     private static bool IsTestProject(this IAnalyzerResult analyzerResult)
     {
-        if (bool.TryParse(analyzerResult.GetPropertyOrDefault("IsTestingPlatformApplication"), out var isTestingPlatformApplication) && isTestingPlatformApplication)
+        // if 'IsTestingPlatformApplication' is defined and true, this is a test project
+        if (bool.TryParse(analyzerResult.GetPropertyOrDefault("IsTestingPlatformApplication", ""), out var isMtp) && isMtp)
```
rouke-broersma: Shouldn't this also use the new `TryGetProperty`? Seems like the same construct as IsTestproject.

src/Stryker.CLI/Stryker.CLI/Logging/LoggingInitializer.cs
dupdob: No. I think my intent was the opposite: make sure log level is debug **at least** instead of forcing verbose, which results in massive logs.
Fixed the typo

src/Stryker.Utilities/Buildalyzer/IAnalyzerResultExtensions.cs
dupdob: changed it. It is more consistent, not sure if it is clearer though

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialisationProcessTests.cs
```diff
@@ -162,18 +179,29 @@ public async Task InitialisationProcess_ShouldThrowOnTestTestIfAskedFor(bool bre
         var folder = new FolderComposite();
         folder.Add(new CsharpFileLeaf());
 
+        var loggerMock = new Mock<ILogger<InitialisationProcess>>(); var target = new InitialisationProcess(inputFileResolverMock.Object, initialBuildProcessMock.Object, initialTestProcessMock.Object, loggerMock.Object);
```
richardwerkman: Split these lines

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialisationProcessTests.cs
```diff
@@ -219,26 +239,39 @@ public async Task InitialisationProcess_ShouldRunTestSession()
         var folder = new FolderComposite();
         folder.Add(new CsharpFileLeaf());
 
-        inputFileResolverMock.Setup(x => x.ResolveSourceProjectInfos(It.IsAny<StrykerOptions>())).Returns(
-            new[] { new SourceProjectInfo() { AnalyzerResult = TestHelper.SetupProjectAnalyzerResult(references: Array.Empty<string>()).Object, TestProjectsInfo = new TestProjectsInfo(new MockFileSystem()) } });
+        var loggerMock = new Mock<ILogger<InitialisationProcess>>();
```
richardwerkman, line 244: Each test contains a lot of plumbing for the target setup and mock creation. I think we can do that once int the class constructor to keep each test clean

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InputFileResolverTests.cs
```diff
@@ -1080,22 +1081,24 @@ public void ShouldSelectAvailableFramework_WhenDesiredNotFound(string targetFram
...
-    [DataRow("net3.0,net462", "net461,net2.0", "net3.0", "net3.0", "net2.0")]
-    [DataRow("net3.0,net462", "net461,net2.0", "net461", "net3.0", "net2.0")]
-    [DataRow("net3.0,net462", "net461,net2.0", "net462", "net462", "net461")]
-    public void ShouldSelectFrameworkBasedOnTestProject(string testFrameworks, string projectFrameworks
+    [DataRow("netcoreapp3.0,net462", "netcoreapp2.0,net461", null, "netcoreapp3.0", "netcoreapp2.0")]
+    [DataRow("netcoreapp3.0,netcoreapp2.0", "netcoreapp3.0,netcoreapp2.0", null, "netcoreapp3.0", "netcoreapp3.0")]
+    [DataRow("netcoreapp3.0,net462", "netcoreapp2.0,net461", "netcoreapp2.0", "netcoreapp3.0", "netcoreapp2.0")]
+    [DataRow("netcoreapp3.0,net462", "netcoreapp2.0,net461", "netcoreapp3.0", "netcoreapp3.0", "netcoreapp2.0")]
+    [DataRow("netcoreapp3.0,net462", "netcoreapp2.0,net461", "net461", "net462", "net461")]
+    [DataRow("netcoreapp3.0,net462", "netcoreapp2.0,net461", "net462", "net462", "net461")]
```
richardwerkman, line 911: Why did the net2.0 and net3.0 rows change to netcoreapp? was the previous testdata incorrect?

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InputFileResolverTests.cs
```diff
@@ -1132,7 +1135,59 @@ public void ShouldSelectFrameworkBasedOnTestProject(string testFrameworks, strin
...
+        var result = target.ResolveSourceProjectInfos(options).SourceProjectInfos.First();
+
+        // Assert
+        result.AnalyzerResult.TargetFramework.ShouldBe(expectedFramework);
+        result.TestProjectsInfo.AnalyzerResults.First().TargetFramework.ShouldBe(expectedTestFramework);
+    }
+
+    // Stryker accepts netframework targert when running on Windows
+    [TestMethodWithIgnoreIfSupport]
+    [IgnoreIf(nameof(Is.NotWindows))]
```
richardwerkman, line 955: Why can't this run on non-windows? Buildalizer is mocked right?

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InputFileResolverTests.cs
```diff
@@ -1132,7 +1135,59 @@ public void ShouldSelectFrameworkBasedOnTestProject(string testFrameworks, strin
...
         // Act
-        var result = target.ResolveSourceProjectInfos(options).First();
+        var result = target.ResolveSourceProjectInfos(options).SourceProjectInfos.First();
+
+        // Assert
+        result.AnalyzerResult.TargetFramework.ShouldBe(expectedFramework);
+        result.TestProjectsInfo.AnalyzerResults.First().TargetFramework.ShouldBe(expectedTestFramework);
+    }
+
+    // Stryker accepts netframework targert when running on Windows
```
richardwerkman:
```suggestion
    // Stryker accepts netframework target when running on Windows
```

src/Stryker.Utilities/Buildalyzer/IAnalyzerResultExtensions.cs
rouke-broersma: Ahh right I missed that we need to fall-through on the isMtp case to further check if it's not a vstest testproject so we can't use the exact same construct.

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -77,64 +68,70 @@ public IReadOnlyCollection<SourceProjectInfo> ResolveSourceProjectInfos(IStryker
...
+
+    private IReadOnlyCollection<SourceProjectInfo> FindProjectsInSolutionMode(IStrykerOptions options, ProjectsTracker solutionInfo,
         string normalizedProjectUnderTestNameFilter)
     {
-        StrykerSolutionInfo solutionInfo = null;
-        var configuration = options.Configuration;
-        // "Any CPU" is the solution-level name; MSBuild requires "AnyCPU"
-        var platform = NormalizePlatform(options.Platform);
+        _logger.LogInformation("Identifying projects to mutate in {Solution}. This can take a while.",
+            FileSystem.Path.GetFileNameWithoutExtension(solutionInfo.SolutionFilePath));
```
richardwerkman: Why without extension? There might be a .sln, .slnx and/or .slnf file. Could be useful to show clearly what we are analyzing

src/Stryker.Utilities/Buildalyzer/IAnalyzerResultExtensions.cs
rouke-broersma: Perhaps that indicates it should be a switch with a fallthrough case instead?

src/Stryker.Solutions/SolutionFile.cs
```diff
@@ -221,6 +267,32 @@ private static SolutionFile AnalyzeSolution(string solutionPath, SolutionModel s
         return result;
     }
 
+    private static Dictionary<string, (string buildType, string platform)> ExtractProjectForGivenConfiguration(SolutionModel solution,
```
rouke-broersma:
```suggestion
    private static Dictionary<string, (string buildType, string platform)> ExtractProjectsForGivenConfiguration(SolutionModel solution,
```

src/Stryker.Solutions/SolutionFile.cs
```diff
@@ -155,61 +156,106 @@ private string GetEffectiveBuildType(string? buildType)
...
-                {
-                    var (projectBuildType, projectPlatform, isBuilt, _) = solutionProject.GetProjectConfiguration(buildType, solutionPlatform);
-                    if (!isBuilt || projectBuildType == null || projectPlatform == null)
-                    {
-                        continue;
-                    }
-
-                    projects[solutionProject.FilePath] = (projectBuildType, projectPlatform);
-                }
+                var projects = ExtractProjectForGivenConfiguration(solution, referencePath, buildType, solutionPlatform);
```
rouke-broersma:
```suggestion
                var projects = ExtractProjectsForGivenConfiguration(solution, referencePath, buildType, solutionPlatform);
```

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -77,64 +68,70 @@ public IReadOnlyCollection<SourceProjectInfo> ResolveSourceProjectInfos(IStryker
...
-        if (solution != null)
-        {
-            var (actualBuildType, actualPlatform) = solution.GetMatching(options.Configuration, options.Platform);
-            _logger.LogDebug("Using solution configuration/platform '{Configuration}|{Platform}'.", actualBuildType, actualPlatform);
-            solutionInfo = new StrykerSolutionInfo(solution.FileName, actualBuildType, actualPlatform);
-            configuration = actualBuildType;
-            platform = NormalizePlatform(actualPlatform);
-        }
+        // analyze all projects
+        solutionInfo.SelectAllProjects();
```
richardwerkman, line 102: What does this mean? Selecting is analyzing? Or is the comment above incorrect

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -77,64 +68,70 @@ public IReadOnlyCollection<SourceProjectInfo> ResolveSourceProjectInfos(IStryker
...
+        _logger.LogDebug("Analyzing {ProjectsCount} projects.", solutionInfo.ProjectCount);
+        // we analyze every project in the solution
```
richardwerkman:
```suggestion
        // analyze every project in the solution
```

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -77,64 +68,70 @@ public IReadOnlyCollection<SourceProjectInfo> ResolveSourceProjectInfos(IStryker
...
+        var mutableProjectsAnalyzerResults = AnalyzeAllNeededProjects(solutionInfo,
+            normalizedProjectUnderTestNameFilter,
+            options,
+            ScanMode.NoScan);
+        // we identify target projects and their associated test projects
```
richardwerkman:
```suggestion
        // identify target projects and their associated test projects
```

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -77,64 +68,70 @@ public IReadOnlyCollection<SourceProjectInfo> ResolveSourceProjectInfos(IStryker
...
+        var (findMutableAnalyzerResults, orphanedProjects) =
+            ExtractMutableProjectTrees(mutableProjectsAnalyzerResults);
+        // we keep only suitable candidates
```
richardwerkman:
```suggestion
        // keep only suitable candidates
```

src/Stryker.Solutions/SolutionFile.cs
```diff
@@ -155,61 +156,106 @@ private string GetEffectiveBuildType(string? buildType)
...
     /// Create a solution file from a list of projects having two build types, Debug and Release, and the provided platforms.
     /// </summary>
+    /// <param name="filePath">solution file name</param>
     /// <param name="projects">list of csproj filenames</param>
     /// <param name="platforms">list of declared platforms. Default to AnyCpu and x86</param>
+    /// <param name="solutionPlatforms">list of solution level platforms</param>
     /// <returns>a solution instance</returns>
     /// <remarks>This method is used for testing purposes. It is mandatory as the underlying solution parser does not support any form of mocking</remarks>
-    public static SolutionFile BuildFromProjectList(List<string> projects, string[]? platforms = null)
+    public static SolutionFile BuildFromProjectList(string filePath, List<string> projects, string[]? platforms = null
```
rouke-broersma: Perhaps consider making this internal so it's only available externally from testing assemblies due to our internals visible from build props

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -238,30 +222,32 @@ private enum ScanMode
...
+
+        // do we have at least one valid target?
+        if (suitableCandidates.Count == 0)
         {
             // no mutable project found
             LogAnalysis(findMutableAnalyzerResults, unusedTestProjects, options.DiagMode);
             throw new InputException("Failed to analyze project builds. Stryker cannot continue.");
         }
 
+        // we keep only on target framework per project
```
richardwerkman:
```suggestion
        // we keep only one target framework per project
```

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -344,405 +307,216 @@ private void LogAnalysis(Dictionary<IAnalyzerResult, List<IAnalyzerResult>> find
...
-        var env = new EnvironmentOptions();
+            // check the new status
+            buildResultOverallSuccess = project.HasValidResults();
 
-        if (!string.IsNullOrEmpty(options.MsBuildPath))
+            if (!buildResultOverallSuccess && !string.IsNullOrEmpty(options.TargetFramework))
+            {
+                // still failed, we can try using target framework option
+                _logger.LogWarning("Project {ProjectFilePath} simulated build failed again. Last attempt, forcing the target framework.", projectLogName);
+                buildResult = project.Analyze(forceFramework: true);
```
richardwerkman, line 385: Why don't we set this by default?

src/Stryker.Core/Stryker.Core/Initialisation/ProjectSimulatedBuildHandler.cs
```diff
@@ -0,0 +1,229 @@
...
+using Microsoft.Extensions.Logging;
+using Stryker.Utilities.Buildalyzer;
+
+namespace Stryker.Core.Initialisation;
+
+/// <summary>
+/// Encapsulates the context of a project analysis, including the project analyzer, the last analysis results
+///  for said project. It also provides reference discovery methods
+/// </summary>
+public class ProjectSimulatedBuildHandler
```
richardwerkman: I find this name a bit confusing. It sounds like an event handler, but it isn't. I'd suggest renaming to `ProjectSimulatedBuildResult`

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -344,405 +307,216 @@ private void LogAnalysis(Dictionary<IAnalyzerResult, List<IAnalyzerResult>> find
...
-    private IAnalyzerResult SelectAnalyzerResult(IEnumerable<IAnalyzerResult> analyzerResults, string targetFramework)
-    {
-        var validResults = analyzerResults.ToList();
-        var projectName = analyzerResults.First().ProjectFilePath;
-        if (validResults.Count == 0)
-        {
-            throw new InputException($"No valid project analysis results could be found for '{projectName}'.");
         }
+        var mutableToTestMap = mutableProjects.ToDictionary(p =>p, p => new MutableProjectTree(p, _logger));
+        var unusedTestProjects = new List<ProjectSimulatedBuildHandler>();
```
richardwerkman: This type is confusing here. I've suggested a rename in that class to clarify this

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -344,405 +307,216 @@ private void LogAnalysis(Dictionary<IAnalyzerResult, List<IAnalyzerResult>> find
...
-        var mutableProject = mutableProjects.FirstOrDefault(p => testProject.ProjectReferences.Contains(p.ProjectFilePath));
-        if (mutableProject == null)
-        {
-            return false;
-        }
-        if (!mutableToTestMap.TryGetValue(mutableProject, out var dependencies))
+        if (testProject.AnalyzerLastResults.Count == 0)
         {
-            mutableToTestMap[mutableProject] = dependencies = [];
+            throw new InvalidOperationException($"You must analyze the test project {testProject.ProjectFileName} before trying to find its references.");
```
richardwerkman: Is this correct? How can the user analyze the test project? I think this should be something like:

```suggestion
            throw new InvalidOperationException($"Failed to analyze the test project {testProject.ProjectFileName} while trying to find its references.");
```

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
```diff
@@ -344,405 +307,216 @@ private void LogAnalysis(Dictionary<IAnalyzerResult, List<IAnalyzerResult>> find
...
-        };
-
-        var language = targetProjectInfo.AnalyzerResult.GetLanguage();
-        if (language == Language.Fsharp)
+        if (testProject.AnalyzerLastResults.Count == 0)
         {
-            _logger.LogError(
-                targetProjectInfo.LogError(
-                    "Mutation testing of F# projects is not ready yet. No mutants will be generated."));
+            throw new InvalidOperationException($"You must analyze the test project {testProject.ProjectFileName} before trying to find its references.");
```
richardwerkman: Same here:

```suggestion
            throw new InvalidOperationException($"Failed to analyze the test project {testProject.ProjectFileName} while trying to find its references.");
```

src/Stryker.Core/Stryker.Core/Initialisation/MutableProjectTarget.cs
```diff
@@ -0,0 +1,94 @@
...
+/// </summary>
+/// <param name="target">analyzer result for target project</param>
+/// <param name="logger">logger used for reporting</param>
+internal class MutableProjectTarget(IAnalyzerResult target, ILogger logger)
+{
+    public IAnalyzerResult ProjectTarget { get; } = target;
+
+    public HashSet<IAnalyzerResult> TestProjects { get; } = [];
+
+    public bool IsValidTarget => ProjectTarget.IsValid() && TestProjects.Count>0 && TestProjects.Any(tp => tp.IsValid());
```
richardwerkman:
```suggestion
    public bool IsValidTarget => ProjectTarget.IsValid() && TestProjects.Count > 0 && TestProjects.Any(tp => tp.IsValid());
```

src/Stryker.Core/Stryker.Core/Initialisation/ProjectsTracker.cs
```diff
@@ -0,0 +1,162 @@
...
+            {
+                _logger.LogInformation("Using solution configuration/platform '{Configuration}|{Platform}'.", Configuration, Platform);
+            }
+        }
+        else
+        {
+            Configuration = configuration;
+            // "Any CPU" is default platform at solution level, but in project files it is "AnyCPU", so we need to convert it to match the project files
+            Platform = platform == "Any CPU" ? "AnyCPU" : platform;
+            _logger.LogInformation("Using project configuration/platform '{Configuration}|{Platform}'.", Configuration??"`default`", Platform ?? "`default`");
```
richardwerkman: Why are we logging this? In a testrun I saw the output:

```cmd
Using project configuration/platform '`default`|`default`'.
```

It is a bit confusing. When we pick default I guess we can just log this as debug since it doesn't add much value to log this.

src/Stryker.Core/Stryker.Core/Initialisation/ProjectSimulatedBuildWrapper.cs
```diff
@@ -0,0 +1,229 @@
...
+        if (projectFileTargetFrameworks.Length > 0)
+        {
+            _logger.LogDebug("Project {ProjectFilePath} supported frameworks: {FrameworkList}.", ProjectFileName, string.Join(',', projectFileTargetFrameworks));
+        }
+        else
+        {
+            if (!string.IsNullOrEmpty(_framework))
+            {
+                projectFileTargetFrameworks=[_framework];
+                _logger.LogWarning("Failed to identify target frameworks for project {ProjectFilePath}. Assuming selected framework ({Framework}) is present.", ProjectFileName, _framework);
```
richardwerkman, line 106: When testing this branch on regex-parser I saw this warining multiple times. Does this have to be a warning? It ends up picking the correct framework anyway.

src/Stryker.Core/Stryker.Core/Initialisation/MutableProjectTarget.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        var inputFiles = builder.Build();
+        builder.InjectHelpers(inputFiles);
+        targetProjectInfo.OnProjectBuilt = builder.PostBuildAction();
+        targetProjectInfo.ProjectContents = inputFiles;
+        logger.LogInformation("Found project {ProjectFileName} to mutate.", ProjectTarget.ProjectFilePath);
+        return targetProjectInfo;
+    }
+
+    public void DumpForAnalysis()
```
richardwerkman: Why is this called dump? It sounds like it is for debugging, but it logs in level Information

src/Stryker.Core/Stryker.Core/Initialisation/MutableProjectTree.cs
```diff
@@ -0,0 +1,84 @@
...
+        if (targetToKeep == null)
+        {
+            logger.LogWarning("Failed to find a valid target for project {Project}. ", project.ProjectFileName);
+            return;
+        }
+        logger.LogInformation("Picking {Framework} for project {Project}. ", targetToKeep.ProjectTarget.TargetFramework, project.ProjectFileName);
+        Targets.Add(targetToKeep);
+    }
+
+    public void DumpForAnalysis()
```
richardwerkman: Same as the other comment, what does this do? Shouldn't this be debug logging?

src/Stryker.Core/Stryker.Core/ProjectComponents/SourceProjects/RelatedSourceProjectsInfo.cs
```diff
@@ -0,0 +1,50 @@
...
+public class RelatedSourceProjectsInfo(
+    ProjectsTracker projectsTracker,
+    IReadOnlyCollection<SourceProjectInfo> sourceProjectInfos,
+    ILogger logger = null)
+{
+    public IReadOnlyCollection<SourceProjectInfo> SourceProjectInfos { get; } = sourceProjectInfos;
+
+    private ProjectsTracker Tracker { get; } = projectsTracker;
+
+    public bool BuildTestProjects(IInitialBuildProcess buildProcess)
```
richardwerkman, line 22: Is this a logical place for this? The classname suggests this is an aggregator. But it mainly builds the whole solution. I suggest renaming this class or moving this somewhere else

src/Stryker.Solutions/SolutionFile.cs
```diff
@@ -155,61 +156,106 @@ private string GetEffectiveBuildType(string? buildType)
...
     /// <remarks>This method is used for testing purposes. It is mandatory as the underlying solution parser does not support any form of mocking</remarks>
-    public static SolutionFile BuildFromProjectList(List<string> projects, string[]? platforms = null)
+    internal static SolutionFile BuildFromProjectList(string filePath, List<string> projects, string[]? platforms = null
+        , string[]? solutionPlatforms = null)
     {
-        var result = new SolutionFile();
-        platforms ??= [ DefaultSolutionType, "x86" ];
+        var result = new SolutionFile {FileName = filePath};
+        solutionPlatforms = DefineSolutionPlatforms(platforms, solutionPlatforms);
+        platforms ??= [ "AnyCPU", "x86" ];
```
richardwerkman, line 202: This is duplicate to line 235, can be removed

src/Stryker.Solutions/SolutionFile.cs
```diff
@@ -155,61 +156,106 @@ private string GetEffectiveBuildType(string? buildType)
...
+    private static string[] DefineSolutionPlatforms(string[]? platforms, string[]? solutionPlatforms)
+    {
+        if (solutionPlatforms != null)
+        {
+            return solutionPlatforms;
+        }
+
+        if (platforms == null)
+        {
+            solutionPlatforms = [ "Any CPU", "x86" ];
```
richardwerkman, line 237: Maybe add a comment why this is the default.

src/Stryker.TestRunner.MicrosoftTestPlatform.UnitTest/MicrosoftTestPlatformRunnerPoolTests.cs
```diff
@@ -176,14 +177,30 @@ public void Dispose_ShouldDisposeAllRunnersInPool()
...
         // The pool uses Parallel.For to create runners, which should complete before constructor returns
         // However, to be defensive against timing issues, verify by checking the actual runners in the pool
-        pool.Runners.Count().ShouldBe(3, "All 3 runners should be available in the pool");
+
+        var timeout = new Stopwatch();
+        timeout.Start();
+        var start = timeout.ElapsedMilliseconds;
+        lock (createdRunners)
+        {
+            while (createdRunners.Count<3 && timeout.ElapsedMilliseconds-start<2000)
```
richardwerkman:
```suggestion
            while (createdRunners.Count < 3 && timeout.ElapsedMilliseconds-start < 2000)
```

richardwerkman (dismissed):
Thanks for yet another great contribution 🥇 

Reviewing took a while, but here are some suggestions to make this more readable and understandable. 

I have another question: do you have some exact edge cases that have been fixed with this PR? And can we add testcases for them so they stay fixed?

src/.run/FullFrameworkApp.Test.run.xml
dupdob: yep, using Rider (on Mac)

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InputFileResolverTests.cs
dupdob: Yes: it turns out net2.0 is recognized as NetFramework 2.0 via `NuGetFramework.Parse` , while older assumed it meant NetCore2.0. 
Long story short: I had to use the correct moniker to keep the test valid. The alternative was adjusting the expected results, but then the data would have been variant of the same situations (netframework everywhere)

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InputFileResolverTests.cs
dupdob: Stryker raises an exception when trying to build net framework projects outside Windows.
Otherwise, the analysis simply fails but the diagnostic is hard to establish for end users.

src/Stryker.Solutions/SolutionFile.cs
dupdob: are you really sure about this? I suggest you have a second look.
Someone add the great idea of using 'Any CPU' as solution level moniker for 'AnyCPU' platform.

src/Stryker.Solutions/SolutionFile.cs
dupdob: done

src/Stryker.Core/Stryker.Core/Initialisation/ProjectSimulatedBuildHandler.cs
dupdob: Agree name could be better, but this is far more than a build result. It now encapsulates the analysis itself.
I will use wrapper, which is still vague, but more acceptable, I think

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
dupdob: Note that the path is also missing, but the full file path is logged at debug level (extension incl.)

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
dupdob: I added a comment to explain. This is not new. When you specify a target framework for a build/analysis, it will be used disregarding the frameworks selected in the project. Which may mess dependency resolution for projects/solution using different frameworks for different assemblies.
So, `targetframework` is used as a _hint_ (favor this version)

src/Stryker.Core/Stryker.Core/Initialisation/InputFileResolver.cs
dupdob: yes, I suspect I blindly accepted some Copilot suggestion

src/Stryker.Core/Stryker.Core/Initialisation/ProjectsTracker.cs
dupdob: Changed the log so that it appears only when a confiugration/platform is configured

src/Stryker.Core/Stryker.Core/Initialisation/ProjectSimulatedBuildWrapper.cs
dupdob: This is a warning because this is a sign that analysis may have gone wrong. A correct analysis should result in a non empty list of supported frameworks.

Not sure what happens with your test, but it would be interesting to dig a bit. Anyway, I propose we keep the warning for now and adjust/correct it on an opportunity basis.
Could be kept as debug, but this may be an interesting fact when people report problems

src/Stryker.Core/Stryker.Core/ProjectComponents/SourceProjects/RelatedSourceProjectsInfo.cs
dupdob: This class is the equivalent of a solution even when no solution file are being used: it keeps together a bunch of related project (hence the name).

This class exists to:

- reduce the responsibilities of InputFileResolver
- abstract the project build step so that InputFileResolver workflow depends less on the presence of a solution file (or not)

I will not strongly defend this naming, but I have no other suggestion to make here.

src/Stryker.Core/Stryker.Core/Initialisation/MutableProjectTree.cs
dupdob: This method is called when the project analys step failed and provide a detailed assessment on which projects succeed and failed. So it is a status dump, which should guide user addressing potential issues, such as some projects requiring some properties being defined or other type of prerequisites.
So, it is info level in order to be useful in general. 
Stryker raises an exception after dumping this info.

dupdob: I am open to suggestion if a new name is really needed

src/Stryker.Core/Stryker.Core/Initialisation/MutableProjectTarget.cs
dupdob: This method is called when the project analys step failed and provide a detailed assessment on which projects succeed and failed. So it is a status dump, which should guide user addressing potential issues, such as some projects requiring some properties being defined or other type of prerequisites.
So, it is info level in order to be useful in general.
Stryker raises an exception after dumping this info.

src/Stryker.Solutions/SolutionFile.cs
richardwerkman: Ah sorry now I see they assign to different variables. But why should one be "Any CPU" and the other be "AnyCPU"? Isn't it better to keep that the same everywhere?

src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialisationProcessTests.cs
dupdob: working on it

dupdob: done, as in, removed a lots of near duplicates

dupdob:
@richardwerkman : I addressed your remarks.
Refactoring the tests allowed me to delete a lot of lines, which is always good. InputfileResolverTests definitely looks better.

src/Stryker.Solutions/SolutionFile.cs
rouke-broersma: @richardwerkman iirc this is Microsoft doing, not user specific config. At project level its without space and at solution level is with space (or inverse, something like that at least)

dupdob: Yep. MS chose Any CPU with a space at the solution level instead of AnyCPU. 
And a lot of funny problems arises from that. 
This PR now replaces ´Any CPU' with 'AnyCPU' at the project level. 
This may happen if a user specifies 'Any CPU' in project mode. When a solution is involved, configuration resolution happens earlier

dupdob:
🍾

PR #3752, opened 2026-07-31
feat(MTP): `perTest` and `perTestInIsolation` coverage

richardwerkman:
This PR adds two new coverage-analysis modes for the new Microsoft Testing Platform (MTP) runner, matching modes that already existed for VSTest: `perTest` and `perTestInIsolation`. They're selected via the existing `--coverage-analysis option`, which defaults to perTest.

todo:

- [x] Investigate issue with perTestInIsolation hanging indefinitely sometimes (solved, only debug mode issue)
- [ ] Fix integration tests
- [x] Run MPT unit tests in azure pipeline
- [x] Improve coverage

## perTestInIsolation

Before every single test, the assembly's test-host process is torn down (DiscardServerAsync) and restarted fresh.
The test runs, then the server is stopped gracefully so MutantControl's ProcessExit handler flushes coverage to a plain coverage file, which is read back immediately.
Because the process is fully isolated per test, no static state can leak between tests, so results get `CoverageConfidence.Exact`.
This is slow (full process restart per test) but maximally accurate.

## perTest (reused process)

The test-host process for an assembly is kept warm across many tests (no restart), for speed.
Tests still run strictly one at a time on a given server, but since the process never exits, there's no ProcessExit event to trigger a coverage flush. The runner needs a way to tell the still-running test host "that test is done, publish what it covered and clear the slate" between each test.
That signal is the epoch file. Because state can leak between tests on a reused process, results only get `CoverageConfidence.Normal` (vs. Exact for the isolated mode).

## The implementation

The epoch file is an 8-byte memory-mapped file, one per (runner, assembly) pair, named `stryker-epoch-{runnerId}-{sanitizedAssemblyName}.txt`. It holds two 4-byte integers:

offset 0 - the request: written by the Stryker runner process
offset 4 - the ack: written by the injected MutantControl code running inside the test host process

It's essentially a lock-free, poll-based cross-process counter handshake, needed because named OS synchronization primitives (`EventWaitHandle` etc.) aren't reliable cross-platform, particularly on Unix

After running a test on the warm server, the runner increments a per-assembly counter and calls `WriteEpochRequest(epochFilePath, epoch)` bumping the request half.
`WaitForEpochAckAsync` polls (1ms) until the ack half equals that epoch number, with a 10s timeout. If it times out, the test's coverage is reported as CoverageConfidence.Dubious instead of hanging forever.
Once acked, the runner reads the per-test coverage file that MutantControl just flushed.

## Tested

I've tested this on multiple projects and the mutation score seems stable between runs with perTest analysis. So I see no reason to force perTestInIsolation for MTP.

## Other changes

minor JSON contract fix (test-case property renamed to "tests", null location fields omitted) needed for the per-test filtering.


richardwerkman:
I'm seeing an issue where the dotnet runtime crashes during coverage analysis due to thread scheduling. But it only seems to happen in debug mode. In release mode the issue doesn't occur. So I'm leaving that for now.

The coverage is reported as 0% because the MTP unit tests are not running in the pipeline. I'm fixing that now.

PR #3767, opened 2026-08-11
fix(mutators): support mutating C# 11 utf-8 string literals

### LINKED ISSUE #2923: UTF-8 string literals are not mutated
**Describe the bug**
I've noticed that UTF-8 string literals are not mutated.

**Expected behavior**
I'd expect the following line to be mutated by the string literal mutator:

``` cs
var stringLiteral = "mystring"u8;
```

PSR94:
## Summary
Adds support for mutating C# 11 UTF-8 string literals (e.g. `u8"mystring"`) in the StringMutator.

## Why this helps
UTF-8 string literals were being ignored by Stryker because their Roslyn `SyntaxKind` was not recognized as a string expression. This allows Stryker to find and mutate UTF-8 string literals just like regular strings.

## Changes made
- Updated `RoslynHelper.IsAStringExpression` to include `SyntaxKind.Utf8StringLiteralExpression`.
- Updated `StringMutator.ApplyMutations` to generate a replacement `Utf8StringLiteralExpression` when the original node is a UTF-8 string literal.
- Added tests in `StringMutatorTests`.

## Testing
- Added unit tests.
- Ran `dotnet test` on `Stryker.Core.UnitTest` (Passed).

Closes #2923

### DIFF
--- a/integrationtest/TargetProjects/NetCore/NetCoreTestProject.MSTest/Constructs.cs
+++ b/integrationtest/TargetProjects/NetCore/NetCoreTestProject.MSTest/Constructs.cs
@@ -16,5 +16,13 @@ namespace NetCoreTestProject.MSTest
 
             Assert.IsTrue(expired == result);
         }
+
+        [TestMethod]
+        public void GetHelloUtf8_ShouldReturnHello()
+        {
+            var result = TargetProject.Constructs.CSharp11.GetHelloUtf8();
+            var expected = "Hello"u8.ToArray();
+            Assert.IsTrue(System.Linq.Enumerable.SequenceEqual(result.ToArray(), expected));
+        }
     }
 }
--- a/integrationtest/TargetProjects/NetCore/NetCoreTestProject.NUnit/Constructs.cs
+++ b/integrationtest/TargetProjects/NetCore/NetCoreTestProject.NUnit/Constructs.cs
@@ -23,6 +23,14 @@ namespace NetCoreTestProject.NUnit
             ClassicAssert.AreEqual(expired, result);
         }
 
+        [Test]
+        public void GetHelloUtf8_ShouldReturnHello()
+        {
+            var result = TargetProject.Constructs.CSharp11.GetHelloUtf8();
+            var expected = "Hello"u8.ToArray();
+            ClassicAssert.AreEqual(expected, result.ToArray());
+        }
+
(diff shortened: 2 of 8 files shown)


rouke-broersma:
Before we review this, I would kindly request that you confirm or deny whether or not you have looked at the old abandoned pr  https://github.com/stryker-mutator/stryker-net/pull/2940

And made sure that any review on that pr is incorporated or not relevant to your contribution. If you haven't, please check that pr first and make sure you incorporate the feedback from there first and let us know when you have.

PSR94:
I reviewed the abandoned PR #2940 and the feedback from @richardelekta. 

The issue they encountered where mutating a UTF-8 string inside a concatenation (e.g. `"Hello"u8 + " "u8`) causes a compiler error is because the C# compiler only supports `+` concatenation for UTF-8 string literals when they are **compile-time constants**. When Stryker injects a mutation via the ternary operator (e.g. `(MutantControl.IsActive(1) ? ""u8 : "Hello"u8) + " "u8`), the expression is no longer a compile-time constant, and the compiler raises an error because the `+` operator is not defined for `ReadOnlySpan<byte>` at runtime.

To address this feedback and ensure we don't generate uncompilable mutants, I will update this PR to explicitly skip mutating `Utf8StringLiteralExpression` if it is part of an `AddExpression` (`+`). This is an inherent limitation of C# 11 UTF-8 literals. I'll push the fix to this branch momentarily!

PSR94:
Just a quick update: I've also reverted the expected mutant counts in `ValidateStrykerResults.cs` back to the master values. Since we're skipping the UTF-8 mutations inside the `AddExpression` (which were the only UTF-8 literals generating new mutants in the integration tests), the generated mutant count exactly matches the original master baseline.

rouke-broersma:
Please add a new mutant in the correct csharp version integration test file so a valid mutant is generated by the mutator for utf8 strings.

PSR94:
Done! I've added a `GetHelloUtf8` method to `CSharp11.cs` that returns a valid UTF-8 string literal (`"Hello"u8`), and added corresponding unit tests to the xUnit, NUnit, and MSTest test projects to ensure the literal is mutated and caught. I've also updated the integration tests' expected mutant counts in `ValidateStrykerResults.cs` to account for the new killed mutant!

PR #3769, opened 2026-08-11
fix(MTP): MTP coverage per mutated assembly

### LINKED ISSUE #3753: MTP runner: multi-project solutions report covered mutants as "No coverage"
**Describe the bug**
With the MTP runner and coverage analysis on (the default), a solution with more than one
mutated project loses coverage for all but one assembly: the lost assembly's mutants are
reported *No coverage* and never tested, even when tests call that code directly.

Cause: the test host loads one compiled copy of the injected `MutantControl` per mutated
assembly, all sharing the host's single `STRYKER_COVERAGE_FILE`.
[`FlushCoverageToFile`](https://github.com/stryker-mutator/stryker-net/blob/ee06d3a10cb09a1364c468298dd226597cf1920d/src/Stryker.Core/Stryker.Core/InjectedHelpers/MutantControl.cs#L245)
writes it with `File.WriteAllText`, so the last flush overwrites the rest, and
[`ReadCoverageData`](https://github.com/stryker-mutator/stryker-net/blob/ee06d3a10cb09a1364c468298dd226597cf1920d/src/Stryker.TestRunner.MicrosoftTestPlatform/SingleMicrosoftTestPlatformRunner.cs#L249)
reads the one surviving value. #3696's per-instance files don't reach this: both copies live
in the same host and write that host's one file.

**Reproduction — on this repo's own MTP fixture**
The MTPSolution fixture already mutates TargetProject and Library, but no sample test
executes Library code, so no host ever loads two copies — which is why CI has never seen
this. The branch below is **master (`ee06d3a1`) plus
[one commit](https://github.com/npnelson/stryker-net/commit/51a4bacde2639adf649f86066610487e713df8ff)**
that adds exactly that test, the Library `ProjectReference` it needs, and a coverage-off
config. No product code is touched.

```bash
git clone --branch repro/mtp-second-assembly --single-branch https://github.com/npnelson/stryker-net.git mtp-repro
cd mtp-repro

# build the CLI once, quietly - otherwise every run reprints the repo's own build warnings
dotnet build src/Stryker.CLI/Stryker.CLI -c Release -v q -p:WarningLevel=0 -p:NoWarn=NU1608 -p:NuGetAudit=false
cd integrationtest/TargetProjects

# ground truth - coverage analysis off (~40 s)
dotnet run --no-build --project ../../src/Stryker.CLI/Stryker.CLI -c Release -- --config-file stryker-config-coverage-off.json --solution MicrosoftTestPlatform.slnx --test-runner mtp --mutate "**/RecursiveMath.cs"

# the bug - identical but for the config, coverage analysis on (the default, ~40 s)
dotnet run --no-build --project ../../src/Stryker.CLI/Stryker.CLI -c Release -- --solution MicrosoftTestPlatform.slnx --test-runner mtp --mutate "**/RecursiveMath.cs"
```

`RecursiveMath.Fibonacci` always returns 0, so the added test captures `Console.Out` and
asserts the sequence it prints — which makes its mutants genuinely killable rather than merely
covered. The two runs differ only in the config file.

Both end with Stryker's per-file summary table, most of whose rows are `Excluded` because
`--mutate` scoped the run to `RecursiveMath.cs`. **Two things to look at:** the
`total mutants will be tested` line, and the `RecursiveMath.cs` row of that table.

```
# coverage analysis off - ground truth
INF] 7     total mutants will be tested
INF] The final mutation score is 100.00 %

# coverage analysis on - the default
INF] 7     mutants got status NoCoverage.   Reason: Not covered by any test.
INF] 0     total mutants will be tested
WRN] It looks like all non-ignored mutants are not covered by a test. Go add some tests!
INF] The final mutation score is 0.00 %
```

and the same row of the summary table from each run:

| `RecursiveMath.cs` | % score | # killed | # timeout | # survived | # no cov | # error |
|---|---|---|---|---|---|---|
| coverage analysis **off** — ground truth | **100.00** | 6 | 0 | 4 | **0** | 1 |
| coverage analysis **on** — the default | **0.00** | 0 | 0 | 10 | **7** | 0 |

Seven mutants the test disposes of — six killed outright, one as a runtime error — are
reported as covered by nothing and never run. 

**Expected behavior**
Every copy's flush reaches the coverage data, so both assemblies' mutants are covered and
tested — the same verdicts with coverage analysis on as off.

**Logs**
The reproduction above is deterministic and self-contained, so I've left logs off rather than
attach noise; happy to add `-L` output from either run if that's useful.

**Desktop**
 - OS: Ubuntu 24.04. Nothing in the mechanism is OS-specific — it is a shared-file write
   inside one test host.
 - Type of project: .NET (core) — this repo's own `integrationtest/TargetProjects` MTPSolution
   fixture, plus the added test above
 - Framework Version: .NET SDK 10.0.110
 - Stryker Version: built from source at master `ee06d3a1`, and at `feature/per-test-coverage-mtp`
   `fbf2ed61`

**Additional context**

*Why coverage-analysis-off is the ground truth.* That these mutants are covered is a
source-level fact, not a measurement: `TestFibonacci` calls `RecursiveMath.Fibonacci(5)` by
name and asserts on its output. Coverage analysis is an optimization whose correctness
condition is that it only skips mutants no test executes — so on and off must agree about
which mutants get tested, and the off run (every mutant actually run, no inference) is the
reference.

*On fixing it.* The naming, ownership and lifetime of these temp files has been decided
independently in at least four recent changes — maybe it's time to treat this as a
cross-cutting concern and unify it?

- **#3448** hands the coverage path to the injected helper as a *bare filename* over
  `STRYKER_COVERAGE_FILE`, so the runner and each copy resolve `Path.GetTempPath()` separately
- **#3696** made the coverage file per test host:
  `stryker-coverage-<pid>-<runnerId>-<nonce>-<assembly>-<hash>.txt`
- **#3724** (open) — the mutant-id file is still `stryker-mutant-<runnerId>.txt`: no pid, no
  nonce, so two Stryker processes on one machine collide on their first runner. That PR gives
  it the #3696 treatment
- **#3752** (open) adds two more artifact kinds, both keyed on runner id plus a managed hash
  code: `stryker-coverage-pt-<runnerId>-<assembly>-<GetHashCode()>.txt` and
  `stryker-epoch-<runnerId>-<assembly>-<GetHashCode()>.txt` — no pid, no nonce, and
  `GetHashCode()` is randomised per process, so a leftover file can't be traced back to the run
  that wrote it

Three ideas in case they're useful:

1. **Merge on write.** Each copy appends its line; the runner unions all lines at read. Smallest
   diff, needs a short retry for colliding appends. Fixes this bug and nothing else.
2. **One file per producer** rather than per host — the copy derives its own suffix and the
   runner unions the set. No shared-file writes at all, so torn reads go away too.
3. **A directory per run.** Give each Stryker process one temp *directory* named with its pid
   and a nonce, and put every artifact for that run inside it — coverage, mutant id, and
   whatever #3752 adds. Names inside can then be simple, because the directory supplies the
   isolation; cross-process collisions become impossible by construction; the runner removes
   the directory on exit; and a sweeper at startup can delete directories whose owning pid no
   longer exists, which is the only thing that would ever clean up after a crash. It also gives
   #3752's per-test artifacts somewhere to live without inventing a fourth scheme.

*Follow-up to #3696.* My Reviewer Note there said per-instance coverage files made
merge-on-write unnecessary — true across runner instances, wrong for the copies inside one
host. This is the case I missed.

fabio-dev:
Fixes stryker-mutator/stryker-net#3753

## The bug

With the MTP runner and coverage analysis on, a solution with more than one mutated project loses coverage for every assembly but one. The lost assembly's mutants are reported `NoCoverage` and never tested, even when tests call that code directly.

A test host loads one compiled copy of the injected `MutantControl` per mutated assembly, and every copy resolves the same `STRYKER_COVERAGE_FILE` the runner handed out. `FlushCoverageToFile` overwrites, so the last flush to land replaces the rest and the runner sees only that one.

#3696 made the coverage file per *test host*, which is a different axis: both copies live inside the same host and write that host's single file.

## The fix

This implements option 2 from the issue — one file per producer rather than per host.

**Producer side.** `MutantControl.BuildCoverageFilePath` suffixes the file name it was handed with a discriminator for the assembly the copy was compiled into, so no two copies in a host share a path. The discriminator is the module id, plus the assembly name trimmed to 16 characters to keep the file recognizable while debugging. Identity does not rest on the name: characters a file name cannot hold are replaced, so `A.B` and `A-B` would collide, and trimming makes names sharing a long prefix collide too. A copy that cannot read its own identity still takes a name of its own rather than the one the runner handed out, which the runner treats as its own bookkeeping.

**Reader side.** `MicrosoftTestingPlatformRunner` reads every file whose name starts with the one it handed out and unions their contents. Deletion uses the same enumeration, so stale files from an earlier run cannot be read as this run's coverage.

**Per-test relay.** #3752 landed `perTest` and `perTestInIsolation` while this was open, and both needed the same treatment:

- Each copy now relays through its own epoch file, created synchronously from the static constructor so the runner finds a relay for every assembly whose mutated code ran during the test. The runner requests the flush from every relay and waits for all of them; waiting on a single ack let it read the files while another copy had not flushed, which credited that assembly's coverage to the next test. The set of relays is fixed when the request goes out, so a copy that appears mid-wait is not waited on — it never received the request, and flushes on the next epoch instead.
- The isolated capture clears the coverage files for the assembly before starting its host. A single shared file gave this for free, since every host rewrote it whole; per-assembly files are left untouched by a host that does not load them, and would otherwise be unioned into the next test at `Exact` confidence.

**Lifecycle.** Per-test files are deleted when per-test mode is left, which is where the record naming them still exists and where the servers have just been disposed. They also carry the process id and a per-instance nonce now, like the whole-run coverage files already did: runner ids are small integers handed out per pool slot, so two Stryker processes on one machine used to share these paths for the same assembly.

The injected helper carries no `using` directives by design, so the new code stays fully qualified, and compiles down to C# 2.

## Measured effect

Full mutation run over a 9-project solution, same configuration, only these files differ:

|                        | Without the fix | With the fix |
| ---------------------- | --------------- | ------------ |
| Mutation score         | 39.01 %         | 61.25 %      |
| `NoCoverage`           | 1486            | 567          |
| Mutants actually tested| 962             | 1761         |

Per project, the assemblies higher up the dependency stack are the ones that were losing everything:

| Project        | Without | With   |
| -------------- | ------- | ------ |
| Application    | 15.0 %  | 76.2 % |
| Private.WebApi | 0.0 %   | 65.8 % |
| Infrastructure | 0.0 %   | 21.0 % |
| WebApi.Shared  | 11.2 %  | 48.5 % |

## Tests

**Unit.** `MutantControlCoverageFileTests` compiles the injected helpers into separate assemblies and loads them, reproducing the situation inside one test host. It asserts that each copy writes its own file and that every mutant survives the flush; that copies whose names cannot be told apart by character — a dotted and a dashed name, and two names of 70 characters differing only at the end — still get their own file; and that each copy relays its epoch independently. Cleanup disarms the loaded copies, whose process-exit flush would otherwise rewrite the files after the test deleted them.

`MicrosoftTestingPlatformRunnerCoverageTests` covers the union at read time, the deletion of every per-assembly file, the broadcast reaching every relay and the wait holding until all of them acknowledge, the isolated capture discarding what an earlier test left, the per-test files being deleted when the mode is left, and the per-test paths naming one run and no other.

**Integration.** No new category. `ExtraProject` joins `MicrosoftTestPlatform.slnx` and is referenced by the XUnit test project, with one test covering it; `ExtraProject` in turn calls a new minimal `ExtraLibrary`, so that single test runs mutated code in two assemblies at once. That stack is what makes the fixture a regression guard: were the two mutated projects referenced side by side, no single test would cover both and the copies would never compete for the file. Both are covered by that one test only, so whichever copy loses the race ends up with nothing killed, and the assertion fires whichever way it goes.

Verified by restoring the shared coverage file: `MTPSolution` then fails with `no mutant of Lesson.cs was killed: the coverage of its assembly was lost`, and passes again with the fix. `CheckEveryMutatedProjectIsCovered` runs before the global counts so a regression names the assembly at fault: a global count can stay correct while one assembly loses coverage and another gains some. It requires one killed mutant per file rather than no uncovered ones, since these files also hold code no test exercises.

An earlier revision of this PR added a `MTPMultipleMutatedProjects` category with four projects of its own. It is gone: the integration suite runs on every pull request, and the shape it tested is reachable from projects that already exist. The CI matrix is one category smaller than before this PR.

## Relationship to other open work

#3752 is merged, and the per-test modes it added are covered above. #3724 does the equivalent isolation work for the mutant-id file, which is the one artifact this PR deliberately leaves alone. The two are complementary rather than overlapping: the per-test coverage and epoch files this PR renames did not exist when #3724 was opened — they came with #3752 — and #3724 does not touch them, while this PR does not touch the mutant-id file. Both do introduce the same qualifier in the same constructor, under different names, so whichever lands second will hit a three-line conflict there; the field added here is meant to be shared, so #3724 can reduce to reusing it.

One gap is known and left open: when no relay exists at all, the runner treats the test as covering nothing and reports it with `Normal` confidence. That is right for a test that touched no mutated code, but a copy whose relay file could not be created is indistinguishable from it, and would silently report empty coverage rather than degrading to `Dubious`. Happy to fix it here if you would rather not carry it.


fabio-dev:
FYI this code has been creating using Claude Code

richardwerkman (changes requested):
Thanks for supplying the fix for this. A note: there is no need to add a whole new TargetProject in the integration test. We try to keep the integration test as small and simple as possible since they are ran in every pr gh action run. You can just add the existing `ExtraProject` project to the MTP solution file to recreate this scenario. Maybe just add 1 testcase in the MTP testproject to create coverage on the extra project.

src/Stryker.Core/Stryker.Core/InjectedHelpers/MutantControl.cs
```diff
@@ -211,6 +211,62 @@ public static System.Collections.Generic.IList<int>[] GetCoverageData()
...
+
+            int length = name.Length < 64 ? name.Length : 64;
+            char[] safeName = new char[length];
+            for (int i = 0; i < length; i++)
+            {
+                char current = name[i];
+                safeName[i] = char.IsLetterOrDigit(current) ? current : '-';
+            }
+
+            return new string(safeName);
```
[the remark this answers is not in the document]
fabio-dev: Indeed, fixed here [74687977b59abe198c7a34cef1d1528bebe2b947](https://github.com/stryker-mutator/stryker-net/pull/3769/changes/74687977b59abe198c7a34cef1d1528bebe2b947)

src/Stryker.Core/Stryker.Core.UnitTest/InjectedHelpers/MutantControlCoverageFileTests.cs
```diff
@@ -0,0 +1,151 @@
...
+            syntaxTrees,
+            references,
+            new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary));
+
+        using var peStream = new MemoryStream();
+        var result = compilation.Emit(peStream);
+        result.Success.ShouldBeTrue(
+            $"the injected helpers should compile: {string.Join(Environment.NewLine, result.Diagnostics.Where(diagnostic => diagnostic.Severity == DiagnosticSeverity.Error))}");
+
+        return Assembly.Load(peStream.ToArray());
```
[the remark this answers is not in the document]
fabio-dev: yes, fixed here [9f8297366b16adb634b234793181ef51f6d299f2](https://github.com/stryker-mutator/stryker-net/pull/3769/changes/9f8297366b16adb634b234793181ef51f6d299f2)

fabio-dev:
> Thanks for supplying the fix for this. A note: there is no need to add a whole new TargetProject in the integration test. We try to keep the integration test as small and simple as possible since they are ran in every pr gh action run. You can just add the existing `ExtraProject` project to the MTP solution file to recreate this scenario. Maybe just add 1 testcase in the MTP testproject to create coverage on the extra project.

@richardwerkman  Ok I understand, I removed the SLNX and replaced it by a new project in the existing `MicrosoftTextPlatform.slnx`
Changes here : [a6f3bb1ad7f5fb59ccf873343073dc2fc46c4026](https://github.com/stryker-mutator/stryker-net/pull/3769/changes/a6f3bb1ad7f5fb59ccf873343073dc2fc46c4026)

src/Stryker.TestRunner.MicrosoftTestPlatform/MicrosoftTestingPlatformRunner.cs
```diff
@@ -247,31 +249,87 @@ public void SetPerTestCoverageMode(bool enabled)
...
         }
     }
 
+    /// <summary>
+    /// Deletes the coverage and epoch files of every assembly this runner set up for per-test capture.
+    /// Both come as one file per mutated assembly the host loaded, named after the path handed out, so
+    /// each one is matched by prefix rather than by that exact path.
+    /// Callers hold <see cref="_serverLock"/>.
+    /// </summary>
+    private void DeletePerTestFiles()
```
richardwerkman: This name could be more clear. 
```suggestion
    private void DeletePerTestCoverageFiles()
```

integrationtest/Validation/ValidationProject/ValidateStrykerResults.cs
```diff
@@ -205,8 +207,17 @@ public async Task CSharp_NetCore_MTPSolution()
         // by the MSTest project count as covered (1 survived + 2 timeout), like in the MSTestMTP run.
         // Before coverage files were split per test host, the final flush overwrote the shared
         // file, usually losing exactly those three mutants to NoCoverage.
-        CheckReportMutants(report, total: 670, ignored: 274, survived: 2, killed: 1, timeout: 2, nocoverage: 357);
-        CheckReportTestCounts(report, total: 10);
+        // Teacher.cs and Lesson.cs belong to two mutated projects stacked on each other (ExtraProject
```
richardwerkman: This is starting to become a huge wall of text. This can be explained in 1 sentence as well.

integrationtest/Validation/ValidationProject/ValidateStrykerResults.cs
```diff
@@ -293,6 +304,26 @@ private void CheckMutationKindsValidity(IJsonReport report)
         }
     }
 
+    /// <summary>
+    /// Asserts that every mutated project contributed its own coverage to the report.
+    /// A global count can stay correct while one assembly loses its coverage and another gains some,
+    /// so each source file is checked on its own and the failure message names the assembly at fault.
+    /// </summary>
+    private void CheckEveryMutatedProjectIsCovered(IJsonReport report, params string[] sourceFiles)
```
richardwerkman, line 306: I think this is a good idea. The number of nocoverage mutants should be incorrect if this happens, so it's a bit redundant. But it might be good to assert more specific for which mutants we expect to have coverage or not. However, then I would like to see that in all our integration test scenarios. Might be good to create a new issue for that.

src/Stryker.Core/Stryker.Core/InjectedHelpers/MutantControl.cs
```diff
@@ -241,6 +250,71 @@ public static System.Collections.Generic.IList<int>[] GetCoverageData()
...
+        /// are replaced, so names such as A.B and A-B would collide, and the name has to be trimmed to keep
+        /// the path within length limits, so names sharing a long prefix would collide too. Two colliding
+        /// copies would write to the same file and hide each other's coverage, which is exactly what this
+        /// naming exists to prevent, so the module id - assigned per compiled module - is what separates
+        /// them; the trimmed name is only kept to make the file recognizable while debugging.
+        /// Never returns an empty string: a copy that cannot read its own identity still has to get a name
+        /// of its own rather than fall back to the name the runner handed out, because the runner treats
+        /// that one as its own bookkeeping and never waits for a relay on it.
+        /// </summary>
+        private static string GetOwningAssemblyDiscriminator()
```
richardwerkman, line 275: This method seems overly complex. Can't we simply use a guid as the discriminator? Or at least make this a lot simpler.

richardwerkman (commented):
Just some generic feedback left. I think this can basically be merged but it can be improved a bit more.

src/Stryker.Core/Stryker.Core/InjectedHelpers/MutantControl.cs
fabio-dev: Yes you're right, done here [7cf486bfd65033490d793e1e9bf9a18a496ea444](https://github.com/stryker-mutator/stryker-net/pull/3769/changes/7cf486bfd65033490d793e1e9bf9a18a496ea444)

integrationtest/Validation/ValidationProject/ValidateStrykerResults.cs
fabio-dev: Ok, i simplified it here [350f378dd8eb94fc87c6b6a3ff78934adb313c1e](https://github.com/stryker-mutator/stryker-net/pull/3769/changes/350f378dd8eb94fc87c6b6a3ff78934adb313c1e)

integrationtest/Validation/ValidationProject/ValidateStrykerResults.cs
fabio-dev: yes it's a good idea, i just created this issue : #3797

richardwerkman (approved):

PR #3771, opened 2026-08-11
fix(report): use relative paths in json report to comply with schema

### LINKED ISSUE #3725: JsonReport breaks schema rules on filepaths
**Describe the bug**
At some point we fixed an issue with solution mode reports by using the full filepath of source files as the key for the files array in the JsonReport. This has several consequences:

1. We do not conform to the spec: https://www.unpkg.com/mutation-testing-report-schema@3.8.4/dist/src/mutation-testing-report-schema.json
> Mutated file, with the relative path of the file as the key.
2. Because we don't conform to the spec here, but Baseline does conform to the spec and matches files in the report based on relative path, baseline reports are either completely non-functional or at the very least don't work cross-system.

**Expected behavior**
JsonReport conforms to spec so consumers of the produced report that do follow the spec can use the report.

**Additional context**
Introduced [here ](https://github.com/stryker-mutator/stryker-net/pull/2567) in **2023**

PSR94:
Fixes #3725. The `JsonReport` was using the absolute `FullPath` of source and test files as the keys in its `files` and `testFiles` dictionaries. This broke the mutation-testing-report schema (which mandates relative paths) and broke the baseline feature across systems. This PR calculates paths relative to the `ProjectRoot` (which corresponds to the run's base path or solution directory), effectively restoring cross-platform baseline matching and schema conformance.

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Reporters/Json/JsonReporterTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Reporters/Json/JsonReporterTests.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using Stryker.Utilities;
 using System.IO;
 using System.IO.Abstractions;
 using System.IO.Abstractions.TestingHelpers;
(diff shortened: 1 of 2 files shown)


rouke-broersma:
Already part of https://github.com/stryker-mutator/stryker-net/pull/3722

Please check if work is already being done before opening more pull request

PR #3776, opened 2026-08-12
docs: fix broken star history chart

OctoBored:
The star history chart in the README currently does not render because the service backing it can no longer access the stargazer data it needs. This change moves the chart to an alternative service that uses a different data source, requires no API token, and keeps the chart working as before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,10 +62,10 @@ Issues for the HTML report should be issued at [mutation-testing-elements](https
 
 ## Star History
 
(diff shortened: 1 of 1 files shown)


rouke-broersma:
Seems like star-history.com has a workaround until they get back access to the data. I prefer implementing the workaround or removing the star chart instead of migrating to a fork of start-history hosted by some unknown other party that seems to give out subdomains to random open source projects.

PR #3785, opened 2026-08-17
Quick fix for the issue with duplicate test projects on master

dupdob:
This PR filters out any quasi-duplicate test project for a given mutable candidate. This can happen with multi-targeting test projects when two or more targets refer to the same target for the mutable projects. For illustration, Polly.Specs net8, net9 and net10 targets all refer to Polly net6

### DIFF
--- a/integration-tests.ps1
+++ b/integration-tests.ps1
@@ -51,9 +51,9 @@ function Run-Stryker {
   Push-Location $WorkingDirectory
   try {
     if ($effectiveArguments.Count -gt 0) {
-      & $stryker @effectiveArguments
+      & $stryker @effectiveArguments --diag
     } else {
-      & $stryker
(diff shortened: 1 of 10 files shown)


richardwerkman:
I think it would be good to update our integration test project to include this scenario and verify only 1 is built. That way we can prevent this issue from poping up again in the future.

dupdob:
This is a quick fix. I plan to add relevant unit tests and modify one of the integration test project for the full fix, but this is not probably a few weeks away

richardwerkman:
Integration testing should be fairly simple. Just add different targets to one test project and verify the number of tests found doesn't change. I get that unit testing is a bit more works so I'm fine with delaying that.

richardwerkman:
Thanks for the quick fix!

PR #3788, opened 2026-08-18
fix(rollback): attribute compile errors caused by mutations erasing throw statements

### LINKED ISSUE #3783: CSharpRollbackProcess crashes with "Internal error due to compile error" on a large project
# Title

[stryker-repro.zip](https://github.com/user-attachments/files/31090137/stryker-repro.zip)

; minimal repro included for the underlying unattributable-mutation mechanism (CS0165/CS0161)

# Labels

🐛 Bug

---

## Describe the bug

On a large .NET 8 class library (~600 files / ~94K lines), every Stryker run aborts after ~3.5 minutes during mutant compilation with:

```
[FTL] Stryker.NET could not compile the project after mutation. This is probably an error for Stryker.NET and not your project. Please report this issue on github with the previous error message.
[ERR] An error occurred during the mutation test run
System.AggregateException: One or more errors occurred. (Internal error due to compile error.)
 ---> Stryker.Abstractions.Exceptions.CompilationException: Internal error due to compile error.
   at Stryker.Core.Compiling.CSharpRollbackProcess.RollbackMutationsInError(ICompilationContent wrapper, ImmutableArray`1 diagnostics, Mode mode, Boolean devMode) in /_/src/Stryker.Core/Stryker.Core/Compiling/CSharpRollbackProcess.cs:line 82
   at Stryker.Core.Compiling.CsharpCompilingProcess.TryCompilation(Stream ms, Stream symbolStream, EmitResult previousEmitResult, Mode mode, Int32 retryCount) in /_/src/Stryker.Core/Stryker.Core/Compiling/CsharpCompilingProcess.cs:line 184
   at Stryker.Core.Compiling.CsharpCompilingProcess.Compile(Stream ilStream, Stream symbolStream) in /_/src/Stryker.Core/Stryker.Core/Compiling/CsharpCompilingProcess.cs:line 95
   at Stryker.Core.MutationTest.CsharpMutationProcess.CompileMutations(MutationTestInput input, CsharpCompilingProcess compilingProcess) in /_/src/Stryker.Core/Stryker.Core/MutationTest/CsharpMutationProcess.cs:line 72
   at Stryker.Core.MutationTest.CsharpMutationProcess.Mutate(MutationTestInput input, IStrykerOptions options) in /_/src/Stryker.Core/Stryker.Core/MutationTest/CsharpMutationProcess.cs:line 62
   at Stryker.Core.MutationTest.MutationTestProcess.Mutate() in /_/src/Stryker.Core/Stryker.Core/MutationTest/MutationTestProcess.cs:line 65
   [...]
```

The crash is preceded by a long series of "Safe Mode" rollbacks for mutations Stryker itself cannot attribute (anonymized paths; ~22 such warnings across ~12 files per run):

```
[WRN] An unidentified mutation in <redacted>/src/Core/Security/AccessExtensions.cs resulted in a compile error (at 163:114) with id: CS0165, message: Use of unassigned local variable 'primaryRights' (Source code: <enum member access>)
[INF] Safe Mode! Stryker will remove all mutations in ResolveDetailAccess and mark them as 'compile error'.
[WRN] An unidentified mutation in <redacted>/src/Core/Data/ConnectionBase.cs resulted in a compile error (at 606:14) with id: CS0165, message: Use of unassigned local variable 'result' (Source code: result)
[INF] Safe Mode! Stryker will remove all mutations in ExecuteReader and mark them as 'compile error'.
[WRN] An unidentified mutation in <redacted>/src/Core/Dispatch/MethodResolver.cs resulted in a compile error (at 274:31) with id: CS0161, message: 'MethodResolver.FailForMissingKey(...)': not all code paths return a value (Source code: internal static MethodInfo FailForMissingKey(...) {{ ...
[INF] Safe Mode! Stryker will remove all mutations in FailForMissingKey and mark them as 'compile error'.
```

This looks like the same family as #3542 (fixed in 4.14.2), but it still occurs on **4.16.0** (latest at time of writing).

## Minimal reproduction of the underlying mechanism

I could not reproduce the terminal crash in a small project — at small scale Safe Mode always recovers — but the *mechanism* that precedes it (mutations producing CS0165/CS0161 that Stryker cannot attribute, triggering "unidentified mutation" + method-wide Safe Mode rollbacks) reproduces reliably with the three code shapes below. In a tiny 5-file project, these shapes alone cause **34 of 85 generated mutants (40%) to be discarded as CompileError**. On our real codebase the same shapes appear at much higher density, and the run ends in the `RollbackMutationsInError` crash instead of recovering.

All three shapes share one property: **the compile error surfaces at a different source location than the mutated node** (the use-site of a local, or the method signature line), which appears to be why the diagnostic→mutant attribution fails.

**Shape 1 — C# 12 collection expressions assigned to locals across if/else chains** (definite assignment depends on every branch; a mutated branch breaks CS0165 at the later use-site):

```csharp
public static (AccessArea[][], AccessArea[][]) ResolveRights(int kind, bool canEdit) {
  AccessArea[][] primaryRights;
  AccessArea[][] secondaryRights;
  if (kind == 1) {
    if (canEdit) {
      primaryRights = [[AccessArea.ViewAlpha, AccessArea.EditAlpha]];
      secondaryRights = [[AccessArea.ViewBeta, AccessArea.EditBeta]];
    } else {
      primaryRights = [[AccessArea.ViewAlpha]];
      secondaryRights = [[AccessArea.ViewBeta]];
    }
  } else {
    primaryRights = [[AccessArea.ViewAlpha]];
    secondaryRights = [[AccessArea.ViewBeta]];
  }
  return (primaryRights, secondaryRights);  // CS0165 reported here when a branch assignment is mutated away
}
```

**Shape 2 — try/catch assignment then use** (statement mutation of the assignment → CS0165 at the use-site):

```csharp
IList<string> transformed;
try {
  transformed = upperInputs.Select(i => i + "!").ToList();
} catch (Exception e) {
  throw new InvalidOperationException("Failed to transform", e);
}
if (transformed.Count != upperInputs.Length) { ... }  // CS0165 reported here
```

**Shape 3 — method whose terminal statement is a `throw` preceded by another statement** (statement mutation wrapping the throw → CS0161 reported at the method signature):

```csharp
public static int FailForMissingKey(string key) {   // CS0161 reported at this line
  _log.Add(key);
  throw new KeyNotFoundException($"No handler matched {key}.");
}
```

A complete runnable solution (2 projects, 5 source files, 12 xunit tests, plain `dotnet new` scaffolding + these shapes) is attached / can be pasted on request — running `dotnet stryker` on it shows the `unidentified mutation` + `Safe Mode!` warnings and the 40% CompileError mutant wastage. (At this scale the run still completes with a score; the crash needs larger volume.)

## Expected behavior

Either (a) mutations that provably break definite-assignment/reachability are not generated (or are attributed correctly and rolled back individually), or (b) Safe Mode rollback recovers fully instead of eventually throwing `CompilationException` from `RollbackMutationsInError` on large inputs.

## What we tried (none of it avoids the crash on the real project)

- `--mutation-level Basic` — crashes identically (~3.5 min).
- `"ignore-mutations": ["Block"]` and `["Block", "Statement"]` — reduces the number of compile-error mutations but still crashes.
- `"mutate"` exclusion patterns (`"!path/to/each/offending/file.cs"`) for every file named in the warnings — **the excluded files were still instrumented and still produced the same compile-error warnings**, which suggests `mutate` filtering is applied after mutation/compilation rather than before instrumentation. (Happy to file that separately if it's unexpected.)

## Environment

- Stryker.NET: 4.16.0 (dotnet tool, local manifest)
- SDK: .NET 10.0.203, target framework `net8.0`
- OS: macOS (darwin, arm64)
- Project under test: class library, ~600 `.cs` files / ~94K lines, `TreatWarningsAsErrors=true`, referenced by an ASP.NET host and a ~4,000-test xunit project; `coverage-analysis: perTest`
- Config: `project` + `test-projects` + `mutate` glob scoped to a 21-file subdirectory (note: instrumentation still covers the whole project regardless of the `mutate` glob)

Full `--diag` log available on request (needs light redaction before sharing).

nateware:
Fixes #3783

## Problem

Mutations erasing a `throw` statement were not attributable when the resulting compile error surfaced at a different source location than the mutated node. The rollback process then fell into safe mode and marked **every** mutation in the enclosing method as a compile error; on large projects the resulting rollback churn ended in the `CompilationException` from `RollbackMutationsInError` reported in the issue.

Two attribution gaps, both in `CSharpRollbackProcess.IdentifyMutationsAndFlagForRollback`:

- **CS0161** (*not all code paths return a value*): when a method's reachability relies on a terminal `throw`, a statement/block mutation erasing that throw triggers CS0161 at the method declaration. The handler only scanned for mutations erasing a `ReturnStatementSyntax`, found nothing, and safe mode discarded the whole method (issue repro shape 3: 4 of 4 mutants lost).
- **CS0165/CS0177** (*use of unassigned local variable*): a `catch` clause that no longer throws lets execution continue past the try/catch without the variable being assigned, reporting CS0165 at the use site. The handler only scanned for mutations erasing an assignment, so once the genuine assignment-erasing mutations were rolled back, the remaining CS0165 was unattributable and safe mode discarded the whole method (issue repro shape 2: 11 of 11 mutants lost, including unrelated string/unary mutations).

Shape 1 from the issue (block mutations erasing collection expression assignments) turned out to already be attributed correctly — those mutants genuinely cannot compile and are rolled back individually without safe mode. A regression test now pins that behavior.

## Fix

Extend both erasing scans to also consider mutations erasing a `ThrowStatementSyntax`:

- CS0161 scans for mutations erasing a `return` **or a `throw`**.
- CS0165/CS0177 first scans for assignment-erasing mutations (unchanged), and when none is found, makes a second pass scanning for throw-erasing mutations.

The guilty mutation is now rolled back individually and unrelated mutations in the same method survive. With the fix, the issue's shape 2 goes from 11/11 mutants lost to 4 genuine compile errors (body/try/catch block mutations and the throw statement mutation), and shape 3 from 4/4 lost to 2 genuine compile errors — with no safe-mode warnings.

## Tests

- `CSharpRollbackProcessTests`: two new unit tests (`ShouldRollbackMutationErasingTerminalThrow`, `ShouldRollbackMutationErasingThrowInCatchClause`) driving real Roslyn diagnostics through the rollback process; each includes an unrelated mutation in the same method that must survive, so the tests fail if safe mode is (wrongly) engaged.
- `CSharpCompilingProcessTests`: three new end-to-end tests running the full mutate → compile → rollback pipeline over the three code shapes from the issue.

All 1550 tests in `Stryker.Core.UnitTest` pass.

## Validation on the original codebase

A locally packed `dotnet-stryker` with this fix was run against the ~600-file / ~94K-line project that originally crashed: the run now completes with a mutation score instead of aborting after ~3.5 minutes, and the `"unidentified mutation"` / `"Safe Mode!"` warnings for the CS0165/CS0161 shapes are gone.

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Compiling/CSharpCompilingProcessTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Compiling/CSharpCompilingProcessTests.cs
@@ -520,6 +520,134 @@ public class Calculator
         projectContentsMutants.Count(t => t.ResultStatus == MutantStatus.Pending).ShouldBe(3);
     }
 
+    [TestMethod]
+    public void ShouldCompileAndRollbackErrorWhenMutationErasesTerminalThrow()
+    {
+        // issue #3783: erasing the terminal throw of a non-void method causes CS0161 at the method
+        // declaration. Only the mutations erasing the throw should be rolled back; the others must survive.
+        var sourceFile = """
+            using System;
+            using System.Collections.Generic;
+
+            namespace ExampleProject;
+
+            public static class MethodResolver
+            {
+                private static readonly List<string> _log = new();
+
+                public static int FailForMissingKey(string key)
+                {
+                    _log.Add(key);
+                    throw new KeyNotFoundException($"No handler matched {key}.");
+                }
+            }
+            """;
+        var projectContentsMutants = MutateAndCompileSource(sourceFile).ToList();
+        // those results can change if mutators are added.
+        // the block mutation on the method body and the statement mutation on the throw are genuine compile errors
+        projectContentsMutants.Count(t => t.ResultStatus == MutantStatus.CompileError).ShouldBe(2);
+        // the statement mutation on _log.Add and the string mutation must survive
+        projectContentsMutants.Count(t => t.ResultStatus == MutantStatus.Pending).ShouldBe(2);
+        projectContentsMutants.Where(t => t.Mutation.Type == Mutator.String)
+            .ShouldAllBe(t => t.ResultStatus == MutantStatus.Pending);
+    }
+
+    [TestMethod]
+    public void ShouldCompileAndRollbackErrorWhenMutationErasesThrowInCatchClause()
+    {
+        // issue #3783: erasing the throw of a catch clause lets the catch complete normally, breaking
+        // definite assignment (CS0165) at a use site after the try/catch. Only the mutations erasing
+        // the assignment or the throw should be rolled back; the others must survive.
+        var sourceFile = """
+            using System;
+            using System.Collections.Generic;
+
+            namespace ExampleProject;
+
+            public static class Transformer
+            {
+                public static int Transform(string[] upperInputs)
+                {
+                    string[] transformed;
+                    try
+                    {
+                        transformed = (string[])upperInputs.Clone();
+                    }
+                    catch (Exception e)
+                    {
+                        throw new InvalidOperationException("Failed to transform", e);
+                    }
+                    if (transformed.Length != upperInputs.Length)
+                    {
+                        return -1;
+                    }
+                    return transformed.Length;
+                }
+            }
+            """;
+        var projectContentsMutants = MutateAndCompileSource(sourceFile).ToList();
+        // those results can change if mutators are added.
+        // block mutations on the method body, try block and catch block plus the statement mutation
+        // on the throw are genuine compile errors
+        projectContentsMutants.Count(t => t.ResultStatus == MutantStatus.CompileError).ShouldBe(4);
+        // the string, equality, unary and 'return -1' block mutations must survive
+        projectContentsMutants.Count(t => t.ResultStatus == MutantStatus.Pending).ShouldBe(4);
+        projectContentsMutants.Where(t => t.Mutation.Type == Mutator.String || t.Mutation.Type == Mutator.Equality)
+            .ShouldAllBe(t => t.ResultStatus == MutantStatus.Pending);
+    }
+
+    [TestMethod]
+    public void ShouldCompileAndRollbackErrorWhenCollectionExpressionAssignmentsAreErased()
+    {
+        // issue #3783: block mutations erasing collection expression assignments across if/else chains
+        // break definite assignment (CS0165) at the use site. They must be attributed and rolled back
+        // individually, without safe mode discarding the collection expression mutations.
+        var sourceFile = """
+            using System;
+
+            namespace ExampleProject;
+
+            public enum AccessArea { ViewAlpha, EditAlpha, ViewBeta, EditBeta }
+
+            public static class AccessExtensions
+            {
+                public static (AccessArea[][], AccessArea[][]) ResolveRights(int kind, bool canEdit)
+                {
+                    AccessArea[][] primaryRights;
+                    AccessArea[][] secondaryRights;
+                    if (kind == 1)
+                    {
+                        if (canEdit)
+                        {
+                            primaryRights = [[AccessArea.ViewAlpha, AccessArea.EditAlpha]];
+                            secondaryRights = [[AccessArea.ViewBeta, AccessArea.EditBeta]];
+                        }
+                        else
+                        {
+                            primaryRights = [[AccessArea.ViewAlpha]];
+                            secondaryRights = [[AccessArea.ViewBeta]];
+                        }
+                    }
+                    else
+                    {
+                        primaryRights = [[AccessArea.ViewAlpha]];
+                        secondaryRights = [[AccessArea.ViewBeta]];
+                    }
+                    return (primaryRights, secondaryRights);
+                }
+            }
+            """;
+        var projectContentsMutants = MutateAndCompileSource(sourceFile).ToList();
+        // those results can change if mutators are added.
+        // only the block mutations erasing branch assignments are genuine compile errors
+        projectContentsMutants.Where(t => t.ResultStatus == MutantStatus.CompileError)
+            .ShouldAllBe(t => t.Mutation.Type == Mutator.Block);
+        projectContentsMutants.Count(t => t.ResultStatus == MutantStatus.CompileError).ShouldBe(4);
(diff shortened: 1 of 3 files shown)


nateware:
The MTPSolution failure doesn't look related to this PR. Same branch passed that workflow at 958ec6a0, then failed on 982c32a2 — and the only difference between those two is the merge of master with the #3786 analyzers bump, none of my code changed.

Also worth noting the mutant counts all came out exactly as expected in the failing run (670/274/2/1/2/357) — it only tripped on `CheckReportTestCounts`, which got 8 tests instead of 10. Other PRs are hitting the same assertion right now, e.g. run 32105655273 (a renovate branch) failed `MultipleTestProjects` on windows with 10 instead of 21. Both are multi-test-host targets, so this smells like the coverage file race that's mentioned in the comment above the MTPSolution assertions in ValidateStrykerResults.cs.

I've pushed an empty commit to retrigger.

src/Stryker.Core/Stryker.Core.UnitTest/Compiling/CSharpCompilingProcessTests.cs
```diff
@@ -520,6 +520,134 @@ void TestMethod(){
...
+        // on the throw are genuine compile errors
+        projectContentsMutants.Count(t => t.ResultStatus == MutantStatus.CompileError).ShouldBe(4);
+        // the string, equality, unary and 'return -1' block mutations must survive
+        projectContentsMutants.Count(t => t.ResultStatus == MutantStatus.Pending).ShouldBe(4);
+        projectContentsMutants.Where(t => t.Mutation.Type == Mutator.String || t.Mutation.Type == Mutator.Equality)
+            .ShouldAllBe(t => t.ResultStatus == MutantStatus.Pending);
+    }
+
+    [TestMethod]
+    public void ShouldCompileAndRollbackErrorWhenCollectionExpressionAssignmentsAreErased()
```
dupdob: This already passes without the proposed change/does not cover

dupdob:
Fixed via #3796

PR #3800, opened 2026-08-24
Complete implementation of analyzer configuration files

### LINKED ISSUE #3698: Stryker 4.16.0 loses Microsoft.Windows.CsWin32 generated symbols during mutation compilation
## Describe the bug

A minimal `net8.0-windows` project using `Microsoft.Windows.CsWin32` 0.3.106 builds successfully and its test passes, but Stryker.NET 4.16.0 fails its internal mutation compilation even when only an unrelated source file is selected for mutation.

Stryker correctly parses the mutate glob as `**/Target.cs` and creates exactly one mutant (`+` to `-`) in that file. During compilation it then fails to resolve `Windows.Win32`, `HWND`, and `PInvoke` used by the unmutated `Win32Anchor.cs`. Rollback reports diagnostics from that file as an unidentified mutation and aborts before producing the JSON report.

The failure reproduces without WPF, DS4Windows, a wrapper tool, or a dirty Git checkout.

## Logs

Minimal repro, exact commands, and the complete `--log-to-file` trace:

https://gist.github.com/anagnorisis2peripeteia/c36bcc954982d87480d57027cdb1734c

The trace shows:

- one mutant created, only in `Target.cs`
- `CS0234`: `Windows.Win32` is missing
- `CS0246`: `HWND` is missing
- `CS0103`: `PInvoke` is missing
- fatal `Stryker.Core.Exceptions.CompilationException`

## Expected behavior

Stryker should preserve or regenerate the CsWin32 source-generator output during mutation compilation and finish testing the scoped `Target.cs` mutant. References to generated symbols in an unrelated, unmutated source file should not abort the run.

## Desktop

- OS: Windows 11 Pro 10.0.26200 (build 26200)
- .NET SDK: 8.0.420 (selected by the repro's `global.json`)
- Target framework: `net8.0-windows10.0.19041.0`
- Test framework: MSTest
- Stryker.NET: 4.16.0
- Microsoft.Windows.CsWin32: 0.3.106

## Additional context

The repro contains seven hand-authored project/source files. The unmutated test passes (1/1), and the Stryker failure occurs in about 15 seconds.

#3605 and #3611 addressed source-generator support in 4.15.0. This is a remaining analyzer-config propagation case in 4.16.0.

### Where it breaks (now reproduced end-to-end — see comments)

A follow-up `--diag` run and a Buildalyzer 9.0.0 probe show:

- Buildalyzer captures `NativeMethods.txt` in `IAnalyzerResult.AdditionalFiles`.
- Buildalyzer captures the generated MSBuild editor-config path in `IAnalyzerResult.CompilerArguments` as `/analyzerconfig:obj\Release\net8.0-windows10.0.19041.0\CsWin32Repro.GeneratedMSBuildEditorConfig.editorconfig`.
- That generated file contains the non-empty `build_property.CsWin32InputMetadataPaths` value required by CsWin32.
- `IAnalyzerResult.Properties` contains only `PkgMicrosoft_Windows_CsWin32`; it does not contain `CsWin32InputMetadataPaths` because that property is assembled by an MSBuild target immediately before the editor-config file is generated.

Stryker's `SimpleAnalyserConfigOptionsProvider` currently constructs global options only from `IAnalyzerResult.Properties` and returns empty per-file options. It does not consume the analyzer-config paths already present in the captured compiler arguments. CsWin32 therefore receives the input text but no metadata paths, produces no generated trees, and the mutated compilation fails on the generated symbols.

Looks like the captured analyzer configs can be consumed via the public `AnalyzerConfig`/`AnalyzerConfigSet` APIs — no new parser, no Buildalyzer change — but that's your call. Happy to PR it (before/after in the comments).

dupdob:
Properly discover and read config files so code generators and analyzers can get required information when needed.

based on PR #3700 from @anagnorisis2peripeteia with streamlined design.

Should fix #3698


richardwerkman:
Can we add an integration scenario for this? So we can be sure this is fixed properly and stays fixed in the future

anagnorisis2peripeteia:
> Can we add an integration scenario for this? So we can be sure this is fixed properly and stays fixed in the future

#3698 was the scenario that brought this up

dupdob:
@richardwerkman : this is not a regression, I made a crude implementation a few years ago due to lack of available documentation at the time and which was not respecting the contract. This PR contains a correct implementation as well as unit tests.
I am not able to test with CsWin32 as I work on MacOS. And it would be better not to have a Windows specific integration test.

PR #3805, opened 2026-08-25
Improve relative path resolution for dotnet with msbuild

### LINKED ISSUE #3804: DirectoryNotFoundException with F# test project in a subdirectory
**Description**

When running `dotnet stryker --solution <sln>` against a solution where an F# project lives in a subdirectory of the solution root, Stryker crashes with an unhandled `System.IO.DirectoryNotFoundException`.

Stryker resolves the F# test project's source file paths relative to the current working directory (most likely solution root), instead of relative to the F# project's own directory. For a F# project nested in a subfolder, the resulting path is wrong and the file cannot be found.

**Expected behavior**

Allow nested F# projects.

**Desktop**
  
 - Windows, dotnet v10.0.111
 - Stryker 4.16.0

**Additional context**

Solution with a minimal repro: https://github.com/bmordue/stryker-fsharp-repro

[log-20260825.txt](https://github.com/user-attachments/files/31416688/log-20260825.txt)

bmordue:
Pass DOTNET_HOST_PATH explicitly when invoking MSBuild through the .NET SDK.

- Use an existing DOTNET_HOST_PATH environment variable, otherwise resolve the host executable from the current runtime location.
- Resolve relative source-file paths returned by Buildalyzer relative to the test project directory.
- Preserve absolute source paths and use the resolved paths when parsing and storing test files.

Adds regression coverage for SDK-hosted MSBuild and relative source-file resolution.

Fixes #3804

### DIFF
--- a/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialBuildProcessTests.cs
+++ b/src/Stryker.Core/Stryker.Core.UnitTest/Initialisation/InitialBuildProcessTests.cs
@@ -133,6 +133,24 @@ public class InitialBuildProcessTests : TestBase
             Times.Once);
     }
 
+            [TestMethodWithIgnoreIfSupport]
+            [IgnoreIf(nameof(Is.NotWindows))]
+            public void InitialBuildProcess_WithSdkMsBuild_SetsDotnetHostPath()
+            {
+            var processMock = new Mock<IProcessExecutor>(MockBehavior.Strict);
+            processMock.SetupProcessMockToReturn("");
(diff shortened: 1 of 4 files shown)


dupdob:
thanks for this PR.
Quick question: what problem does setting DOTNET_HOST_PATH resolve ? My understanding is that this variable is set by MSBuild and should be used by tools invoked during the build process, not the other way around. 

Is there any relationship with the path resolution problem?

bmordue:
This fix was enough to resolve my issue with F# projects, as described in the bug.

bmordue:
I believe this has now been superseded by #3809

PR #3831, opened 2026-09-11
chore: Update migration guide with new output path details

richardwerkman:
Add note about disk baseline output path changes in Stryker v5.


rouke-broersma:
discussed that this isn't needed as there aren't any migration steps and release notes will already mention this change

PR #3839, opened 2026-09-13
feat(compilation): follow project C# language version

### LINKED ISSUE #3662: Improve C# version handling
Stryker interpret the `latest` language version settings in the context of its internal compiler. It will result in failed run for projects that relies on syntax contract that are broken by the latest C# version (see issue #3490 for an example)

Stryker should interpret this setting within the context of the target framework, that is, the latest C1 version supported for said framework

Alternatively, user can change this settings to the appropriate language version manually, or remove it altogether as automatic detection has been fixed in V4.15. Nonetheless this remains cumbersome. Furthermore, if the `latest` setting work today for projects, it may break with a future release.

**Additional context**
Important addition: current versions of Stryker do not handle C# version related issues gracefully, are they are typically reported as mutation related error, with unclear error messages that impair problem analysis

Evangelink:
## Summary

- use the `LangVersion` reported by each analyzed project for source, test, and source-generator parsing
- leave Roslyn language-version resolution untouched when the project does not report a value
- retain `language-version` as a deprecated, validated no-op with a warning, and omit it from newly generated configs
- update configuration documentation and add focused regression coverage

## Validation

- `dotnet build .\src\Stryker.slnx --no-restore --verbosity minimal`
- focused Stryker.Core tests: 55 passed
- focused Stryker.CLI test: 1 passed

Fixes #3662


dupdob:
~~I am late to the party, but it looks to me like this change actually creates issues instead of fixing them.
If understand this correctly, code parsing now defaults to latest C# version instead of using the parameter. This impacts  backup discovery mode, which triggers when there is no csproj IIRC.~~
~~For me, the proper fix is to either keep the parameter for backup discovery mode or decide to get rid of it altogether 
(since I am not sure if it is still relevant)~~

Edit: I realized I was completely wrong here, my bad

rouke-broersma:
The parameter is now no-op/ logs a message indicating it's not used so we can fully remove it in the next breaking change. Instead the langversion from the project is used if set (discovered from buildalyzer) and if not set by the user we don't pass any language version to roslyn at which point roslyn as far as the docs state should auto determine the correct langversion for the target framework same as a dotnet build would. 

This is also how we discussed this in the issue was it not?

Unless I misread the implementation, latest is no longer used anywhere.

dupdob:
yeah, you're right. I even got the backup strategy logic wrong.
This is because this is not what I had in mind when discussing this 😄 . My idea was to properly align `latest` to the framework version, not getting rid of the parameter altogether. 
Definitely a smarter move

Evangelink:
You got me worried for a bit but hopefully I am reading the comments late enough :)
Thanks for raising the concerns!

Issue #2381, opened 2023-01-15
Unity support


codmw44:
### Is your feature request related to a problem? Please describe.
Add support of Unity 3D. It has C# code but do not use the whole infrastructure of .NET

### Describe the solution you'd like
Add special runner for unity. Unity doesn't support concurrent test running.
Add special detector of test csproj as Unity uses its own UnityEditor.TestRunner
Add able to execute Stryker in folder that contains .sln and csproj at one place
Add able to rerun unity tests without closing in batchmode (should save many time by overhead of opening Unity)

### Describe alternatives you've considered
I have concerns about cecil. Will it works with Unity DLLs? by csproj it has the signature of .NET Framework 4.7. 
I made some tests and have warnings with them
```[10:57:56 INF] Stryker will use a max of 1 parallel testsessions.
[10:57:56 WRN] Stryker is running in single threaded mode due to concurrency being set to 1.
[10:57:56 INF] Identifying project to mutate.
[10:57:57 INF] The project /Users/andrewv/Desktop/DummyProject/Common.ExtensionPoints.PublicAPI.csproj will be mutated.
[10:57:58 INF] Analysis complete.
[10:57:58 INF] Building test project /Users/andrewv/Desktop/DummyProject/Common.ExtensionPoints.Tests.csproj (1/1)
[10:58:13 INF] Total number of tests found: 12.
[10:58:13 INF] Initial testrun started.
[10:58:27 WRN] Original project under test /Users/andrewv/Desktop/DummyProject/Library/ScriptAssemblies/Common.ExtensionPoints.PublicAPI.dll could not be loaded. 
Embedded Resources might be missing.
System.IO.FileNotFoundException: Could not find file '/Users/andrewv/Desktop/DummyProject/Library/ScriptAssemblies/Common.ExtensionPoints.PublicAPI.dll'.
File name: '/Users/andrewv/Desktop/DummyProject/Library/ScriptAssemblies/Common.ExtensionPoints.PublicAPI.dll'
   at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirectory, Func`2 errorRewriter)
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String path, OpenFlags flags, Int32 mode)
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize)
[...]

Issue #3421, opened 2026-02-03
MTP test runner shows errors when running on .NET Framework test project


richardwerkman:
> I tried mutating .NET Framework 4.7.2 MSTest unit tests but got various "Stryker failed to test 1 mutant(s)." errors.  This is my first time doing this, so I'm not sure if the errors are due to my mistakes or the new MTP support.  I'm running from the source directory, referencing the solution and test-projects in stryker-config.json plus I included "coverage-analysis": "off" as well.
> 
> The command line I used was: dotnet stryker --testrunner mtp -o
> 
> [stryker_log.txt](https://github.com/user-attachments/files/25024474/stryker_log.txt) 

 _Originally posted by @lushess255 in [#3094](https://github.com/stryker-mutator/stryker-net/issues/3094#issuecomment-3837320488)_

Add an integration test for this scenario and provide a fix.

PR #3840, opened 2026-09-13
feat(unity): fail fast for unsupported test projects

Evangelink:
## Summary

- recognize Unity-generated test projects from the documented `nunit.framework` and `UnityEngine.TestRunner` references
- fail before project build and test execution with a clear Unity-specific message
- explain that a dedicated Unity runner is required because VsTest and Microsoft Testing Platform cannot execute Unity tests
- cover detection and fail-fast behavior with focused unit tests

## Scope

This is a small prerequisite for #2381, not Unity runner support. It prevents Unity projects from reaching later test-discovery failures or receiving misleading missing-adapter guidance.

The earlier Unity fixture and cross-platform integration category were removed after maintainer feedback. The PR now changes only project analysis and its unit tests; it adds no Unity dependencies, sample project, integration workflow, or CI matrix entries.

Progresses #2381.

## Validation

- `AnalyzerResultExtensionsTests` and `InputFileResolverTests`: 128 passed
- effective PR diff: four files, with no integration-test surface changes


rouke-broersma:
Are we actually able to executed unity nunit unit tests after correctly recognizing them or does recognizing them decrease clarity because we now crash later on in the process? If so I would prefer recognizing them and failing fast with a message that we do not support unity unit testing.

Evangelink:
Let me build a full e2e case to fully validate where we are (I stopped at the ticket request).

Evangelink:
@rouke-broersma Good call — the E2E confirms we cannot execute Unity Test Framework tests through the existing runners yet. In e8658d21 the packaged CLI now analyzes a Unity-shaped EditMode project and fails immediately with a Unity-specific unsupported message, before build or test execution, instead of suggesting a missing VsTest adapter. The new `UnityUnsupported` integration category runs this contract across the hosted OS matrix.

rouke-broersma (dismissed):
While I appreciate the effort of adding integration tests, I think in this case an integration test to keep confirming that we _don't_ support something is a bit much to me. At best it will do nothing, at worst it will break due to dependencies or other unity specific things and cost us maintenance work.

rouke-broersma:
@Evangelink did you intend to close the pull request? Maybe I was not clear enough in my feedback. It would be nice to fail-fast if unity unit tests are detected, I just don't think a complete integration test confirming this behavior is necessary.

Evangelink:
Oh right sorry and thanks for the message. I thought you said "now that we know it doesn't work let's do nothing". I'll clean up and reopen tomorrow.

Side note, it's hackathon week at MS so I can spend a few tokens on Stryker if there are some things you want me to look into.

rouke-broersma:
@Evangelink if you want something fun to work on you could investigate implementing mutation server protocol in stryker.NET: https://github.com/stryker-mutator/editor-plugins/tree/main/packages/mutation-server-protocol

I don't think we created a proper issue explaining what it is and how we would want to integrate, I can provice more details if you're interested. But the protocol I think is pretty self-explanatory. The idea is to add something like a `serve` command which starts a jsonrpc server that listens for mutation testing commands and invokes Stryker.Core. We can then invoke stryker from IDE's like vscode using the test explorer. It is already implemented for stryker-js.

Evangelink:
@rouke-broersma Thanks for clarifying — I reopened the PR and removed the full Unity fixture, integration category, and CI matrix changes in 3d91d1fc. The PR now keeps only Unity project recognition, the actionable fail-fast message, and focused unit coverage (`AnalyzerResultExtensionsTests` and `InputFileResolverTests`, 128 passing locally).

richardwerkman:
@Evangelink If you still have some tokens to spend, we could use some help to look into https://github.com/stryker-mutator/stryker-net/issues/3421 :)

Evangelink:
Sure assign me to it. I'll have something on the afternoon

PR #3841, opened 2026-09-13
fix(mtp): isolate mutations when mixing is disabled

### LINKED ISSUE #3742: mtp runner lacks on option to test each mutant in its own process
**Describe the bug**

With `"test-runner": "mtp"`, mutants that are *provably equivalent* (no test can possibly detect them) are intermittently reported **Killed** when the code-under-test holds **process-global state** — e.g. any CSLA `BusinessBase`, whose validation rules are registered once per process and kept in a static cache.

Two independent levers each drive the false-kills to **zero**:

- running the same solution at **`--concurrency 1`**, and
- removing the process-global state — an otherwise-identical **stateless** codebase never misattributes, even at ~8,600 mutants / 17 min.

This looks like a per-mutant **isolation** problem: the `mtp` shared test-server is reused across *concurrently-evaluated* mutants, so a genuine failure produced by one mutant is attributed to another — here, an equivalent one → false "Killed". Consequences: the mutation **score is inflated** (equivalent mutants counted as detected) and the per-mutant survivor set is **not reproducible** between identical runs.

**Minimal repro:** https://github.com/gem4511/stryker-mtp-isolation-repro — two runnable solutions (a CSLA case and a stateless control):

| solution | process-global state | concurrency | provably-equivalent "canary" mutants reported **Killed** |
|---|---|---|---|
| CSLA (100 business objects) | CSLA per-type rule cache | default | **~9–18 of 200** (varies run-to-run) |
| CSLA | same | **`--concurrency 1`** | **0 of 200** |
| stateless control (100 functions) | none | default | **0 of 200** |
| stateless, scaled ~8,600 mutants / 17 min | none | default | **0 of ~1,700** |

Each "canary" is a covered-but-unobservable mutation. In the CSLA case it is the removal of `context.AddSuccessResult(true)` from a single-rule success branch — equivalent because CSLA clears that rule's prior broken result on every re-run regardless. The README includes a one-command proof (delete every canary line, `dotnet test` → all 300 pass), so the canaries are demonstrably unkillable and any **Killed** count > 0 is a false positive.

Notes that rule out the obvious alternative explanations:
- Tests are stateless and run with `[assembly: CollectionBehavior(DisableTestParallelization = true)]`, so this is **not** intra-suite test parallelism.
- The repro sets `"additional-timeout": 30000`, and the misattributed mutants are reported `Killed`, not `Timeout` — so this is **not** the "timeouts are scored as killed" behavior.

**Expected behavior**

Each mutant is evaluated in isolation. A mutant that no test can detect (equivalent) must be reported **Survived**, deterministically, regardless of `--concurrency`.

**Logs**

The repro reproduces in ~30 s, so running it directly may be easiest — but I'm happy to attach `-L` logs from a run if useful; just let me know.

 - OS: Windows 11
 - Type of project: core (`net10.0`)
 - Framework Version: .NET SDK 10.0.301; xunit.v3 3.2.2; Microsoft.NET.Test.Sdk 18.5.1; runner `mtp`
 - Stryker Version: 4.16.0

**Additional context**

The CSLA half of the mechanism is documented, not guessed. CSLA registers each business-object type's rules **once per process** and caches them in **static** state — Rockford Lhotka: *"Validation rules are loaded once per-AppDomain per business object type. They are kept in a static cache … they are never unloaded — that is intentional"* (performance). In source, `BusinessRuleManager` holds a `static Lazy<ConcurrentDictionary<…>>` keyed by `Type`, and `BusinessBase.InitializeBusinessRules()` calls `AddBusinessRules()` exactly once per type under an `Initialized`-guarded double-checked lock. So CSLA's caching is intentional and thread-safe *for registration*; the isolation gap appears to be on the runner side — a shared test-server reused across concurrently-evaluated mutants. (Consistent with `mtp` not yet doing per-test coverage, i.e. the server being heavily shared.)

The precise leak across the shared server is the one inferred step; the CSLA static cache and the two levers (stateless → 0, `--concurrency 1` → 0) are directly demonstrated in the repro. Happy to run any diagnostic you suggest.

Possible mitigation: per-mutant process isolation (avoid reusing one test-server across concurrently-evaluated mutants). CSLA exposes no public API to reset its rule cache, so cache-clearing is not a clean option. `--concurrency 1` restores correct results today, at a large wall-clock cost.

Evangelink:
## Summary

- restart MTP test-host processes before and after each mutation when `disable-mix-mutants` is enabled
- clear the active mutant after each isolated run so later discovery or test sessions cannot inherit it
- add focused coverage for sequential reuse, opt-in isolation, and concurrent runner sessions
- document the MTP process-startup tradeoff

## Why

The MTP runner already schedules one mutant per test run, so `disable-mix-mutants` did not change its behavior. Test-host processes were still reused across mutations, allowing process-global state and MTP test-node state to affect later verdicts.

The stateless reproduction from #3742 reports 96.00% instead of the hand-verified 60.00% on current `master` and on #3837. With this change and `disable-mix-mutants`, it reports the correct 60.00%, all 18 equivalent mutants survive, and all 12 expected tests are attributed.

This is intentionally opt-in to avoid imposing test-host startup cost on every MTP mutation run. It complements #3695, which automatically isolates mutants classified as static or requiring isolation.

## Validation

- focused process-isolation unit tests: 3 passed
- MTP runner unit suite: 241 passed; one pre-existing 1-second server-start timeout test remains failing and also fails alone
- packaged `TUnit` MTP integration category and validation test passed
- external stateless #3742 regression harness: 60.00% reported, 18/18 expected survivors, 12/12 expected test attribution
- full `MTPSolution` integration was attempted but restore is blocked by the existing `NU1403` hash mismatch for `NUnit3TestAdapter` 6.3.0

Fixes #3742


richardwerkman:
I don't think we should use `disable-mix-mutants` for this behaviour. The setting is described as:

> Stryker combines multiple mutants in the same testrun when the mutants are not covered by the same unit tests. This reduces the total runtime. You can disable this behavior and run every mutation in an isolated testrun. This can be useful when mixed mutants have unintended side effects.

That is applicable to vstest but not to mtp, since we haven't implemented the mix mutants feature anyway. I don't think we need to implement it since the overhead is non-existent with mtp. 

I prefer #3695 which is a more sensible solution. @Evangelink thanks for your contribution, but I'm going to close this as unwanted

Evangelink:
Perfect! Glad to see MTP makes things easier. Please let me know if there are architectural missing points or extensibility that would make your life easier, faster or more reliable.

PR #3854, opened 2026-09-16
fix(mtp): run .NET Framework test apps directly

### LINKED ISSUE #3421: MTP test runner shows errors when running on .NET Framework test project

Evangelink:
## Summary

- execute MTP `.exe` test applications directly while continuing to use `dotnet` for `.dll` applications
- add a .NET Framework 4.8 MSTest 4.2.2 regression fixture and validation
- enable the Windows `MSTestMTP` + `netframework` integration-test matrix entry

## Validation

- MTP unit tests: 242 passed
- .NET Framework MTP integration: 2 tests discovered, 2 mutants killed, 0 runtime errors
- existing .NET MTP integration passed

Fixes #3421


Evangelink:
## Why `.github/workflows/integration-test.yaml` changed

The integration matrix explicitly excluded `MSTestMTP` with `netframework`. Removing that single exclusion makes the new regression fixture run on Windows CI. Without the YAML change, the fixture and validation would exist in the repository but would never be enforced by CI.

## Other implementation notes

- **Why `.exe` is launched directly:** MSTest MTP targeting .NET Framework emits an executable. Passing that executable to `dotnet` makes the host look for `hostpolicy.dll` and discovers zero tests. Modern `.dll` test applications keep the existing `dotnet` launch path.
- **Why the fixture has short names:** full-framework MSBuild and assembly loading still encounter legacy Windows path limits. Short project and folder names reduce failures in long checkout/worktree paths.
- **Why MSTest 4.2.2:** the fixture uses the same current MSTest SDK version as the repository's MTP unit-test project instead of copying the older integration fixture version.
- **Why process startup has explicit cleanup:** if starting the command fails synchronously, the pre-opened test-server log stream is disposed before the exception is rethrown.

.github/workflows/integration-test.yaml
richardwerkman: These can all be removed, except for TUnit

Issue #3855, opened 2026-09-16
Use IBlockingDataConsumer for MTP per-test coverage


Evangelink:
**Is your feature request related to a problem? Please describe.**

Stryker's Microsoft Testing Platform runner currently uses an epoch relay for per-test coverage when reusing a test-host process. After each test, the runner updates a memory-mapped file; a background thread in the injected `MutantControl` polls it every millisecond, flushes coverage to disk, resets the accumulated coverage, and acknowledges the epoch so the runner can safely read the result.

This works across the MTP versions Stryker currently supports, but it introduces a continuously polling thread, extra relay files, acknowledgement timeouts, and substantial synchronization code.

MTP 2.3 introduced the experimental `IBlockingDataConsumer` marker interface. A blocking consumer receives data inline, so the framework's `PublishAsync` call does not return until `ConsumeAsync` completes. This could give Stryker a reliable test-completion boundary at which to flush coverage without polling.

**Describe the solution you'd like**

Add a Stryker MTP test-host extension implementing `IBlockingDataConsumer`. It would consume terminal `TestNodeUpdateMessage` events and synchronously flush the coverage accumulated by the injected `MutantControl` before the test framework proceeds.

For zero-touch loading, MTP 2.4's dynamic extension manifests could load this extension into the test process when Stryker starts the test application with `--enable-dynamic-extensions`. The existing epoch relay would remain as a compatibility fallback for test applications using older MTP versions.

The implementation must account for frameworks that publish multiple terminal or parent/child updates for one requested test. Coverage should be accumulated for the complete request and must not be overwritten by an early terminal update.

**Decision requested from maintainers:** Is the reduced polling and synchronization complexity worth introducing a version-dependent path while Stryker supports both old and new MTP versions? Implementation and a linked PR should wait for approval of that direction.

**Describe alternatives you've considered**

It is entirely reasonable to avoid the conditional code. Stryker can keep the current epoch relay unchanged and adopt `IBlockingDataConsumer` only after support for MTP versions older than 2.4 is dropped. At that point, the blocking consumer could replace the relay rather than coexist with it.

Another alternative is to require users to reference or register a Stryker extension in their test projects, but that would add configuration and is less attractive than either dynamic loading or retaining the existing relay.

**Additional context**

- `IBlockingDataConsumer` is experimental and available from MTP 2.3.
- Zero-touch dynamic extension loading is available from MTP 2.4 and must be explicitly enabled for each run.
[...]

PR #3856, opened 2026-09-16
feat(mtp): use blocking consumer for per-test coverage

Evangelink:
Relates to #3855.

## Summary

- add a `netstandard2.0` MTP extension implementing the experimental `IBlockingDataConsumer`
- use inline terminal test updates to flush per-test coverage without the 1 ms epoch-polling thread on MTP 2.4+
- dynamically load the extension through an atomic, reference-counted manifest with stale-manifest cleanup
- retain the existing epoch relay for MTP 1.x through 2.3
- append and union coverage from multiple terminal or parent/child updates

## Compatibility

The blocking path is enabled only when the test application's dependency manifest reports Microsoft.Testing.Platform 2.4 or newer, because zero-touch dynamic extension loading starts in 2.4. Older applications continue using the existing relay unchanged.

The extension targets `netstandard2.0`, so it can load into supported .NET 8, .NET 9, and .NET 10 test hosts.

This PR is intentionally a draft pending maintainer approval of the version-dependent path. If maintainers prefer to avoid conditional code, it is entirely fine to close or defer this PR and adopt `IBlockingDataConsumer` only after Stryker stops supporting MTP versions older than 2.4.

## Validation

- MTP runner unit suite: 252 passed
- injected `MutantControl` coverage tests: 6 passed
- TUnit / MTP 2.4 mutation integration and report validation passed
- xUnit / MTP 1.9 fallback mutation integration and report validation passed
- solution restore succeeds in locked mode


dupdob:
Thanks for this PR. Quick question: have you tested this logic against xUnit (and MTP 2.x)? I am asking this because I am afraid that xUnit only performs asynchronous notifications due to its own design architecture. In that case, this approach would not work. 
This changed was introduced with xUnit V3, which forced us to (re) introduce a degraded mode with VsTest.

And if this change is not compatible with the dominant test framework, I am afraid it has limited value

Evangelink:
Good point, let me double check all of that

